PR #3, opened 2015-09-29
Resolve race condition on resizing bqplot Figures

SylvainCorlay:
@jasongrout This should solve the issue you were seeing earlier on the race condition with the svg transform.

### DIFF
--- a/bqplot/nbextension/Figure.js
+++ b/bqplot/nbextension/Figure.js
@@ -404,15 +404,9 @@ define(["nbextensions/widgets/widgets/js/widget", "./components/d3/d3", "base/js
             this.svg.attr("width", null);
             this.svg.attr("viewBox", "0 0 " + this.model.get("min_width") +
                                         " " + this.model.get("min_height"));
-            setTimeout(_.bind(this.update_layout2, this), 0);
-        },
-        update_layout2: function() {
             var rect = this.el.getBoundingClientRect();
             this.width = rect.width > 0 ? rect.width : this.model.get("min_width");
             this.height = rect.height > 0 ? rect.height : this.model.get("min_height");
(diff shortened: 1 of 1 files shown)


bqplot/nbextension/Figure.js
```diff
@@ -404,13 +405,20 @@ define(["nbextensions/widgets/widgets/js/widget", "./components/d3/d3", "base/js
             this.svg.attr("width", null);
             this.svg.attr("viewBox", "0 0 " + this.model.get("min_width") +
                                         " " + this.model.get("min_height"));
-            setTimeout(_.bind(this.update_layout2, this), 0);
+            var that = this;
+            return new Promise(function(resolve, reject) {
+                setTimeout(resolve(_.bind(that.update_layout2), that), 0);
```
jasongrout: I think this should be `setTimeout(function() {resolve(that.update_layout2())}, 0);`

bqplot/nbextension/Figure.js
```diff
@@ -404,13 +405,20 @@ define(["nbextensions/widgets/widgets/js/widget", "./components/d3/d3", "base/js
...
+
         },
         update_layout2: function() {
             var rect = this.el.getBoundingClientRect();
             this.width = rect.width > 0 ? rect.width : this.model.get("min_width");
             this.height = rect.height > 0 ? rect.height : this.model.get("min_height");
-            setTimeout(_.bind(this.update_layout3, this), 0);
+            var that = this;
+            return new Promise(function(resolve, reject) {
+                setTimeout(resolve(_.bind(that.update_layout3), that), 0);
```
jasongrout: Again, I think this should be `setTimeout(function() {resolve(that.update_layout3())}, 0);`

PR #5, opened 2015-09-29
A couple of Python 3 fixes

SylvainCorlay:
In the example notebooks and for the orientation of axes.


rgbkrk:
Hey thanks, I just bumped into this and was about to submit patches.

PR #16, opened 2015-10-01
Use a secure address for cloning

blink1073:
The current method gives me this:

```bash
$ git clone git@github.com:bloomberg/bqplot.git
Cloning into 'bqplot'...
Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```


SylvainCorlay:
Thanks!

PR #21, opened 2015-10-05
Get the bower js files before building the egg info

jasongrout:
Previously, bdist_egg would build the manifest before getting the js files, so it would leave the js files out. This affected the conda install, for example.  This commit ensures that the js files are always retrieved before building the manifest.


SylvainCorlay:
Thanks.

PR #25, opened 2015-10-07
Jshint fixes

ChakriCherukuri:
Fixed JSHint errors

### DIFF
--- a/bqplot/nbextension/Axis.js
+++ b/bqplot/nbextension/Axis.js
@@ -100,14 +100,15 @@ define(["nbextensions/widgets/widgets/js/widget", "./components/d3/d3", "./utils
                 if (this.axis_scale.model.type === "ordinal") {
                     this.axis.tickValues(this.axis_scale.scale.domain());
                 } else if (this.axis_scale.model.type === "log") {
+                    var i, r;
                     var allticks = this.axis_scale.scale.ticks();
                     var oom = Math.abs(Math.log10(this.axis_scale.scale.domain()[1] / this.axis_scale.scale.domain()[0]));
                     if (oom < 2) {
                         this.axis.tickValues(allticks);
                     } else if (oom < 7) {
                         useticks = [];
-                        for (var i = 0; i < allticks.length; i++) {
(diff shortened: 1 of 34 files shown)


SylvainCorlay:
As per in-person conversation with @ChakriCherukuri, the places where we used the square-brackets notation instead of the dot notation are the cases where we consider that the object has a "dictionary" semantic, i.e. corresponds to data, while the dot notation corresponds to an object semantic.

In order to avoid text editors with built-in linter integration to scream, we could add a `/*jshint -W069 */ directive` in the JavaScript file.

SylvainCorlay:
@sccolbert, what do you think of the above?

sccolbert:
I agree, when the keys are not known. Thinking about it from a typings perspective, you type a dict-like object as:

```typescript
interface MyDict {
  [key: string]: any;
}
```

But, if you expect certain keys to be present and finite you would type it like this (assuming all are optional):

```typescript
interface MyOffset {
  value?: number;
  scale?: number;
}
```

You can also have a combination:

```typescript
interface MyArgs {
  value?: number;
  scale?: number;
  [key: string]: any;
}
```

When framing things like that, I would say that if you know the expected structure of the object, treat it like struct with optional properties, rather than a dict. But it you don't know the structure, or need to dynamically query for keys, treat it as a dict. 

From what I see in the diff, I would treat the objects like a struct instead of a dict.

SylvainCorlay:
Ok, so the following mark attributes

 - `scales`
 - `preserve_domain`
 - `offset`
 - `interactions`
 - `[un]selected_styles`

reflect dictionaries on the Python side, and I think that we should use square bracket notation to access their attributes.

sccolbert:
IMHO it doesn't matter what they reflect on the Python side. They are separate languages. If the keys are more-or-less fixed, it's really more like a struct. If Python had JS-like object literals, you would probably be using those instead of a dict.

Again, this is just my opinion. This is your library :)

SylvainCorlay:
Thanks - merging.

PR #31, opened 2015-10-09
Cleaned up old commented-out code in OHLC.js

ClaytonPassmore:
I took out the old commented-out code that I wrote. I figured I would explain what it does here so that you do not feel guilty removing it.

A picture is worth 1,000 words, so take a look:
highlightbehaviour

Currently, the selection must be covering the centre of the mark in order for it to be highlighted. You can see that in this picture, the mark to the left of the cursor is not highlighted.

The old code would cause the mark to the left of the cursor to be highlighted because the selection is hitting its tail. Like this:
highlight

Side note: I moved some variables around so they were closer to the context they were being used in.


SylvainCorlay:
@ClaytonPassmore awesome! Getting back to you shortly.

SylvainCorlay:
Thanks it is all fine.

ClaytonPassmore:
@SylvainCorlay happy to help!

PR #35, opened 2015-10-11
Moved animate_dur attribute from marks to figure

ChakriCherukuri:
1. Moved animate_dur attribute from marks to figure. Since figure contains axis and multiple marks it makes sense for all of them to transition together.
2. Updated the example notebooks to reflect the above change
3. Fixed couple of bugs in Pie mark
4. Added a custom tween for Pie transitions

Note: Axis gridlines and ticks should also transition when individual marks transition. I am working on it and will create a separate PR for that.

### DIFF
--- a/bqplot/figure.py
+++ b/bqplot/figure.py
@@ -26,7 +26,8 @@ Figure
    Figure
 """
 
-from traitlets import Unicode, Instance, List, Dict, CFloat, Bool, Enum, Float
+from traitlets import (Unicode, Instance, List, Dict, CFloat, Bool, Enum,
+                       Float, Int)
 from ipywidgets import DOMWidget, register, Color, widget_serialization
 
 from .scales import Scale, LinearScale
--- a/bqplot/figure.py
+++ b/bqplot/figure.py
@@ -73,7 +73,7 @@ class Figure(DOMWidget):
         location of the legend relative to the center of the figure
     fig_color: Color (default: None)
         background color of the figure
-    animate_dur: nonnegative int (default: 0)
+    animation_duration: nonnegative int (default: 0)
         Duration of transition on change of data attributes, in milliseconds.
 
 
@@ -114,7 +114,7 @@ class Figure(DOMWidget):
                             'bottom-left', 'bottom', 'bottom-right', 'right'],
                            default_value='top-right', sync=True,
                            display_name='Legend position')
-    animate_dur = Int(0, sync=True, display_name='Animation duration')
+    animation_duration = Int(0, sync=True, display_name='Animation duration')
 
 
     def save(self):
--- a/bqplot/figure.py
+++ b/bqplot/figure.py
@@ -72,6 +72,9 @@ class Figure(DOMWidget):
         location of the legend relative to the center of the figure
     fig_color: Color (default: None)
         background color of the figure
+    animate_dur: nonnegative int (default: 0)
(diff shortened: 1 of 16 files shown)


SylvainCorlay:
:+1: it could be a good time to rename this to `animate_duration`. We should figure out how to deprecate a trait attribute name.

SylvainCorlay:
@ChakriCherukuri can you please also rebase?

ChakriCherukuri:
Added the following changes - 
1. renamed "animate_dur" to "animation_duration" as per request of @SylvainCorlay 
2. enabled axis, gridlines and tick animations
3. Now enabling animations only on data updates (not when rendering plots)
4. added an example notebook which showcases animations of various marks on data updates
5. enabled animations (only on data updates) of non-positional data attributes of Scatter

@SylvainCorlay and @ssunkara1 please review (esp axis related changes)

SylvainCorlay:
OK, I think this needs a simple rebase.

ChakriCherukuri:
rebase done

ssunkara1:
@ChakriCherukuri In the axis example, with the offset, the grid_lines do not go beyond the axis towards the negative numbers.

ChakriCherukuri:
@ssunkara1 fixed the gridlines to go beyond the axis when offset is specified. Please review

ssunkara1:
@ChakriCherukuri sorry for the delay. The change with the offset looks good. Thanks!

I just noticed a small bug. If I set the `grid_lines` to `solid` or `dashed` and then change them to `none`, the tick size still is the height or width of the figure. Hence there are solid lines in place of the grid lines.

Also, I think we should remove the string `'none'` and change it to a python `None`. Let me know what you think.

ssunkara1:
@ChakriCherukuri Also, I noticed that the logic for animation in Lines and Hist is slightly different from the others. In these two, there is no explicit check for animate, which is present in the other marks. Any reason why they are different?

SylvainCorlay:
Not sure about changing `'none'` to `None` since it would not be backward compatible.

ChakriCherukuri:
@ssunkara1 Hist animations are tricky and currently the way they are done (each bar expanding in horizontal direction) is distracting and don't serve the purpose. I will do them later. For now it's better not to do Hist animations. Let me know if you think otherwise.

I will fix the gird lines bug. Thanks for letting me know.

I will also do the line animations (only on data updates)

ChakriCherukuri:
@ssunkara1 Can you please check the axis issue we'd discussed today? Everything else looks good

SylvainCorlay:
This is looking good to me. It is a true improvement. Unless @ssunkara1 does not comment or merge, I will merge soon.

ChakriCherukuri:
@SylvainCorlay There's an issue with offsets which I and @ssunkara1 are looking into. Once we resolve it he will merge it.

SylvainCorlay:
ok, let me know. I would like to push out a release before tomorrow's presentation, preferably with that PR in.

ChakriCherukuri:
@SylvainCorlay since there are lot of changes it'd better you cut a release without this. I want to do more testing

ChakriCherukuri:
@SylvainCorlay fixed all the issues. Can you also test on your end to see if everything looks good? Then we can merge

SylvainCorlay:
The Animation notebook is great. Thanks.

Issue #36, opened 2015-10-12
Won't install on top of "docker-demo-image" (jupyter/demo on DockerHub) UnicodeDecodeError


mjbright:
I've tried installing this on the Jupyter/demo image, but I get the following error.

The following is the end of the output log of a build built on top of the jupyter/demo image:

Any idea how to fix this?

If you need to test, my Dockerfile is at
     https://github.com/mjbright/jupyter_trusted_builds/blob/master/demo/Dockerfile
```
Step 31 : RUN pip install bqplot
Collecting bqplot
Downloading bqplot-0.3.8.tar.gz (451kB)
Requirement already satisfied (use --upgrade to upgrade): ipywidgets in /opt/conda/lib/python3.4/site-packages (from bqplot)
Requirement already satisfied (use --upgrade to upgrade): numpy in /opt/conda/lib/python3.4/site-packages (from bqplot)
Requirement already satisfied (use --upgrade to upgrade): pandas in /opt/conda/lib/python3.4/site-packages (from bqplot)
Requirement already satisfied (use --upgrade to upgrade): python-dateutil>=2 in /opt/conda/lib/python3.4/site-packages (from pandas->bqplot)
Requirement already satisfied (use --upgrade to upgrade): pytz>=2011k in /opt/conda/lib/python3.4/site-packages (from pandas->bqplot)
Requirement already satisfied (use --upgrade to upgrade): six>=1.5 in /opt/conda/lib/python3.4/site-packages (from python-dateutil>=2->pandas->bqplot)
Building wheels for collected packages: bqplot
Running setup.py bdist_wheel for bqplot
Stored in directory: /home/jovyan/.cache/pip/wheels/e4/f9/62/512546616698b2ac0b48bf9042ce807cb914987993d592b453
Successfully built bqplot
Installing collected packages: bqplot
Successfully installed bqplot-0.3.8
---> 5c92ffed1fbe
```

```
Step 32 : RUN python -m bqplot.install --symlink --user --force
[91m/opt/conda/bin/python: Error while finding spec for 'bqplot.install' (<class 'UnicodeDecodeError'>: 'ascii' codec can't decode byte 0xc3 in position 98318: ordinal not in range(128))
```

SylvainCorlay:
@mjbright thanks for reporting. I am not experienced with docker, and I have not met this problem yet. Have you tried running the second step without the `--symlink` argument?

mjbright:
Thanks for the quick reply.

[...]

PR #37, opened 2015-10-13
explicitly say the install file is utf-8

jasongrout:
Attempts to address #36, though it's weird that it seems that we can happily encode the file to ascii.

### DIFF
--- a/bqplot/install.py
+++ b/bqplot/install.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+# -*- coding: utf-8 -*-
 
 import argparse
 from os.path import dirname, abspath, join


SylvainCorlay:
We might have to do this on all the file. In this PR, it could be interesting to do it in `__init__.py` as well since it is used to access `install.py`.

SylvainCorlay:
@mjbright does this fix your issue with docker?

mjbright:
Hi, sorry I've been travelling I did try without success but haven't been
able to spend more time investigating ... and probably won't for a month
now.

SylvainCorlay:
This needs a rebase.

SylvainCorlay:
I have searched for UTF chars and not found a single one so this should not be necessary.

jasongrout:
Let's close this for now, then.

jasongrout:
@mjbright - since this doesn't seem to fix the issue for you, I'm closing this.  Please let us know if you see this issue again and we'll try to help debug it better.

PR #41, opened 2015-10-14
Use country name as chart title in example.

SylvainCorlay:
- Use the country full name as title in the chart for the MarketMap example
- better default radius in Pie example.


SylvainCorlay:
@ssunkara1 will merge if you :+1:

ssunkara1:
@SylvainCorlay Looks good!

PR #42, opened 2015-10-14
Move pyplot to the top-level

SylvainCorlay:
@jasongrout this simply removes the intermediate pyplot directory as per discussion in person.


SylvainCorlay:
@ssunkara1 what do you think of this?

ssunkara1:
This looks good! Merging..

PR #51, opened 2015-10-25
Bug fixes

ChakriCherukuri:
1. Fix to make the diamond mark work in scatter
2. Fixed a bug in 'drag_ended' method where x_scale and y_scale are not defined
3. Using the correct event handler for change in 'default_rotation'

### DIFF
--- a/bqplot/nbextension/Scatter.js
+++ b/bqplot/nbextension/Scatter.js
@@ -827,12 +827,14 @@ define(["./components/d3/d3", "./Mark", "./utils", "./Markers"], function(d3, Ma
             if (!this.model.get("enable_move")) {
                 return;
             }
-            if(!this.drag_started) {
+            if (!this.drag_started) {
                 return;
             }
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
:+1:

bqplot/nbextension/Scatter.js
```diff
@@ -827,12 +827,14 @@ define(["./components/d3/d3", "./Mark", "./utils", "./Markers"], function(d3, Ma
...
-            if(!this.drag_started) {
+            if (!this.drag_started) {
                 return;
             }
             var dot = this.dot,
                 default_colors = this.model.get("default_colors"),
-                len = default_colors.length;
+                len = default_colors.length,
+                x_scale = this.scales.x,
+                y_scale = this.scales.y;
```
SylvainCorlay: I wonder how this worked before. Was this bug introduced recently?

PR #58, opened 2015-10-30
Enabling marker strings in pyplot (a la matplotlib)

ChakriCherukuri:
1. Added a new mark called MarkerLines (which extends Lines and adds markers)
2. Enabled 'marker strings' in pyplot as 3rd argument (e.g. 'g-o', green dots with solid line etc)
3. Updated Pyplot.ipynb to show examples of marker strings


ChakriCherukuri:
@ssunkara1 @rmenegaux 
I agree with your suggestions to add markers to Lines as opposed to created a separate mark. The only issues are handraw and basis interpolation will not work properly (but this is expected).

The last 7 commits contain changes where I enhanced Lines to have markers and removed MarkerLines mark. Please review

ssunkara1:
@ChakriCherukuri Sorry for the delay. Everything looks great! just one small thing. When I pass `color` attribute to a Lines object, the dots are not of the correct color. The legend is of the correct color though.

Apart from that I am good to merge.

PR #60, opened 2015-10-30
Area Chart

rmenegaux:
Added area chart functionalities to the `Lines` mark

* `fill` controls what area to fill
   * `none` is the default no-fill
   * `top`, `bottom` fills to the plot area upper and lower limits
   * `inside` is the previous behavior for filling, it closes the path and fills it in

* `fill_colors` and `fill_opacities` control the styling of the filling, they work as `opacities` and `colors` currently do
* If no `fill_colors` is provided, the fill color defaults to the corresponding line's color.

### DIFF
--- a/bqplot/marks.py
+++ b/bqplot/marks.py
@@ -249,12 +249,12 @@ class Lines(Mark):
     fill: {'none', 'bottom', 'top', 'inside'}
         Fill in the area defined by the curves
     fill_colors: list of colors (default: [])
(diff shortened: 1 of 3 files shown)


rmenegaux:
Ready for review

rmenegaux:
@SylvainCorlay perfected by your comments

SylvainCorlay:
As per discussion with @rmenegaux we will probably need to handle the `curve_subset` differently for this case.

PR #67, opened 2015-12-03
First pass at Selection on GridHeatMap

dmadeka:
This PR enables a click interaction for selection on the GridHeatMap. The selection works as in other marks but mimics Excel in that, the `ctrl` key allows additional selection, while the `Shift` key selects the entire range from the last selected cell to the current clicked cell.

image

The range selection allows arbitrary ranges to be selected with the `Shift` key. 

image

The `grid_map.selected` trait returns a list of list containing the row index and the column index of each selected cell.

### DIFF
--- a/bqplot/nbextension/GridHeatMap.js
+++ b/bqplot/nbextension/GridHeatMap.js
@@ -31,8 +31,10 @@ define(["./components/d3/d3", "./Mark", "./utils", "underscore"],
                 that.create_tooltip();
             });
 
+            this.selected_indices = this.model.get("selected");
             this.selected_style = this.model.get("selected_style");
             this.unselected_style = this.model.get("unselected_style");
+            this.anchor_style = this.model.get("anchor_style");
             this.display_el_classes = ["heatmapcell"];
             return base_render_promise.then(function() {
                 that.event_listeners = {};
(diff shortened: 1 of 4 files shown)


dmadeka:
@ssunkara1 @SylvainCorlay Any thoughts?

@rmenegaux Thanks for the tips.

bqplot/nbextension/GridHeatMap.js
```diff
@@ -121,6 +122,73 @@ define(["./components/d3/d3", "./Mark", "./utils"], function(d3, MarkViewModule,
...
+                    _.range(Math.min(row, row_index), Math.max(row, row_index)+1).forEach(function(i) {
+                        _.range(Math.min(column, col_index), Math.max(column, col_index)+1).forEach(function(j) {
+                            var cell_num = that._cell_nums_from_indices([[i, j]])[0];
+                            if (selected.indexOf(cell_num) === -1) {
+                                selected.push(cell_num);
+                                idx.push([i, j]);
+                            }
+                        })
+                    });
+                } else if(!(d3.event.ctrlKey)) {
```
dmadeka: @rmenegaux if we move this to the top and make the next if statement an else. This will replicate Excels behaviour exactly, in that the `Shift` key will create a new block and the double `Shift` will have no effect. 

What does everyone else think?

SylvainCorlay:
I love it!
As per conversation in person, it would be interesting if the anchor of the selection was styled differently.

SylvainCorlay:
@dmadeka I am looking at it at the moment. It seems to work well except that
 - the Crtl key on OS X should not be the one used, but rather the Command key. For this to work, you would need to use the `Accel` modifier key, which is Control on windows and command on OS X.
 - can you clear the output of the cells before committing notebooks?

SylvainCorlay:
@dmadeka ok, let's do the following: check for *both* `crtlKey` and `metaKey` and add a comment specifying that what we want to do is:

```JavaScript
d3.event.getModifierState("Accel")
```

to use the virtual "Accel" modifier, which is not yet supported on chrome.

dmadeka:
@SylvainCorlay this is ready for review.

The metaKey event handler from d3 has been added and selection has been made like Excel. An anchor cell has also been added.

Thanks,

Dhruv

SylvainCorlay:
:tada:

PR #78, opened 2016-01-06
setup.py: fail with exception when bower cannot be found.

hohn:
In our installation, bower was not on the path and the installation quietly failed.  With the exception, it should stop.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -141,6 +141,8 @@ class Bower(Command):
                     cwd=here,
                     env=os.environ.copy(),
                 )
+            else:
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
:+1: can you please make it a `RuntimeError` rather than a base `Exception` and specify `developer installation` in the error message?

hohn:
Done.

SylvainCorlay:
Thanks!

hohn:
You're welcome!

PR #94, opened 2016-02-10
scales domain setting bug fix

ssunkara1:
There was an issue with scales in the same mark being used for different attributes. Suppose the same scale is used for `x` and `y`, then the last attribute to be set gets control of the domain as the key in the dictionary was the mark model id. 

Adding the attribute as a suffix to the key so that it does not over write values set by other attributes.


SylvainCorlay:
I am :+1:  on this change. Although, when you do `this.id + "_x"`, you could also do

`this.id + '_' + name` since we are looping over scale names.

ssunkara1:
@SylvainCorlay removed the code for scatter. For others, all scales are not equivalent.

PR #95, opened 2016-02-17
IE11 support

SylvainCorlay:
IE11 does not support classList for svg elements.

### DIFF
--- a/bqplot/nbextension/Figure.js
+++ b/bqplot/nbextension/Figure.js
@@ -22,7 +22,8 @@ define(["nbextensions/widgets/widgets/js/widget", "base/js/utils",
 
         initialize : function() {
             this.setElement(document.createElementNS(d3.ns.prefix.svg, "svg"));
-            this.el.classList.add("bqplot", "figure");
+            // Internet Explorer does not support classList for svg elements
+            this.el.setAttribute("class", "bqplot figure");
             Figure.__super__.initialize.apply(this, arguments);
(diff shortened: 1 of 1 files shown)


bqplot/nbextension/Figure.js
```diff
@@ -22,7 +22,8 @@ define(["nbextensions/widgets/widgets/js/widget", "base/js/utils",
 
         initialize : function() {
             this.setElement(document.createElementNS(d3.ns.prefix.svg, "svg"));
-            this.el.classList.add("bqplot", "figure");
+            // Internet Explorer does not support classList for svg elements
+            this.el.setAttribute("class", "bqplot figure");
```
sccolbert: `this.el.className = "bqplot figure"` is another option

sccolbert:
IE 11 also doesnt support the multi-argument form of `classList.add` in general

bqplot/nbextension/Figure.js
SylvainCorlay: Thanks. This does not seem to work on chrome, which complains that the `className` property only has a getter and no setter.

sccolbert: weird. that must also be specific to svg nodes, as we use `className` all over phosphor.

SylvainCorlay: The `className` property is a read-only `SVGAnimatedString` object...

PR #96, opened 2016-02-18
Various bug fixes

ssunkara1:
Fixed the following bugs:

Positioning of tooltip when scrolling horizontally.
`MarketMap` colors are not applied on initial draw.
`LinearColorScale` `mid` attribute not updating.
`ColorAxis` auto tick format not working.
Tooltip is blocking mouse events after it is invisible.
`FlexLine` update of  the domain of the scales for `x`, `y` and `color` were failing.

Added the ability to output normalized data in a `Histogram` so that it can be used to compare different Histograms.


SylvainCorlay:
@ssunkara1 is this ready?

SylvainCorlay:
Note for self: check for default value in the webpack branch for the `normalized` attribute.

ssunkara1:
@SylvainCorlay Yeah this is ready

PR #109, opened 2016-03-13
Create single requirejs entry point for bqplot

SylvainCorlay:
@jdfreder, one step in the direction of having the web2 example work without overloading loadClass for bqplot. The target of the PR is the "webpack" branch, where I have the ipywidgets 5 compatibility changes.

I noticed an issue here: I had to create a map for `jupyter-js-widgets` because the bqplot nbextension happens to be loaded before `jupyter-js-widgets`, and the configuration of the requirejs for jupyter-js-config is not there yet.

```JavaScript
requirejs.config({
    map: {
        '*' : {
            "bqplot": "nbextensions/bqplot/index",
            "jupyter-js-widgets": "nbextensions/widgets/extension",
        }
    }
});
```

I don't see much we could do about it.


jdfreder:
I think you could call notebook's base/js/utils.trailets load_extension method

SylvainCorlay:
ok, thanks, that is a good idea.

SylvainCorlay:
@jdfreder I updated the pr to call load_ipython_extension and remove some requirejs shims.

PR #113, opened 2016-04-08
Move bqplot JavaScript into a separate node module

SylvainCorlay:
This also completely remove bower.

- [ ] generate embedder that does not webpack jupyter-js-widgets.


SylvainCorlay:
@jdfreder what I should probably do is have the webpack'-ed extension not actually webpack jupyter-js-widgets.

PR #119, opened 2016-04-13
Multi selector fix

### LINKED ISSUE #118: Multi Selector
I have a question regarding the Multi Selector elaborated [here](https://github.com/bloomberg/bqplot/blob/master/examples/Interaction%20Layer.ipynb).

What do you do to multi-select? I use mouse to select an interval but when I select again, only the first the interval is updated. How can I select a new interval?

Thanks

SylvainCorlay:
Fixes #118 . Thanks to @ssunkara1 .


ssunkara1:
Thanks @daviddaileo

PR #125, opened 2016-04-20
WIP Save as svg

SylvainCorlay:
This makes a change to the toolbar to save an svg file instead of a png.

We should probably allow both.


dmadeka:
@SylvainCorlay Any hope for this?

ecx2093:
Any progress? This is a must-have for exporting publication quality vector graphics.

PR #147, opened 2016-05-18
Add callbacks for all drag behavior in Scatter

dmadeka:
This pull request adds two more callbacks to Scatter, namely the `on_drag_start` which is called when a point triggers a d3 `dragstart` event and `on_drag` which is called while a point is being dragged.

Unfortunately, it seems, for `dragstart` and `dragend` the current position of the `Scatter` point cannot be known. If they can, a similar template for `drag` can be followed.

@ssunkara1 @SylvainCorlay


dmadeka:
@ssunkara1

PR #148, opened 2016-05-18
[WIP] Image mark

izaid:
This is a massive work in progress, but I've got some basic things working so far. More soon.

### DIFF
--- a/bqplot/marks.py
+++ b/bqplot/marks.py
@@ -1134,3 +1134,42 @@ class GridHeatMap(Mark):
 
     _view_name = Unicode('GridHeatMap').tag(sync=True)
     _model_name = Unicode('GridHeatMapModel').tag(sync=True)
+
+
+@register_mark('bqplot.Image')
+class Image(Mark):
+
+    # Mark decoration
+    icon = 'fa-globe'
+    name = 'Image'
+
+    # Scaled attributes
+    color = Dict(allow_none=True).tag(sync=True, scaled=True, rtype='Color',
+                                      atype='bqplot.ColorAxis')
+
+    # Other attributes
+    scales_metadata = Dict({'color': {'dimension': 'color'}}).tag(sync=True)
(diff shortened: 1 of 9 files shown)


izaid:
The NdArray trait is too slow for images, by the way. We need to update the _to_json method to pass the data as bytes in a field.

For now, I can pass the bytes directly as a unicode trait, just like Jupyter does.

SylvainCorlay:
Woot! Thanks for starting this!

Regarding the NdArray, there is a buffer attribute in the comm protocol that we can use for binary data. We should move to that eventually.

izaid:
Okay, so this is a very basic Image mark that now works. I've even added an example notebook at "Image.pynb". Obviously, there is a lot more to do here. The limitation really is the JSON serialization, as I mentioned above. I think we should merge this if @SylvainCorlay is happy, then I'd like fix the JSON serialization, come back, and make this much better.

@SylvainCorlay Are you happy with what I've done so far? If not, can you give me comments on what you'd like changed?

Right now, the image has to be set with a binary data source, which is weird but at least it works. We'll change it to an NdArray once the JSON problem is resolved.

dmadeka:
@izaid I noticed nothing was changed in any of notebooks apart from `Image.ipynb`. Can you remove them from the PR?

izaid:
@dmadeka Yeah, sure, I'll make sure they are all removed before we merge. What I'd really like to know is what more you would like before you consider this "mergable". Again, my goal here is to get a basic Image mark in, fix up other stuff, then come back.

SylvainCorlay:
Hi all I can look into this and specifically the binary serialization
tomorrow morning.
Sylvain

dmadeka:
Why don't we discuss on the [gitter](https://gitter.im/bloomberg/bqplot)?

SylvainCorlay:
I am not available at the moment.

izaid:
Closing this, will reopen shortly.

PR #150, opened 2016-05-20
[WIP] Sending NdArray with base64 instead of JSON

izaid:
This PR changes the NdArray trait to efficiently encode / decode a NumPy array's data, instead of turning it all into JSON.

### DIFF
--- a/bqplot/traits.py
+++ b/bqplot/traits.py
@@ -37,6 +37,8 @@ import pandas as pd
 import warnings
 import datetime as dt
 
+from base64 import b64encode
+
 
 # Numpy and Pandas Traitlets
 class CInstance(Instance):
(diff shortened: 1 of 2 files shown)


bqplot/traits.py
```diff
@@ -112,18 +114,20 @@ def _to_json(a, obj=None):
...
                 a = a.astype(np.float64)
             elif np.issubdtype(a.dtype, np.datetime64):
                 dtype = 'date'
                 a = a.astype(np.str)
+            elif a.size == 0:
+                a = np.empty(0)
             else:
                 dtype = a.dtype
-            return {'values': a.tolist(), 'type': str(dtype)}
+            return {'values': a.tolist(), 'data': b64encode(a), 'type': str(a.dtype), 'shape': a.shape, 'size': a.size}
```
sccolbert: Is `values` supposed to remain included here? Otherwise this seems to be *more* work, rather than less.

izaid:
No, it's not. It'll go when this is not WIP, I just wanted to be clear what I was doing.

jasongrout:
We could save the 33% base64 overhead if we instead used the numpy binary format: http://docs.scipy.org/doc/numpy/neps/npy-format.html. We'd need to implement a simple parser on the javascript side for the javascript bytes object.

SylvainCorlay:
I am closing this since we are going to move to binary serialization for bqplot attributes.

Issue #154, opened 2016-06-03
Scatter does not have select in interactions for click and hover


dmadeka:
Scatter doesn't have select as an interaction in `process_interactions` for `select`

PR #160, opened 2016-06-14
Add selection for click in Scatter Chart interactions

dmadeka:
@ssunkara1 @ChakriCherukuri @rmenegaux This is regarding #154.

Should we have the `shift` key functionality for `Scatter` as well? It doesnt seem to make sense to me, since inherently there's no connection between the points index and its x-coordinate, should the shift move in x or y coordinates then?

### DIFF
--- a/examples/Mark Interactions.ipynb
+++ b/examples/Mark Interactions.ipynb
@@ -23,6 +23,77 @@
    ]
   },
   {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "#### Scatter Chart Selections\n",
+    "\n",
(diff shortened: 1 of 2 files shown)


dmadeka:
@ssunkara1 @rmenegaux I added a `hover` select in case you want to trigger a specific type of behavior on `hover`. I have `click` overriding it, but we can create an attribute called `hovered_point`, the good thing is here we dont need another style like `hovered`, but then we can't `hover` and then we can't `hover`, trigger a behavior and then `select` that point. Is that worth adding another style and having to override apply_styles etc?

PR #163, opened 2016-06-16
Renamed ordinal scales to categorical scales

rmenegaux:
Since the ordinal scales do not take into account the ordering of the data (even if there is one), it makes more sense to call them categorical scales.
A categorical scale is just a mapping from a finite set of values to another finite set of values.


dmadeka:
@rmenegaux Not sure about this one. The `OrdinalScale` exposes `d3.scale.ordinal`, which is probably its namesake. 

It's a major API change that will break any code using `OrdinalScale` for a minor naming inconvenience (alternately why not `NominalScale`). Anyway, we should first deprecate it with warnings before we switch it, if at all.

rmenegaux:
@dmadeka Reinstated the ordinal scales, which will trigger deprecation warnings if used.

dmadeka:
As discussed in the dev meeting, the point of `bqplot` is to expose d3, and despite a misnaming, we might want to stick to the d3 naming convention

Issue #155, opened 2016-06-03
Scatter `hovered_style` and `unhovered_style`


dmadeka:
Scatter should have the ability to style a point on hover, as it does on select

PR #165, opened 2016-06-16
Hover focus

dmadeka:
@rmenegaux Added the hover focus ability.

Might want to make a `Super` call in the `apply_styles`, and the style changed code is very similar except for what it looks for. Addresses #155

### DIFF
--- a/js/src/Scatter.js
+++ b/js/src/Scatter.js
@@ -595,6 +595,60 @@ define(["d3", "./Mark", "./utils", "./Markers", "underscore"],
             this.touch();
         },
 
+	scatter_click_handler: function(args) {
+            var data = args.data;
+            var index = args.index;
+            var that = this;
+            var idx = this.model.get("selected");
+            var selected = idx ? utils.deepCopy(idx) : [];
+            // index of bar i. Checking if it is already present in the list.
+            var elem_index = selected.indexOf(index);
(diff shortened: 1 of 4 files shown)


js/src/Scatter.js
```diff
@@ -729,6 +791,31 @@ define(["d3", "./Mark", "./utils", "./Markers", "underscore"],
...
+            this.clear_style(this.unselected_style);
+            this.clear_style(this.hovered_style);
+	    this.clear_style(this.unhovered_style);
+
+            this.set_default_style(all_indices);
+
+            this.set_style_on_elements(this.selected_style, this.selected_indices);
+            var unselected_indices = (!this.selected_indices) ?
+                [] : _.difference(all_indices, this.selected_indices);
+            this.set_style_on_elements(this.unselected_style, unselected_indices);
```
dmadeka: Probably everything before this can go into a super call. What do you think @ssunkara1 ?

dmadeka:
@ssunkara1 Whats the protocol here for a super call?

dmadeka:
@SylvainCorlay The `super` call seems to be tough here. For some reason, `Scatter` (or other closure variables but not `d3`) are defined in some functions but not in others. Is there a reason for this?

dmadeka:
@rmenegaux Made the requested changes. Feel free to comment/merge.

PR #166, opened 2016-06-17
Reinstated save_png method of figure

dmadeka:
@SylvainCorlay Brought back the save_png function for the `Figure`, @stonebig asked for a stable beta, I think this should probably be included

### DIFF
--- a/js/src/Figure.js
+++ b/js/src/Figure.js
@@ -173,9 +173,17 @@ define(["jupyter-js-widgets", "d3", "underscore"],
                     that.el.parentNode.appendChild(that.tooltip_div.node());
                     that.create_listeners();
                     that.update_layout();
+                    that.model.on("msg:custom", that.handle_custom_messages,
+				  that);
                 });
             });
         },
+
+	handle_custom_messages: function(msg) {
+            if (msg.type === 'save_png') {
+		    this.save_png();
+	    }
+	},
 
         replace_dummy_nodes: function(views) {
             _.each(views, function(view) {
--- a/bqplot/figure.py
+++ b/bqplot/figure.py
@@ -91,6 +91,11 @@ class Figure(DOMWidget):
         is responsible for making sure that the width and height are greater
         than the sum of the margins.
 
+    Methods
+    -------
+
+    save_png:
+       Saves the figure as a png file
     """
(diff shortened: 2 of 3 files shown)


SylvainCorlay:
The issue is that the save_png makes each view save the png.

dmadeka:
As in all views or the current view? Because the latter seems to be what we want right?

----- Original Message -----
From: Sylvain Corlay <reply@reply.github.com>
To: bqplot@noreply.github.com
CC: DHRUV MADEKA, author@noreply.github.com
At: 17-Jun-2016 15:42:08


The issue is that the save_png makes each view save the png.

SylvainCorlay:
All the views... It is a kernel function that sends a message to all the
views...

That is why we need to make the toolbar a child widget and have the save
button only save the one of the view of the parent figure...

dmadeka:
Alternately, we could have a little button on the `Figure` itself right? A controllable mini-toolbar thats displayed or not and clicking that could save the `Figure`. Because right now to save, you have to create a `Toolbar` instance, plug it to the `Figure` and then display it and click save. There's no Pythonic way or simple way to do it

SylvainCorlay:
Actually, the direction I wanted to take was to move up the toolbar widget
as a child widget and have the figure instantiate a default toolbar. This
would make it more easily extensible ...

Sylvain

dmadeka:
Well. Ideally, the `Toolbar` should allow you to configure any properties of the `Figure` right? So that's the right way to go I guess

dmadeka:
btw @SylvainCorlay will the `Toolbar` as a child solve the issue that we aren't able to programatically save a `png`?

SylvainCorlay:
No, I am still ok with your PR!

dmadeka:
okay @SylvainCorlay yours to merge then

js/src/Figure.js
```diff
@@ -173,10 +173,18 @@ define(["jupyter-js-widgets", "d3", "underscore"],
...
+				  that);
                 });
             });
         },
 
+	handle_custom_messages: function(msg) {
+            if (msg.type === 'save_png') {
+		    this.save_png();
+	    }
+	},
```
SylvainCorlay: ah, formatting.

SylvainCorlay:
Thanks

PR #168, opened 2016-06-20
Fix Marker Color Bug

### LINKED ISSUE #156: `Lines` `marker` attributes color bug
If you add a `marker` to a `Line` after a view has been instantiated, then the markers are colored Black and don't have the same color as the `Line`

dmadeka:
@rmenegaux Fixes #156


dmadeka:
@rmenegaux should be good to go

PR #171, opened 2016-06-21
Euro map

dmadeka:
Added European countries `Map`


dmadeka:
@rmenegaux @SylvainCorlay Done! We should play with projections and see which the best one is

SylvainCorlay:
@dmadeka @rmenegaux I reverted the PR as per discussion. Can you reopen?

PR #176, opened 2016-06-27
Figure SVG size based on min width and height

bfelds:
In update_layout this change will explicitly use min_height
and min_width for the viewport.


bfelds:
assign: @SylvainCorlay

PR #177, opened 2016-06-29
Update Hover Styles

dmadeka:
@rmenegaux


dmadeka:
@rmenegaux good to go

Issue #181, opened 2016-06-30
Pyplot notebook doesnt work with Python 3


dmadeka:
As pointed out, we should fix this quickly

PR #183, opened 2016-07-01
pyplot adding order for compatible scales

ssunkara1:
Added a class level attribute called `preference_order` for Scales. If two scales have the same range and domain type, then the scale automatically selected by pyplot is based on the value of the new attribute.


dmadeka:
fixes #181

PR #184, opened 2016-07-01
Update Light Theme

dmadeka:
@bloomberg/bqplot-developers I updated the `theme-light` `Figure` background  because the grey is a bit too contrasting to something closer to white.

What other changes can we think of? Maybe crisper axes?

### DIFF
--- a/js/src/bqplot.less
+++ b/js/src/bqplot.less
@@ -125,8 +125,21 @@
     .common();
     svg.bqplot {
         .axis {
-            rect, path, line {
+            path, line {
                 stroke: #EEEEEE;
+            }
+            rect {
+                stroke: white;
+                opacity: 1.0;
+            }
+            .tick line {
+            	stroke: white;
+            	stroke-width: 1.4;
+                opacity: 1.0;
+            }
+            .tick text {
+                fill: #7f7f7f;
+                font: 12px sans-serif;
             }
         }
         .stick, .zeroLine {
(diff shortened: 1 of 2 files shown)


kpfleming:
There's no need to @ mention the developers team when you open an issue or PR; they'll all get notified that the issue was opened anyway.

dmadeka:
Kevin, that's correct. I do it more to signal to everyone that this is an issue they should all look at and vote on. Perhaps we should have a more efficient way of doing this

----- Original Message -----
From: Kevin P. Fleming <reply@reply.github.com>
To: bqplot@noreply.github.com
CC: DHRUV MADEKA, author@noreply.github.com
At: 01-Jul-2016 19:35:59


There's no need to @ mention the developers team when you open an issue or PR; they'll all get notified that the issue was opened anyway.

SylvainCorlay:
@dmadeka what do you think of a ggplot style for the light theme?
 - gray background
 - white gridlines
 - no axis line but keeping the tick marks and tick values in gray.

meat_smooth

dmadeka:
Okay @SylvainCorlay @ssunkara1 Ive started this, but it looks like this is going to have to be a bike shed. Here is how it looks now:

bqplot

SylvainCorlay:
This is cool! 

FYI. If you update ipywidgets to upstream master, the toolbar will look nicer. It is almost like ggplot now. The tickmarks are a bit darker in ggplot though.

dmadeka:
@SylvainCorlay The `ticks` seem to be connected to the gridlines. And the axes labels differ from matplotlib's `ggplot` theme and the actual `ggplot2` which would you prefer, I kept the former

js/src/bqplot.less
```diff
@@ -125,8 +125,21 @@
...
+                stroke: white;
+                opacity: 1.0;
+            }
+            .tick line {
+            	stroke: white;
+            	stroke-width: 1.4;
+                opacity: 1.0;
+            }
+            .tick text {
+                fill: #7f7f7f;
```
SylvainCorlay: If you remove that `fill` attribute, you retrieve the dark tick marks like in ggplot.

SylvainCorlay:
This is looking great

PR #187, opened 2016-07-05
Pyplot adding order for compatible scales

ssunkara1:
This is an equivalent to PR #183 . This is for the master branch rather than for 0.5.x

### DIFF
--- a/bqplot/pyplot.py
+++ b/bqplot/pyplot.py
@@ -390,7 +390,7 @@ def _draw_mark(mark_type, options={}, axes_options={}, **kwargs):
                                   for key in Scale.scale_types
                                   if Scale.scale_types[key].rtype == rtype and
                                   issubdtype(dtype, Scale.scale_types[key].dtype)]
-            sorted_scales = sorted(compat_scale_types, key=lambda x: x.preference_order)
+            sorted_scales = sorted(compat_scale_types, key=lambda x: x.preference)
             scales[name] = sorted_scales[0](**options.get(name, {}))
             # Adding the scale to the conext scales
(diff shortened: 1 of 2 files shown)


ssunkara1:
WIP is because I haven't tested this yet with Python 3.x. Will remove the WIP once I am done with that.

ssunkara1:
@SylvainCorlay Right now, I can't get the jupyter notebook with python 3.x to display widgets. 

Can you check this when you have time? With this change, `pyplot.plot` should plot the graph with negative values. Previously, this was returning a `LogScale` which results in no graph being displayed.

dmadeka:
@ssunkara1 Do you want me to take it from here?

ssunkara1:
@SylvainCorlay can you confirm that there is no issue with the tabs in this pr? Thank you!!

SylvainCorlay:
It seems to work. My 2c about the variable name:

what about `precedence`?

ssunkara1:
@SylvainCorlay seems good. will change.

dmadeka:
@SylvainCorlay Why not `preference`? So we can maintain the 1, 2, 3 order?

ssunkara1:
@SylvainCorlay @dmadeka made the change to `preference`.

ssunkara1:
@SylvainCorlay I am not convinced about `precedence`. I think, going forward, we will be adding scales which are of lower precedence than the ones we already have. So to avoid having to go to negative values, I think it is better to have a higher number for things of lower precedence.

Anyway I am ok with it either way since it is just a convention. I have a commit ready. You take the final call on this one.

PR #188, opened 2016-07-06
Change README.md to remove bq. imports

dmadeka:
@SylvainCorlay @rmenegaux This change is because none of the examples or talks ever use bq. This is pretty much the only place. I guess it needs to be changed.

I also added a title.


dmadeka:
@SylvainCorlay What do you think?

SylvainCorlay:
:+1:

PR #192, opened 2016-07-08
fixing date format for numpy change

ssunkara1:


ssunkara1:
Please refer #194. Both fix the same issue with slightly different solutions.

PR #194, opened 2016-07-08
fixing date format for numpy change version 2

ssunkara1:


ssunkara1:
This is for the same problem as PR #192 .
The difference is that, in this version, we are not relying on the string representation of a datetime64. The issue came up because of a change in the string representation. This gives us exact control of the format of the string which is passed across. But I do not really like the code there. Up to debate if this one is better or #192 is better.

aburgm:
I like this one more because it seems to be more robust. What if they change the string representation again? Also, would #192 break compatibility with numpy versions before 1.11?

ssunkara1:
@aburgm good point. Yes #192 will break compatibility with previous versions of numpy.

dmadeka:
+1 is there no better way to convert without relying on the string representation to convert to string?

This is going to `pandas datetime` to `datetime.datetime` to string

Issue #129, opened 2016-04-27
Switching order of data series does not update plot colors, but will update legend colors


ssunkara1:
When the labels text is same but the order is switched, the line chart labels do not match with the respective values.

jasongrout:
Here is a small example illustrating this:

```python
import bqplot.pyplot as plt
import pandas as pd
import numpy as np

df = pd.DataFrame(np.random.randn(100,4), columns=list('ABCD'))
df['A'] = np.arange(100)

plt.figure()
plt.plot(x=[df.A]*2, y=[df.A, df.B], labels=['A','B'], axes_options={'x': {'grid_lines': 'dashed'}, 'y': {'grid_lines': 'dashed'}})
plot_fig0 = plt.current_figure()
plot_fig0.marks[0].display_legend = True
plt.show()
```
```python
# Switches colors appropriately
with plot_fig0.hold_sync():
    plot_fig0.marks[0].y = [df.B, df.A]
    plot_fig0.marks[0].labels = ['B', 'A']
```
```python
# Switches labels and colors in the legend, but does not change colors in the plot
with plot_fig0.hold_sync():
    plot_fig0.marks[0].y = [df.A, df.B]
    plot_fig0.marks[0].labels = ['A', 'B']
```

PR #210, opened 2016-07-14
remove indexing by label in Lines.js

rmenegaux:
The curves should not be indexed by their labels since we do not impose a one-to-one mapping between them.
This fixes issue #129
It will also make the legend display properly in case there are identical labels.


SylvainCorlay:
LGTM

PR #214, opened 2016-07-14
line labels move with the line

rmenegaux:
This does two things:
- Separates the creation of the labels and their positioning. The label positions update correctly now, including with panning and zooming.
- Adds a listener to a change in the `labels` trait, that correctly updates the inline labels.
Also updates the d3 bound data of the elements, to keep the names consistent.


dmadeka:
@SylvainCorlay good to merge

PR #215, opened 2016-07-16
Added 'solid' grid lines by default

dmadeka:
@ssunkara1 @SylvainCorlay Now that the grid lines are more discreet, we should probably make them the default


SylvainCorlay:
@dmadeka can you make the change of the default value on the js side?

PR #216, opened 2016-07-16
Modify README for new design

dmadeka:
@SylvainCorlay If you want to push better screenshots, can you do that?


dmadeka:
@SylvainCorlay Any hope for doing the screenshots/casts? Which is the software? @rmenegaux has a Mac as well

PR #217, opened 2016-07-17
Allow Label to take multiple values

dmadeka:
@SylvainCorlay @ssunkara1 This is per our discussion that `Label` should take multiple values. I think globally, this PR will be redundant soon, but is necessary.

I think `Scatter`, `Label` (or even `Image`) is basically placing an element at a certain point on the scale. It should probably be moved to a global `Object` (or global `Scatter` mark like Altair) and have some custom behaviours, but the drag, placement, coloring etc. should all be an `Objects` property. 

PS `Objects` should allow us to place any DOM element at the point right, so why not put an entire `Figure` so as to have a `Scatter` of `Figure`s?

### DIFF
--- a/bqplot/marks.py
+++ b/bqplot/marks.py
@@ -839,6 +839,8 @@ class Label(Mark):
                    rtype='Number', min_dim=1, max_dim=1)
     rotation = NdArray(None, allow_none=True).tag(sync=True, scaled=True,
                        rtype='Number', min_dim=1, max_dim=1)
+    opacity = NdArray(None, allow_none=True).tag(sync=True, scaled=True,
+                      rtype='Number', min_dim=1, max_dim=1)
     x_offset = Int().tag(sync=True)
     y_offset = Int().tag(sync=True)
     scales_metadata = Dict({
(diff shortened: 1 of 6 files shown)


dmadeka:
Okay, so I've moved the `Scatter` groups to a uniform name, so that if we were to do `Objects` we could use a parent class where we lay out the groups in certain positions and then append the corresponding element in the base class. What do you think @ssunkara1 ?

Also, @SylvainCorlay  if we kept this design, any hope of passing a widget or DOM element to be appended at the position? I feel like this would be the natural evolution?

dmadeka:
@ssunkara1 @rmenegaux @SylvainCorlay Okay, I think I have a first cut.

@rmenegaux The move seems to have been broken, could you have a look for a quick second?

I made `font_size` an `Int` attribute and made `font_unit` the suffix to control what type of sizing. This makes the scale `default` value etc much easier to do.

dmadeka:
As per @rmenegaux comments, the `Objects` move should happen in this PR

dmadeka:
@ssunkara1 Fixed the Wealth notebook

ssunkara1:
@dmadeka Sorry for the delay in reviewing the PR. It looks good! Thanks.

I ran into one minor issue. In the first example, if I drag one label and then drag the second label, all the labels disappear. When I debug, I see that `model.mark_data` on the js side has a few elements which are `undefined`.

dmadeka:
Yeah, that's really strange. What's happening is that when the drag is called the second time, the index `i` that is being passed to it is totally different even though the `mark_data` is correct at the end of the first drag. Im puzzled

Any ideas @SylvainCorlay @rmenegaux ?

ssunkara1:
@dmadeka Raised a PR to your branch with the fix. Have a look

dmadeka:
Thanks again @ssunkara1

dmadeka:
@ssunkara1 good?

ssunkara1:
Yeah, looks good. Merging.

PR #224, opened 2016-07-22
fixing setup.py for windows

ssunkara1:
image

the `check_call` command works only for .exe files. .cmd files are not found by that unless the additional argument is passed and npm is not a .exe file. 

I checked that this works on windows and Ubuntu. Have not checked it on a mac.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -121,7 +121,9 @@ class NPM(Command):
 
     def has_npm(self):
         try:
-            check_call(['npm', '--version'], shell=True)
+            ## shell=True needs to be passed for windows to look at non .exe files.
+            shell = (sys.platform == 'win32')
+            check_call(['npm', '--version'], shell=shell)
             return True
         except:
             return False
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
Thanks. Shell = True is a big security risk on unix systems. We should probably only do this on windows.

@jasongrout comments?

ssunkara1:
I read that it is a risk if we are dynamically deciding the command to run directly from an external source. Since we are specifically running `npm` I do not think it is. But I do not know enough about this. :)

SylvainCorlay:
Same here. It is a Pavlov reflex w.r.t. shell=True

jasongrout:
See https://github.com/jupyter/jupyterlab/pull/144

ssunkara1:
Thanks @jasongrout Updated the PR.

SylvainCorlay:
@jasongrout we should do the same in 
 - pythreejs
 - ipyleaflet
 - widgetsnbextension
 - widget-cookiecutter
 - ipympl
 - nglview

jasongrout:
Yep, agreed.

jasongrout:
Perhaps we should make a standalone `npm.py` file in the same directory as `setup.py` that took care of common npm things?

PR #225, opened 2016-07-24
Complete transition to nodejs format for imports

SylvainCorlay:
There is not much in there beside changes in how things are imported.

Anything that touches js will conflict so we should merge quickly.


SylvainCorlay:
Ready.

PR #240, opened 2016-07-29
Fix dark theme for bqplot

SylvainCorlay:
dark


SylvainCorlay:
@dmadeka I am iterating on this since the dark theme is broken at the moment. Will cut a 0.7.2 shortly so merging.

dmadeka:
Wait, we should be careful before cutting a release

dmadeka:
Are we sure we don't want white axes?

dmadeka:
The ticks are not very visible in this theme now

PR #242, opened 2016-07-29
Lighter tick marks

SylvainCorlay:
screen shot 2016-07-29 at 5 26 10 pm

The color is the same as the one of the rendered markdown and button captions.


dmadeka:
Im good to merge but Ill have to test for the grid lines

PR #248, opened 2016-08-03
Bug fix with domain reverse

ssunkara1:
This is fixing an issue introduced in #237. 

With the previous change, if we update `reverse` from `True` to `False` nothing happens. Since the function was being called whenever `reverse` is changed, I think the logic of the function is fine.

With regards to the issue that @jmabille was facing, I ran into the issue before but I cannot reproduce it now. I think the issue (@jmabille correct me if I am wrong) was that when no data is passed to the `DateScale`, the default domain of the scale is from `[today, epoch date (1970, 1, 1)]`. That issue is slightly different and does not have much to do with the `reverse`. 

I marked the PR as WIP as I want the issue @jmabille faced to be fixed as part of this.

### DIFF
--- a/js/src/LinearScaleModel.js
+++ b/js/src/LinearScaleModel.js
@@ -28,13 +28,19 @@ var LinearScaleModel = scalemodel.ScaleModel.extend({
 
     initialize: function() {
         LinearScaleModel.__super__.initialize.apply(this, arguments);
+    },
+
+    set_init_state: function() {
         this.type = "linear";
         this.global_min = Number.NEGATIVE_INFINITY;
         this.global_max = Number.POSITIVE_INFINITY;
-        this.on_some_change(["min", "max"], this.min_max_changed, this);
-        this.min_max_changed();
+    },
+
+    set_listeners: function() {
         this.on("change:reverse", this.reverse_changed, this);
         this.reverse_changed();
+        this.on_some_change(["min", "max"], this.min_max_changed, this);
+        this.min_max_changed();
     },
 
     min_max_changed: function() {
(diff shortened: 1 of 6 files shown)


js/src/LinearScaleModel.js
```diff
@@ -47,7 +47,7 @@ var LinearScaleModel = scalemodel.ScaleModel.extend({
 
     reverse_changed: function() {
         this.reverse = this.get("reverse");
-        if(this.domain.length > 0 && this.reverse) {
+        if(this.domain.length > 0) {
```
SylvainCorlay: The problem with your change is that reverse is always called.

SylvainCorlay:
The issue that we were facing with Johan also happened with LinearScales which are not date scales.

js/src/LinearScaleModel.js
ssunkara1: But it reverses with respect to the previous state on change of the attribute reverse, which is fine. Also, with the previous change, if reverse is set to `False` from `True` nothing happens. If you can give me the code to reproduce the bug, it will be great.

jmabille:
The issue can be reproduced with the following code

```python
from IPython.display import display
import bqplot as bq

x_sc = bq.LinearScale(min=0.0, max = 1.8)
y_sc = bq.LinearScale(min=0.0, max = 1.8)

ax_x = bq.Axis(scale=x_sc)
ax_y = bq.Axis(scale=y_sc, orientation='vertical')

fig = bq.Figure(axes=[ax_x, ax_y])
display(fig)
```

SylvainCorlay:
👍

dmadeka:
everything looks good, LGTM

PR #250, opened 2016-08-04
Add null color for GridHeatMap

dmadeka:
@ssunkara1 @vedantam This should patch the issue, probably not the best way to handle it though


rmenegaux:
Merging this for now as a fix, although this should probably be addressed in `ColorScale` itself.

Issue #255, opened 2016-08-08
MarketMap raises a JavaScript error


dmadeka:
![image](https://cloud.githubusercontent.com/assets/7892382/17468639/e7fe44b6-5cf7-11e6-970a-8f0b1ed1b2dd.png)

PR #256, opened 2016-08-08
Market Map title, font and bug fix

dmadeka:
@ssunkara1 This adds a `font_style` and `title` attribute. Also, with the new `ipywidgets` you need to unpack the layout, so that the `model.state_change` isn't `undefined`.(see #255). I can also move the `update_data` to the model in this PR if you like but me and @vedantam probably need the font/title.

Also, the notebook was buggy since `ipywidgets` was imported before `from bqplot import *` which means that `bqplot`'s `Label` was being used rather than the `ipywidgets` one

### DIFF
--- a/js/src/MarketMapModel.js
+++ b/js/src/MarketMapModel.js
@@ -72,6 +72,7 @@ var MarketMapModel = basemodel.BaseModel.extend({}, {
         axes: { deserialize: widgets.unpack_models },
         tooltip_widget: { deserialize: widgets.unpack_models },
         style: { deserialize: widgets.unpack_models },
+        layout:  { deserialize: widgets.unpack_models }
     }, basemodel.BaseModel.serializers)
 });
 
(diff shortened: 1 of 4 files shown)


SylvainCorlay:
Looks good to me. Thanks for these bugfixes!

js/src/MarketMapModel.js
```diff
@@ -69,6 +72,7 @@ var MarketMapModel = basemodel.BaseModel.extend({}, {
         axes: { deserialize: widgets.unpack_models },
         tooltip_widget: { deserialize: widgets.unpack_models },
         style: { deserialize: widgets.unpack_models },
+        layout:  { deserialize: widgets.unpack_models }
```
SylvainCorlay: note: once we get rid of basemodel, this will come from the DOMWidget serializers.

dmadeka:
@ssunkara1 I guess we should `Backport` this

PR #258, opened 2016-08-08
Remove explicit jquery dependency

SylvainCorlay:
Using native DOM apis in Toolbar.js

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -25,7 +25,6 @@
     "webpack": "^1.12.14"
   },
   "dependencies": {
-    "bootstrap": "^3.3.5",
     "d3": "^3.5.16",
     "jquery": "^2.1.4",
     "jupyter-js-widgets": "^1.0.0",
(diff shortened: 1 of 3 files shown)


dmadeka:
I can test this, but let's not merge till that's done

SylvainCorlay:
👍

SylvainCorlay:
arg, the output_area_resize is an event baked in for the fullscreen hack.

dmadeka:
Along with the triple update_layout functions I believe

SylvainCorlay:
triple?

dmadeka:
```JavaScript
    update_layout: function() {
    update_layout2: function() {
    update_layout3: function() {
```

SylvainCorlay:
ok I see what you mean. The `update_layout` functions are required regardless of the resize event.

SylvainCorlay:
Ready for review. Not removing the `output_area_resize` for now, because it is coupled with bqnb. This will not be required anymore when we release ipywidgets 6.0 which will have the phosphor resize events.

PR #267, opened 2016-08-10
Market map title and font changes backport

ssunkara1:


dmadeka:
@ssunkara1 Good to merge both. What do you think?

PR #269, opened 2016-08-12
Minor bug fix in Market Map font styles

ssunkara1:
Right now, if there are more than 1 MarketMaps in the same notebook, changing the `font_style` for any one of the MarketMaps, changes the style for all of the MarketMaps.


dmadeka:
oops, d3.selectAll was bad

PR #272, opened 2016-08-15
Fixing issue with Pandas data frame trait

ssunkara1:
There was a bug in `PandasDataFrame` trait where we were assuming that the index is not named. This PR fixes that issue. Because of this, we were getting the issue in `MarketMap` when we click on the cell.

There is another reason why this issue surfaced only now. Previously in `ipywidgets`, only attributes whose value changed were sent across to the python side on call of `.touch` function. But now, in the first and second calls, many more attributes are sent across. @jasongrout is looking at the issue on the `ipywidgets` side.

### DIFF
--- a/bqplot/traits.py
+++ b/bqplot/traits.py
@@ -233,9 +233,9 @@ class PandasDataFrame(Instance):
         super(PandasDataFrame, self).__init__(*args, **kwargs)
 
     def _from_json(self, value, obj=None):
-        if value is not None:
+        if value is not None or value != []:
             df = pd.DataFrame(value)
-            df = df.set_index('index')
+            df = df.set_index(self.index_name)
             df.index.name = None
         else:
             df = pd.DataFrame()
@@ -243,6 +243,11 @@ class PandasDataFrame(Instance):
 
     def _to_json(self, df, obj=None):
         if df is not None:
+            # Name of the index column
+            self.index_name = df.index.name
+            if self.index_name is None:
+                # If the index is not named, reset_index() names the column 'index'
+                self.index_name = 'index'
             return df.reset_index().to_dict(orient='records')
         else:
             return []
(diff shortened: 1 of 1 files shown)


ssunkara1:
@SylvainCorlay From what I can see, in the first call to `touch` after the model is `initialized`, the attribute `_buffered_state_diff` of the `WidgetModel` has all the attributes of the Model and hence all the attributes are sent across to the python side. 

After the first two calls (investigating why the second call is affected), `_buffered_state_diff` seems to be working fine. Do you have any idea why this could be the case?

SylvainCorlay:
Are you using ipywidgets 5.x or master?

ssunkara1:
I was using `ipywidgets` master.

SylvainCorlay:
ok, master changes the way we define default values that may cause something like this. When we switch to ipywidgets 6, there `model.defaults` will have to be changed to a function...

jasongrout:
`model.defaults` doesn't *have* to change to a function. But anything that refers to the defaults from the base widget should use `_.result` (http://underscorejs.org/#result) or the equivalent, just like backbone does (http://backbonejs.org/docs/backbone.html#section-54). This is backwards compatible and covers the case where defaults is a function or an attribute.

So, for example, https://github.com/bloomberg/bqplot/blob/2d597baf9b3c581905f8d75f7b5c44ed09b7b4db/js/src/HandDrawModel.js#L22 would be (I think, without testing...)
```javascript
defaults: _.extend({}, _.result(widgets.WidgetModel.prototype, 'defaults'), {
```

SylvainCorlay:
Yeah but to be more homogeneous across the board, I would prefer it to always be a function.

jasongrout:
sure, but in the interim, downstream libraries need to have a backwards compatible solution. So even if they move to having a function, they should still use something like _.result to get the upstream values.

PR #204, opened 2016-07-13
Changing the json format of date objects

ssunkara1:

Issue #264, opened 2016-08-09
Required numpy version


SylvainCorlay:
@ssunkara1 what is the minimal version of numpy for your recent changes in date serialization?

It would be nice to add it to the conda recipe and setup.py.

PR #273, opened 2016-08-15
Date trait bug fix

ssunkara1:
This shows up in the `Interaction Layer.ipynb` example notebook. This is to account for the change in the `numpy datetime64` string representation change. This should've been fixed #204 but I missed it.

### DIFF
--- a/examples/Interactions/Interaction Layer.ipynb
+++ b/examples/Interactions/Interaction Layer.ipynb
@@ -552,7 +552,7 @@
    },
    "outputs": [],
    "source": [
-    "br_sel_dt.observe(brush_callback, names=['brushing'])"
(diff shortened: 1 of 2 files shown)


examples/Interactions/Interaction Layer.ipynb
```diff
@@ -961,26 +961,23 @@
   }
  ],
  "metadata": {
+  "anaconda-cloud": {},
```
SylvainCorlay: We probably don't need this part of the diff.

SylvainCorlay:
Adding a version requirement for numpy to close #264 in both meta.yml and setup.py would be great!

SylvainCorlay:
I merged as a separate PR with the numpy version requirement in setup.py (including your commit)

PR #281, opened 2016-08-16
NdArray Trait Fixes

dmadeka:
@ssunkara1 @SylvainCorlay Here is the change. Big thanks to @ssunkara1 , but @SylvainCorlay you should test

### DIFF
--- a/bqplot/traits.py
+++ b/bqplot/traits.py
@@ -142,6 +142,26 @@ class NdArray(CInstance):
         else:
             return np.asarray(value)
 
+    def set(self, obj, value):
+        #TODO: We shouldnt have to overload the set because
+        # numpy doesn't support == comparions
+        new_value = self._validate(obj, value)
+        try:
+            old_value = obj._trait_values[self.name]
+        except KeyError:
+            old_value = self.default_value
+
+        obj._trait_values[self.name] = new_value
+        try:
+            silent = np.array_equal(old_value, new_value)
+        except:
+            # if there is an error in comparing, default to notify
+            silent = False
+        if silent is not True:
+            # we explicitly compare silent to True just in case the equality
+            # comparison above returns something other than True/False
+            obj._notify_trait(self.name, old_value, new_value)
+
     def validate(self, obj, value):
         # If it is an object, I have to check if it can be cast into a date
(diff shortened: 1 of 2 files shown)


bqplot/traits.py
```diff
@@ -135,7 +135,9 @@ def _to_json(a, obj=None):
             return {'values': a, 'type': None}
 
     def _set_value(self, value):
-        if value is None or len(value) == 0:
+        if isinstance(value, np.ndarray) and len(value.shape) == 0:
+            return np.asarray(None)
```
SylvainCorlay: why not return value? np.array(1) is not a correct value?

bqplot/traits.py
```diff
@@ -148,12 +150,16 @@ def validate(self, obj, value):
...
         min_dim = self.get_metadata('min_dim', 0)
         max_dim = self.get_metadata('max_dim', np.inf)
+        allow_none = getattr(self, 'allow_none', False)
         shape = np.shape(value)
         dim = 0 if value is None else len(shape)
         if (dim > 1) and (1 in shape):
             value = np.squeeze(value) if (self.squeeze) else value
             dim = len(np.shape(value))
-        if dim > max_dim or dim < min_dim:
+        if allow_none is True and dim == 0:
```
SylvainCorlay: shouldn't `self.allow_none`  always be defined?

dmadeka:
@SylvainCorlay yup, you're right. made the changes

dmadeka:
Okay @ssunkara1 Noticed the buffered state issue, and pointed out it may be related to the way we compare `numpy` arrays, @SylvainCorlay pointed to set which is the issue, I overload it here, but ideally we shouldnt have to do this

bqplot/traits.py
```diff
@@ -135,13 +135,33 @@ def _to_json(a, obj=None):
...
+        except KeyError:
+            old_value = self.default_value
+
+        obj._trait_values[self.name] = new_value
+        try:
+            silent = np.array_equal(old_value, new_value)
+        except:
+            # if there is an error in comparing, default to notify
+            silent = False
+        if silent is not True:
```
SylvainCorlay: we can probably simplify this into `if not np.array_equal(old_value, new_value)`

PR #286, opened 2016-08-17
Bug fixes for default values

ssunkara1:
Previously we were passing `() or `(0,)` as args to the constructor of `NdArray` and `PandaDataFrame` traits which was setting the default value. 

Also includes a bug fix for `_from_json` of `PandasDataFrame` trait.


SylvainCorlay:
LGTM

Issue #290, opened 2016-08-19
NdArray default_value does nothing


rmenegaux:
The `default_value` of the `NdArray` trait does not work, it is not taken into account at all. 

```python
from ipywidgets import Widget
from bqplot.traits import NdArray

class A(Widget):
    ''''''
    x = NdArray(default_value=[1, 2]).tag(sync=True, min_dim=1)
    
a = A()
```
The above code fails because of a dimension mismatch (dim < min_dim), as somehow `validate` is called with `None` as an argument.

```python
class B(Widget):
    ''''''
    x = NdArray(default_value=[1, 2]).tag(sync=True)
    
b = B()
b.x
```
returns `None`

We are looking into this with @ssunkara1

SylvainCorlay:
The base class (Instance) does not take a default value.

Although passing `args=[default_value]` to `super.__init__` should do the job for now. Eventually we should stop inheriting from Instance.

dmadeka:
@SylvainCorlay I thought the `Instance` `__init__` called super which should call the `TraitType` init and the **metadata that is passed should set the default_value?

PR #291, opened 2016-08-19
Fixing issue with default_value for custom traits

ssunkara1:
@rmenegaux @SylvainCorlay

### DIFF
--- a/bqplot/traits.py
+++ b/bqplot/traits.py
@@ -261,12 +261,12 @@ class PandasDataFrame(Instance):
 
     def __init__(self, *args, **kwargs):
         default_value = kwargs.pop('default_value', None)
-        # The first entry in args should be the `klass` or else self.klass
-        # is assigned as the class. The second entry should be a tuple whose
+        # The first entry in args should be the `klass`. If it is None, self.klass
+        # is assigned as the klass. The second entry should be a tuple whose
         # first element is passed to the default constructor of the klass.
         # Hence, we should set the following as the args
         if len(args) == 0:
-            new_args = args + (None, (default_value,))
+            new_args = (None, (default_value,))
         super(PandasDataFrame, self).__init__(*new_args, **kwargs)
         self.tag(to_json=self._to_json, from_json=self._from_json)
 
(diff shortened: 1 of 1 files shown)


ssunkara1:
This addresses issue #290

bqplot/traits.py
```diff
@@ -145,6 +145,8 @@ def _set_value(self, value):
     def set(self, obj, value):
         #TODO: We shouldnt have to overload the set because
         # numpy doesn't support == comparions
+        #TODO: This should also be removed once _buffered_state_diff is
+        # fixed in ipywidgets
```
SylvainCorlay: I  don't agree with this comment. The fact that NdArray emits an event on any assignment (even when the new value is equal to the old one) is a bug regardless of the _buffered_state_diff issue in ipywidgets.

This issue only exposed the problem with nd array.

bqplot/traits.py
```diff
@@ -186,12 +188,23 @@ def _from_json(value, obj=None):
...
+    # default constructor with args as the first argument and kwargs
+    # as the second argument. The constructor of np.ndarray
+    # requires shape, data type and stride to be passed.
+    def make_dynamic_default(self):
+        return self._default_value
+
     def __init__(self, *args, **kwargs):
         self.squeeze = kwargs.pop('squeeze', True)
+        default_value = kwargs.pop('default_value', None)
+        self._default_value = self._cast(default_value)
```
SylvainCorlay: why not only doing
```
if 'default_value' in kwargs:
    kwargs['default_value'] = self._cast(default_value)
```
and in `make_dynamic_default`
```
return self.default_value
```

bqplot/traits.py
ssunkara1: Well, previously, `_buffered_state_diff` only returned what trait changed. Now for the first two calls it returns everything. 

The behavior of `_buffered_state_diff` should be consistent with every call I feel. It should either return everything always or return only the changed things but not change depending on if it is the first call or the fourth.

SylvainCorlay: I agree that there is an issue with `_buffered_state_diff`, but the fact that the numpy array emits an event on any assignment is a bug regardless.

Therefore, even when we have fixed the `_buffered_state_diff` thing, we should keep this.

ssunkara1:
@SylvainCorlay Yeah, I agree it is better to have the `set` over-ridden even when `_buffered_state_diff` changes. I removed that comment.  Made the other change with `default_value` as well.

ssunkara1:
Do not merge as of now. Still testing all edge cases. I will let you know when I am good.

Thanks.

ssunkara1:
@SylvainCorlay Ready to be reviewed.

Thanks.

PR #292, opened 2016-08-19
Set default data attributes of marks to []

rmenegaux:
The default value of the principal data traits of marks (`x` and `y` for Lines) is `None`, even though they do not allow `None`.
We should be able to first create the Mark instance, and then set the data attributes programmatically.

This depends on #291


ssunkara1:
Thanks @rmenegaux This looks good.

PR #293, opened 2016-08-20
Remove uses of IPython.display.display

SylvainCorlay:
This removes all the uses of `IPython.display.display` and only uses the displaying of the last computed value as it was suggested by Fernando.

### DIFF
--- a/examples/Advanced Plotting/Animations.ipynb
+++ b/examples/Advanced Plotting/Animations.ipynb
@@ -23,8 +23,7 @@
    "outputs": [],
    "source": [
     "import numpy as np\n",
-    "import bqplot as bq\n",
-    "from IPython.display import display"
+    "import bqplot as bq"
    ]
   },
   {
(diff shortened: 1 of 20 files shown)


examples/Advanced Plotting/Animations.ipynb
```diff
@@ -23,8 +23,7 @@
    "outputs": [],
    "source": [
     "import numpy as np\n",
-    "import bqplot as bq\n",
-    "from IPython.display import display"
+    "import bqplot as bq"
```
dmadeka: Can we drop the as bq here too? Sorry I thought I did that

SylvainCorlay:
Done.

dmadeka:
@sylvaincorlay thanks! I'm only worried that people won't realize they can use display. Should we leave one example with markdown text saying that they can use display to display the plot?

SylvainCorlay:
Although that is not really the point of bqplot... It is a jupyter thing.

dmadeka:
Okay, so no display? Is there ever any use for display? apart from displaying one widget and then another without using a VBox? Do we ever do that in the examples?

SylvainCorlay:
We don't do that in the examples. I removed the double displays...

dmadeka:
Id say keep one example. If we want to display and then change without running two cells its useful

dmadeka:
+1

PR #294, opened 2016-08-22
Disable user select and set default cursor on axis ticks

SylvainCorlay:
In the transition to native DOM api calls, the front `-` went missing. Moving this to the css, and adding a rule for the cursor.


SylvainCorlay:
@dmadeka I will automerge in a few.

dmadeka:
Sure I wasn't sure what this one was doing. Sorry, was out this weekend otherwise I wouldve reviewed the others.

SylvainCorlay:
It prevents the axis ticks to be selectable, which was causing issues with certain selectors, and makes the cursor the standard one when hovering the ticks.

PR #295, opened 2016-08-22
Fix OHLC

SylvainCorlay:
@ssunkara1 the recent changes on NdArray broke OHLC. The default value for `y` should be 2-D. hence I picked `[[]]`. Besides the default for squeeze was True which was causing `[[]]` to be squeezed to `[]`.

### DIFF
--- a/bqplot/marks.py
+++ b/bqplot/marks.py
@@ -329,11 +329,11 @@ class Lines(Mark):
     name = 'Lines'
 
     # Scaled attributes
-    x = NdArray(default_value=[]).tag(sync=True, min_dim=1, max_dim=2,
+    x = NdArray(default_value=[], squeeze=True).tag(sync=True, min_dim=1, max_dim=2,
                 scaled=True, rtype='Number', atype='bqplot.Axis')
-    y = NdArray(default_value=[]).tag(sync=True, min_dim=1, max_dim=2,
+    y = NdArray(default_value=[], squeeze=True).tag(sync=True, min_dim=1, max_dim=2,
                 scaled=True, rtype='Number', atype='bqplot.Axis')
-    color = NdArray(None, allow_none=True).tag(sync=True, scaled=True,
+    color = NdArray(None, allow_none=True, squeeze=True).tag(sync=True, scaled=True,
                     rtype='Color', atype='bqplot.ColorAxis', min_dim=1, max_dim=1)
 
     # Other attributes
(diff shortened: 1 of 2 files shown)


ssunkara1:
I think we want `squeeze=True` for `Lines`. I will check that and get back to you.

SylvainCorlay:
We could set it to True in Lines. I don't think that setting it to True by default is what a regular user would expect here.

SylvainCorlay:
@ssunkara1 I would like to push a release for today's build.

ssunkara1:
@SylvainCorlay  I think in `OHLC` we should make `min_dim = 2` and set `default_value = [[]]`. Because if I pass a single dimensional array to `OHLC` it just shows up a blank chart. 

Also, I think `squeeze` should be `True` by default. Because, without that, none of the `Marks` will plot arrays of shape `(n, 1)`. It will either not show anything as in the case of `Lines` or throw a dimension mismatch in the case of `Scatter` etc. For `OHLC`, we would have to pass `squeeze=False`. 

I mean, we have had `squeeze = True` from almost the start and I do not see many people complaining about it. I think not being able to plot arrays of shape `(n, 1)` will be a bigger concern as matplotlib etc handle it. And, about changing the thing they are passing, we also do it in the case they pass a list. We convert it to an numpy array. So in my opinion, we can live with it.

SylvainCorlay:
Did you see any issue in the example notebooks and other usages that we have in bqplot?

Regarding the squeeze attribute, `squeeze = True` results in a non trivial transformation that returns a new copy of the array... So this should be a conscious opt-in feature. If we are ever going to separate the traits from bqplot that is eventually going to be an issue.

ssunkara1:
The example notebooks do not have an issue. 

But if you try to plot any array of shape `(n, 1)` it fails. I feel it is bad if we can't do that. And I do not think it is a non-trivial transformation. It is collapsing all dimensions of length 1. When I want to plot something of shape `(n, 1)` the transformation is exactly what I would expect. `(n, 1)` occurs a lot more frequently than we would like. Especially `PandasDataFrame.values` returns things of shape `(n, 1)` quite a lot.

SylvainCorlay:
grumbling

SylvainCorlay:
How about I set squeeze=True in most NdArray trait types?

ssunkara1:
We can ask what others feel about this. I don't want to impose anything. @dmadeka @rmenegaux @jasongrout @ChakriCherukuri what do you guys this?

ssunkara1:
I am good with setting `squeeze=True` for all `NdArray` traits apart from `OHLC` and `BoxPlot`.

dmadeka:
@ssunkara1 @SylvainCorlay Im good with `squeeze=True` for most NdArray Traits as well. Lets do that and try to cut a release so we can test?

rmenegaux:
@ssunkara1 @SylvainCorlay I'm good with truing the squeezing as well

SylvainCorlay:
Pushed the `squeeze=True` everywhere in marks but boxplot's `y`.

SylvainCorlay:
@ssunkara1 merge?

ssunkara1:
Cool Thanks

PR #297, opened 2016-08-23
Simplify Array trait type

SylvainCorlay:
- [x] make `to_json` and `from_json` free functions and pass them to `tag` in the marks

### DIFF
--- a/bqplot/traits.py
+++ b/bqplot/traits.py
@@ -84,6 +84,42 @@ class Date(TraitType):
             return value
 
 
+def array_from_json(value, obj=None):
+    if value is not None:
+        array_dtype = {
+            'date': np.datetime64,
+            'float': np.float64
+        }.get(value.get('type'), object)
+        return np.asarray(value['values'], dtype=array_dtype)
+
+def array_to_json(a, obj=None):
+    if a is not None:
+        if np.issubdtype(a.dtype, np.float):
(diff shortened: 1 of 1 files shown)


bqplot/traits.py
```diff
@@ -98,110 +84,83 @@ def from_json(value, obj=None):
...
         if (dim > 1) and (1 in shape):
-            value = np.squeeze(value) if (self.squeeze) else value
+            value = np.squeeze(value) if self.squeeze else value
             dim = len(np.shape(value))
         if self.allow_none and dim == 0:
             return value
+        # validate dimension
         if (dim > max_dim or dim < min_dim):
-            raise TraitError("Dimension mismatch")
+            raise TraitError('Dimension mismatch: ' + str(dim))
```
dmadeka: Shouldnt we add more to this?

`Dimension mismatch: 0` doesn't help a lot no?

SylvainCorlay: ah that is a debug thing. there was no intension of adding it.

SylvainCorlay:
I have another PR coming on the top of this one that makes dtype and order kwargs instead of metada and remove the call to tag so that you have to call `.tag(sync=True, **array_serialization)`, but this first step should not be disruptive.

SylvainCorlay:
Closing for another coming PR.

PR #299, opened 2016-08-24
Move Index to Site Map

dmadeka:
@madridnick @SylvainCorlay This is a move to remove the confusion created by the Index name. Ive moved Sylvain's talk to the Introduction. This also helps because, the display of the YouTube video needs the cell to be run, which the structure of `Index.ipynb` didn't lend to, but `Introduction.ipynb` does.

Im open to other suggestions as well.


madridnick:
LGTM

SylvainCorlay:
what was the confusion about `Index.ipynb`?

dmadeka:
@SylvainCorlay Just that its not consistent with other docs, I do like `Site Map`. Is there any reason we should keep `Index`, like binder etc?

SylvainCorlay:
@madridnick @dmadeka I think that we should keep the `index.ipynb` name for the entry point. This is the current convention for the mybinder project, and will therefore become the expected entry point.

PR #301, opened 2016-08-24
Domain not updating bug fix

ssunkara1:


SylvainCorlay:
I will leave you guys test this.

rmenegaux:
@ssunkara1 `Scatter` doesn't seem to have the problem, probably because it does not check `this.mark_data.length === 0` in `update_domains`. Could it be a simpler fix to just remove this condition in the other mark models?
```python
import bqplot.pyplot as plt

plt.figure()
l1 = plt.scatter([0, 1], [0, 1])
l2 = plt.scatter([0, 10], [10, 10])
plt.show()

l2.y = []
```

SylvainCorlay:
>  Could it be a simpler fix to just remove this condition in the other mark models?

Good point.

ssunkara1:
Thanks @rmenegaux I removed the check for length.

PR #303, opened 2016-08-25
Counties map and election app

dmadeka:
What does everyone think? Any ideas on a cool election notebook?

So far I did, fraction of votes for each candidate. Looks like:

image


madridnick:
@dmadeka, any reason why this image is predominately red? Red = republican?

dmadeka:
@madridnick The data etc were from [here](https://www.kaggle.com/benhamner/2016-us-election), so I just tried to recreate it. 

But, I tried `Blues` and we might have a bug @ssunkara1

dmadeka:
Will open a PR for the bug in `Blues`:

image

dmadeka:
This is the updated example with Pollster. Have a look at the notebook and see if it runs for you.

@ssunkara1 @madridnick @SylvainCorlay 

image

dmadeka:
The county-example still has a few bugs in the numerical codes that are assigned to counties by the census, ill have to decode those.

dmadeka:
@SylvainCorlay This is ready for review

SylvainCorlay:
I love it. Can you make the `min_width` and `min_height` of ts_fig smaller (e.g. 400/250).

dmadeka:
Okay, done. Also, switched the primaries to 2012-election results from the FEC:

image

dmadeka:
@SylvainCorlay Added the license file with proper attribution (maybe needs a review from @kpfleming )

dmadeka:
It seems this is pretty difficult to merge. Maybe we should just strip the county map?

kpfleming:
The license/attributions file looks OK, although if the license text from each source can be directly incorporated that would be better since that would eliminate concerns about those target URLs ever becoming dead.

SylvainCorlay:
> The license/attributions file looks OK, although if the license text from each source can be directly incorporated that would be better since that would eliminate concerns about those target URLs ever becoming dead.

Thanks fir dropping by. We were thinking of completely removing the csvs for the election results and have the user of the notebook run the pip install command (in a cell) to install the pollster library voluntarily...

dmadeka:
>  The license/attributions file looks OK, although if the license text from each source can be directly incorporated that would be better since that would eliminate concerns about those target URLs ever becoming dead.

> Thanks fir dropping by. We were thinking of completely removing the csvs for the election results and have the user of the notebook run the pip install command (in a cell) to install the pollster library voluntarily...

Unfortunately, the 2012 County results have to be compiled from a range of csv's. I could have them download the csv's from the web and then compile the DataFrame. But that's the tough one to exclude the csv on.

SylvainCorlay:
> Unfortunately, the 2012 County results have to be compiled from a range of csv's. I could have them download the csv's from the web and then compile the DataFrame. But that's the tough one to exclude the csv on.

gotcha

dmadeka:
The county by county election results is also from HuffPost, so it's CC-BY-NC-SA as well. What's the final word on this? Should I change the license files and merge? Or per @SylvainCorlay the csv files License is incompatible with the bqplot LICENSE. Should I just merge the notebook plus the county map and forget the csv's?

@kpfleming @SylvainCorlay

dmadeka:
Okay, the file is now from https://catalog.data.gov/dataset/2008-presidential-general-election-county-results-direct-download .. The LICENSE should be good and the csv is removed, so the notebook imports the data and cites the python pollster.Is that okay?

@SylvainCorlay @kpfleming

SylvainCorlay:
This is fine with me. Merging as-is. 

Note: we might be able to drop the license section on pollster since pollster is not distributed as part of the package or source code or with the notebook...

dmadeka:
Okay, should we have a disclaimer that this notebook installs a package?

SylvainCorlay:
maybe. I have not given it any thoughts.

Issue #307, opened 2016-08-30
Race condition with setting scales


ssunkara1:
When `scales` of a `Mark` are changed, till the time the `scale` views are created, setting any of the data attributes results in an error and makes the `Mark` non-responsive.

SylvainCorlay:
Does this solve the issue occuring when updating data attributes at the same time as scales?

ssunkara1:
This was fixed by #308.

PR #308, opened 2016-08-30
Fixing race condition

ssunkara1:
We can't clear `this.scales` in `Mark` view before the promise callback because changes to other traits can happen before the promise is resolved, mark views will throw an error. (We assume `this.scales` is always set)


SylvainCorlay:
Does this solve the issue occuring when updating data attributes at the same time as scales?

ssunkara1:
@SylvainCorlay are you talking about #307 ? This does fix that issue.

PR #311, opened 2016-09-01
World Map Unicode Fixes

dmadeka:
@SylvainCorlay @fm75 might fix the ascii issue, but im not sure if this is the right way to go. Essentially, I change the non-utf-8 characters from say Aland Islands, Cote D'Ivoire Reunion etc


dmadeka:
This isn't a good idea. d3 uses $\lambda$ etc in it's code, if its not the WorldMap its something else. Can we specify a default encoding for a module?

fm75:
Closing this seems 100% wrong. If unicode in the python source is causing installation failures without requiring the installer (person who is doing the install), to configure his environment, then bqplot is doomed. IMO

dmadeka:
Fair enough.

SylvainCorlay:
I don't think that the worldmap's json changes anything.

fm75:
[Issue 36](https://github.com/bloomberg/bqplot/issues/36) 
The failure to install properly in a Docker container seems repeatable. Two people independently have experienced the same failure. I outlined a work-around that allows it to install properly. While this could be implemented in a chef/ansible/... recipe for a jupyterhub install, it still seems likely that others will encounter the same difficulty.

I would recommend stripping out as much of the bqplot code as possible to establish a smaller repeatable error and track down and fix all encoding errors in the data. 

Until this is fixed, perhaps a work-around note in the README.md would be appropriate.

SylvainCorlay:
@fm75 I agree but this PR does not change bqplot's code. It changes the content of a json data file.

SylvainCorlay:
Actually, I seems that I was wrong. Dhruv's fix might just be what we need.

@fm75 do you have a simple way to test this?

SylvainCorlay:
I successfully installed bqplot in a docker instance from both the current master branch and pypi. So I have not been able to reproduce the issue.

However since multiple people have reported it and it is the only place in the code base where we have non-ascii characters, I am inclined to merge this.

PR #319, opened 2016-09-15
Require pollster 0.1.6

SylvainCorlay:


examples/Applications/Visualizing the US Elections.ipynb
```diff
@@ -38,7 +38,7 @@
     "    print('Pollster not found. Installing Pollster..')\n",
     "    import pip\n",
     "    try:\n",
-    "        pip.main(['install', 'pollster'])\n",
+    "        pip.main(['install', 'pollster==0.1.6'])\n",
```
dmadeka: Does it support >=?

SylvainCorlay: Yes, but pollster does not seem to be super maintained... So we should probably pin it for now.

PR #328, opened 2016-09-20
Map Refactor

dmadeka:
Comes on top of #325 and allows the `Map` mark to take arbitrary GeoJSON files. The next step is consistent `selected` and `hovered` theming.

The additional list containing names was moved in favor of a properties dict for each subunit that contains any properties the `Tooltip` may be used to display. So for example if a GeoJSON file has the property 'Capital' it can be accessed through the tooltip.

Major API change, will need testing.


dmadeka:
@ssunkara1 ping. Should allow arbitrary GeoJSON provided the arcs are under the key, subunits

SylvainCorlay:
:+1:

PR #331, opened 2016-09-23
Splitting Scatter

rmenegaux:
This abstracts out the common code of `Scatter` and `Label` into `ScatterBase`, which basically places elements on x,y scales, and handles interactions. `Scatter` and `Label` are then left with the drawing of the individual elements. This makes way for future marks, like Chernoff faces :smile:

Changes that are visible from a user perspective:
- Renamed Scatter `default_colors` to `colors`, which is consistent with Lines and Labels.
- If no scales are passed to `Scatter`, it plots directly on the `Figure` scales
- Label now handles hovering and selecting
- `apply_styles` was called twice on hover. This fixes it.


SylvainCorlay:
👍 this is great! thanks for tackling this one.

dmadeka:
Why is this still a WIP?

ssunkara1:
@rmenegaux There is a very minor issue with legend positioning for which I raised a PR to your branch. And this will also need a rebase.

After that, I am good to merge. Looks Great!! :+1:

rmenegaux:
@ssunkara1 All set

ssunkara1:
@rmenegaux Thanks! There is one small issue with defaults. https://github.com/rmenegaux/bqplot/pull/2

Issue #330, opened 2016-09-23
Issue with updating Mark attributes


ssunkara1:
When a mark attribute (e.g., `selected`) is being set from the `javascript` side (because of a `python` change) and then after that, a second attribute (e.g., `x`) are set from the `python` side, the value of `x` is reset to it's previous value.  To reproduce the issue, the following code can be used.

``` python
import numpy as np

from IPython.display import display
import bqplot.pyplot as plt
from bqplot.interacts import BrushIntervalSelector
import time

plt.figure()
line = plt.plot([], [])
fig = plt.current_figure()

intsel = BrushIntervalSelector(scale=plt.get_context()['scales']['x'],  marks=[line], line_width=1)
fig.interaction = intsel
display(fig)

def plot_timeseries():
    time.sleep(1)
    x, y = np.arange(100), np.random.rand(100)     
    with line.hold_sync():
        line.x, line.y = x, y

plot_timeseries()
```

ssunkara1:
This is fixed in `ipywidgets` `6.0.0b1`.

Issue #327, opened 2016-09-20
Issue with multiple histograms on the same scale


ssunkara1:
When two histograms are on the same `sample` scale, and both are updated in quick succession, the second histogram's `sample` is not updated. Still trying to narrow down on the cause of the issue. Will update the issue with code once I narrow down the code to reproduce it consistently.

SylvainCorlay:
Does this happen only when there is a transition?

ssunkara1:
This is fixed in `ipywidgets` `6.0.0b1`.

PR #332, opened 2016-09-23
Temporary sync fix

ssunkara1:
This addresses issue #330. I believe this also addresses issue #327.

I think, the root issue is that in `WidgetModel` the attribute `_buffered_state_diff` is not set correctly on initialization. So when the first `sync` from `javascript` happens, the `change` handlers of all the attributes are notified. Also, if any of the other attributes changed, they are being overwritten by the old values which came back in the comm message. 

The fix is temporary. I am not sure if I want this to be merged. It is just trying to make sure that what I think as the root issue is actually the issue.

### DIFF
--- a/js/src/BaseModel.js
+++ b/js/src/BaseModel.js
@@ -23,6 +23,14 @@ var BaseModel = widgets.WidgetModel.extend({
         _model_module: "bqplot"
     }),
 
+    initialize: function() {
+        BaseModel.__super__.initialize.apply(this, arguments);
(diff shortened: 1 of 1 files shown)


ssunkara1:
@SylvainCorlay @jasongrout

ssunkara1:
@SylvainCorlay @jasongrout did you get a chance to look at this?

ssunkara1:
This has been fixed in `ipywidgets` `6.0.0b1`.

PR #336, opened 2016-09-25
Jlab

SylvainCorlay:
Updated version of #324 for demo. Superseeded by #337.

PR #337, opened 2016-09-25
Phosphor resize and jupyterlab integration.

SylvainCorlay:
This comes on the top of #336 . The point is to use phosphor resize events for dynamic layouts.


jasongrout:
Superseded by #348.

Issue #338, opened 2016-09-26
Scatter and Label drag issue


ssunkara1:
When `restrict_x` or `restrict_y` is set to `True`, `x` and `y` attributes of `Scatter` and `Label` do not update.

```python
import numpy as np
import bqplot as bq


def scatter_updated(data):
    with line.hold_sync():
        line.x = [scat.x[0], scat.x[-1]]
        line.y = [np.mean(scat.y), np.mean(scat.y)]

scat_scales = {'x': bq.LinearScale(), 'y': bq.LinearScale()}
scat = bq.Scatter(x=x, y=y, scales=scat_scales, enable_move=True, restrict_y=True)
line = bq.Lines(x=[x[0], x[-1]], y=[np.mean(y), np.mean(y)], scales=scat_scales)
ax_x, ax_y = bq.Axis(scale=scat_scales['x']), bq.Axis(scale=scat_scales['y'], orientation='vertical')

scat.observe(scatter_updated, ['x', 'y'])
bq.Figure(marks=[scat, line], axes=[ax_x, ax_y])

PR #339, opened 2016-09-26
Drag bug fix in Scatter and Label

ssunkara1:
Fixes issue #338 

Thanks Daniel for pointing this out.

### DIFF
--- a/js/src/Scatter.js
+++ b/js/src/Scatter.js
@@ -973,8 +973,8 @@ var Scatter = mark.Mark.extend({
         var restrict_x = this.model.get("restrict_x"),
             restrict_y = this.model.get("restrict_y");
         if (restrict_x && restrict_y) { return; }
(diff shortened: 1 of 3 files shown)


dmadeka:
Can you clean up the commit history?

ssunkara1:
@dmadeka Yeah was in the process of doing that. Now it is done.

js/src/Scatter.js
```diff
@@ -993,6 +993,8 @@ var Scatter = mark.Mark.extend({
             // saving on move if flag is set
             this.update_array(d, i);
         }
+        var i = 0;
+        i = i + 1;
```
dmadeka: Let's make it one commit so this isn't there

dmadeka: Because you delete it in the second commit

ssunkara1: @dmadeka Merged the two commits.

PR #343, opened 2016-10-04
(WIP) First pass at tick angles

dmadeka:
Introduces a `rotate_angle` attribute to each Axis that controls each tick's rotation.

@ssunkara1 @madridnick This should solve the issue of tick's not being vertical. I need to resolve the margin, perhaps by allowing a `tick_offset` dict that lets the user decide.

### DIFF
--- a/bqplot/axes.py
+++ b/bqplot/axes.py
@@ -27,7 +27,7 @@ Axes
    ColorAxis
 """
 
-from traitlets import Int, Unicode, Instance, Enum, Dict, Bool
+from traitlets import Int, Unicode, Instance, Enum, Dict, Bool, Float
 from traittypes import Array
 from ipywidgets import Widget, Color, widget_serialization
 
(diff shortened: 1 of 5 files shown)


SylvainCorlay:
💯

SylvainCorlay:
The build failure comes from the latest pandas release. Please disregard.

SylvainCorlay:
I have fixed the travis build in master. FYI

ssunkara1:
@dmadeka The PR looks good to me! Can you rebase? I am good to merge it!

dmadeka:
@ssunkara1 Slipping mate, the issue is if the text is very long the displacement is too big. It needs multi-line text wrapping as well. Also, auto-angle calculation would be nice no?

ssunkara1:
Yeah I agree that multi-line text wrapping and auto-angle calculation would be nice. I have a feeling that in svg detecting when the text overlaps might be hard. With the new smaller default size, the text even for mm-dd-yy format overlaps frequently. That is why I was thinking we can merge this as it is and then work on the multi-line and automatic angle calculation. What do you think?

dmadeka:
@ssunkara1 The rotation is too buggy, because the displacement of the (start/mid/end) of the text (depending on the angle) from the tick marker isn't consistent. I will try to do this before SciPy.

As a proposal for a fix, since Bruno has complained about the small font size for the tick markers, we make that a parameter and increase the default Figure size. @SylvainCorlay , opinions?

dmadeka:
@SylvainCorlay go for it. Im still out on contributing directly

rmenegaux:
This was dealt with in PR #669, closing.

PR #351, opened 2016-10-26
HeatMap

rmenegaux:
Basic HeatMap mark, that plots 2d `color` data against 1d `x` and `y`.

`HeatMap` uses canvas to draw a rectangle of color `color_{i,j}` "centered" around each point of coordinates `(x_i, y_j)`

It can be used as an Image mark too.

### DIFF
--- a/bqplot/marks.py
+++ b/bqplot/marks.py
@@ -1170,7 +1170,7 @@ class GridHeatMap(Mark):
         `start` aligns the column values passed to be aligned with the start of the
         tiles and `end` aligns the column values to the end of the tiles.
     anchor_style: dict (default: {'fill': 'white', 'stroke': 'blue'})
-        Controls the style for the element which serves as the anchor duringi
+        Controls the style for the element which serves as the anchor during
         selection.
 
     Data Attributes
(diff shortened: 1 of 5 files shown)


dmadeka:
GridHeatMap and HeatMap? Can you detail this PR more?

rmenegaux:
@dmadeka `HeatMap` is to be used when there is too much data for the `GridHeatMap`. The latter makes every cell a DOM element, which is certainly good for interactivity and flexibility, but will crash the browser if there are too many cells.
`HeatMap` uses canvas to draw the cells, which allows for a much greater number of elements. This is at the expense, however, of the interactivity of svg, and as of now I have not coded in the tooltip and other interactions.
It is still WIP as there are some padding issues I need to fix.

dmadeka:
Can we call it `Image` ala mpl?

dmadeka:
also, would be cool to be able to do this:

http://bl.ocks.org/mbostock/b75883984a032ea0ba26

dmadeka:
@rmenegaux Still a WIP?

rmenegaux:
This works fine as far as I can see. I just haven't figured out a good way to do the padding yet. As of now it has none.

dmadeka:
@rmenegaux Can you open an issue?

dmadeka:
@rmenegaux Can you add a documented example notebook?

rmenegaux:
:+1:

PR #352, opened 2016-10-31
Update for ipywidgets 6.0 beta

SylvainCorlay:
@jasongrout

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ install:
   - conda update -q conda
   # Useful for debugging any issues with conda
   - conda info -a
+  - conda install pyyaml
   - conda env create -f test-environment.yml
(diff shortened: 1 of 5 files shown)


SylvainCorlay:
Had to add yaml to the the testing miniconda...

jasongrout:
Seems to work well. I have an issue with the tooltip graph in the second example of the Interaction/Mark Interaction example. The tooltip graph doesn't show up, just a transparent black rectangle. But it seems to do that on master too, so I'll merge this and we can debug it as a separate issue.

ssunkara1:
@jasongrout The tooltip issue seems to be coming from `ipywidgets` `Button` doesn't display font-awesome icons. Do you know why that could be happening?

PR #353, opened 2016-11-01
Minor bug fix and Styling changes

ssunkara1:
Fixed the issue where width and height are being set to -120.
Changed the default figure to be bigger. I felt the current default is too small.

@rmenegaux @SylvainCorlay @jasongrout

### DIFF
--- a/bqplot/figure.py
+++ b/bqplot/figure.py
@@ -18,10 +18,10 @@ r"""
 Figure
 ======
 
-.. currentmodule:: bqplot.figure
+.. currentmodule:: bqplot.figure::
 
-.. autosummary::
-   :toctree: _generate/
+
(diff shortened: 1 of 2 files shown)


bqplot/figure.py
```diff
@@ -125,7 +125,7 @@ class Figure(DOMWidget):
     layout = Instance(Layout, kw={
             'flex': '1',
             'align_self': 'stretch',
-            'max_width': '700px'
```
ssunkara1: I feel setting `min_width` rather than `max_width` by default as, globally, we are trying to expand whenever there is space available. Setting `max_width` prevents that.

jasongrout:
Well, let's merge as-is, and we can open a new PR to make it a bit smaller min width.

PR #356, opened 2016-11-02
Add staticlab files to wheel

SylvainCorlay:
@jasongrout


SylvainCorlay:
The released npm package with tag next will be 0.2.0.

PR #361, opened 2016-11-04
Switch to new js default value for ipywidgets

SylvainCorlay:


jasongrout:
Looks like a pretty straight-forward change to me. Thanks!

PR #362, opened 2016-11-16
Remove the `fig` kwarg in hline and vline

rmenegaux:
The `fig` kwarg in `hline` and `vline` was superfluous, since they call `plot` which already has a `figure` kwarg.

### DIFF
--- a/bqplot/pyplot.py
+++ b/bqplot/pyplot.py
@@ -393,24 +393,20 @@ def legend():
         m.display_legend = True
 
 
-def hline(level, fig=None, **kwargs):
+def hline(level, **kwargs):
     """Draws a horizontal line at the given level.
 
     Parameters
     ----------
     level: float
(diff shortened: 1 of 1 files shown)


ssunkara1:
@SylvainCorlay Do you know where I can find the log for the Travis build? I could find that the build is failing for python 2.7 but I cannot find what the specific error is.

jasongrout:
@ssunkara1 - click on the red x above, then click on the row that is failing. Here is the relevant part of the log:

```
300.29s$ nvm install stable
Version 'stable' not found - try `nvm ls-remote` to browse available versions.
The command "nvm install stable" failed and exited with 3 during .
Your build has been stopped.
```
Seems like there is an error in the travis config?

ssunkara1:
This looks good. Thanks @rmenegaux

PR #369, opened 2016-12-01
Add a data attribute to pyplot's plot functions

dmadeka:
@madridnick @ssunkara1 Added support for a data argument ala matplotlib.

Currently it works only for the args (not the kwargs) of the plotting functions and follows matplotlib in that it assumes whatever names you pass it is supported by the _get_item of your data.

```
plt.figure()
plt.plot('Security 1', 'Security 2', data=price_data)
plt.show()
```

### DIFF
--- a/bqplot/pyplot.py
+++ b/bqplot/pyplot.py
@@ -50,7 +50,10 @@ from numpy import arange, issubdtype, array, column_stack
 from .figure import Figure
 from .scales import Scale, LinearScale, Mercator
 from .axes import Axis
-from .marks import Lines, Scatter, Hist, Bars, OHLC, Pie, Map, Label
+from .marks import (
+        Lines, Scatter, Hist, Bars, OHLC, Pie, Map,
+        Label, topo_load
+    )
 from .toolbar import Toolbar
 from .interacts import (
         BrushIntervalSelector, FastIntervalSelector, BrushSelector,
@@ -209,6 +212,20 @@ def close(key):
     fig.close()
     del figure_registry[key]
     del _context['scale_registry'][key]
+
+
+def _process_data(func):
+    """Helper function to handle data keyword argument
+    """
+    def _mark_with_data(*args, **kwargs):
+        data = kwargs.pop('data', None)
+        if data is None:
+            return func(*args, **kwargs)
+        else:
+            data_args = [data[i] for i in args]
+            return func(*data_args, **kwargs)
+
+    return _mark_with_data
 
 
 def scales(key=None, scales={}):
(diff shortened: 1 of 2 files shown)


dmadeka:
@ssunkara1 I think this is good to merge, and I'd like to have it in the next release as a proof of concept. Can I merge?

ssunkara1:
This looks great! A few comments: 

1. I would expect to pass data attributes like `color` for  `Scatter`, `Bar` by passing  the column name in the data frame. 

2. For `plot` function, I think if only one argument is passed, the `x` should be the index of the data frame.

SylvainCorlay:
Remark: this closes #57 .

Have you looked at how matplotlib does this?

dmadeka:
@ssunkara1 Thats a difficult one. It would mean a separate data function for each attribute it seems. 

@SylvainCorlay Exactly like this I guess.

dmadeka:
https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/__init__.py#L1625

dmadeka:
okay @ssunkara1 @rmenegaux @SylvainCorlay this handles kwargs now as well. Can you check it out and make sure all is good? I tried the different pyplot notebooks and all works.

@SylvainCorlay I made a change to plt.geo as well, should be compatible with all previous syntax, but better now.

bqplot/pyplot.py
```diff
@@ -211,6 +222,23 @@ def close(key):
...
+    """Helper function to handle data keyword argument
+    """
+    def _data_decorator(func):
+        def _mark_with_data(*args, **kwargs):
+            data = kwargs.pop('data', None)
+            if data is None:
+                return func(*args, **kwargs)
+            else:
+                data_args = [data[i] if isinstance(i, string_types) else i for i in args]
+                data_kwargs = {kw: data[kwargs[kw]] if isinstance(kwargs[kw], string_types) else kwargs[kw] for kw in list(set(kwarg_names).intersection(list(kwargs.keys())))}
```
dmadeka: @SylvainCorlay

bqplot/pyplot.py
```diff
@@ -91,6 +94,14 @@
 MARKER_CODES = {'o': 'circle', 'v': 'triangle-down', '^': 'triangle-up',
                 's': 'square', 'd': 'diamond', '+': 'cross'}
 
+import sys
+PY3 = sys.version_info[0] == 3
+
+if PY3:
+    string_types = str,
+else:
+    string_types = basestring,
```
SylvainCorlay: it is better to check for PY2 (otherwise it will fail for python 4 :) )

dmadeka: :+1:

bqplot/pyplot.py
SylvainCorlay: wow

PR #370, opened 2016-12-02
Tooltip figure bug fix

ssunkara1:
If we trigger `displayed` on the `view` before the `widget` in the `tooltip` is displayed, the `width` and `height` are set to negative values as the `el.clientHeight` and `el.clientWidth` are 0. 

This PR fixes it by not triggering `displayed` when `tooltip` is not necessarily displayed.

### DIFF
--- a/examples/Interactions/Mark Interactions.ipynb
+++ b/examples/Interactions/Mark Interactions.ipynb
@@ -11,7 +11,8 @@
     "from __future__ import print_function\n",
     "from bqplot import *\n",
     "import numpy as np\n",
-    "import pandas as pd"
+    "import pandas as pd\n",
+    "from ipywidgets import Layout"
    ]
   },
   {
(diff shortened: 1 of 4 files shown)


ssunkara1:
@jasongrout @SylvainCorlay

jasongrout:
@ssunkara1 and I talked about this. We decided that:

1. Widgets do a confusing thing - the `displayed` trigger confuses the notions of attaching to the DOM and displaying on the page.
2. In this case, we *do* want to notify the view that we've attached to the page, but not that we are displaying it.
3. @ssunkara1 is going to kludge a solution - pass some extra arguments to the displayed trigger, which will then be available from the displayed promise. Then in bqplot itself, check the value of what the displayed promise resolves to, and decide to layout or not based on that.

SylvainCorlay:
Why not simply triggering displayed when the displayed promise of the tooltip resolves?

jasongrout:
This is about the mark triggering the tooltip's display event. At that time, the tooltip is attached to the DOM, but (usually) has `display: none`.

Another solution (which I've taken with the Figure widget to handle being inside of tabs) is to layout *both* when it is attached to the DOM *and* when it is made visible.

SylvainCorlay:
> This is about the mark triggering the tooltip's display event. At that time, the tooltip is attached to the DOM, but (usually) has display: none.

Gotcha

ssunkara1:
@jasongrout This is not ready yet. There is an issue which I am trying to resolve

jasongrout:
What issue are you trying to resolve?

ssunkara1:
@SylvainCorlay @jasongrout This is ready now. To handle the cases where `relayout` is not called, I am setting the `viewbox` of the svg in `render` so that the enclosing div does not shrink to something smaller.

jasongrout:
Looking at it now...

jasongrout:
There are some subtleties with this after #371, but I'll take care of those in that PR.

PR #371, opened 2016-12-02
Update to latest widgets

jasongrout:
Also, make the top-level element a div, and relayout on the after-show phosphor message.

### DIFF
--- a/bqplot/market_map.py
+++ b/bqplot/market_map.py
@@ -27,9 +27,10 @@ Market Map
    SquareMarketMap
 """
 
-from traitlets import Int, Unicode, List, Dict, Enum, Bool, Instance
+from traitlets import Int, Unicode, List, Dict, Enum, Bool, Instance, Float
 from traittypes import Array, DataFrame
-from ipywidgets import DOMWidget, CallbackDispatcher, Color, widget_serialization
+from ipywidgets import (DOMWidget, CallbackDispatcher, Color,
+    widget_serialization, Layout)
 
 from .traits import array_serialization, dataframe_serialization, dataframe_warn_indexname
 from .marks import CATEGORY10
@@ -82,17 +83,6 @@ class MarketMap(DOMWidget):
         True, the finer elements are blurred
 
     Map Drawing Attributes
-
-    map_width: int
-        minimum width of the entire map
-    map_height: int
-        minimum height of the entire map
-    map_margin: dict
-        margin for the market map plot area with respect to the entire display
-        area
-    preserve_aspect: bool
-        boolean to control if the aspect ratio should be preserved or not
-        during a resize
     cols: int
         Suggestion for no of columns in the map.If not specified, value is
         inferred from the no of rows and no of cells
@@ -107,6 +97,18 @@ class MarketMap(DOMWidget):
     row_groups: int
         No of groups the rows should be divided into. This can be used to draw
         more square cells for each of the groups
+
+    Layout Attributes
+
+    map_margin: dict (default: {top=50, bottom=50, left=50, right=50})
+        Dictionary containing the top, bottom, left and right margins. The user
+        is responsible for making sure that the width and height are greater
+        than the sum of the margins.
+    min_aspect_ratio: float
+         minimum width / height ratio of the figure
+    max_aspect_ratio: float
+         maximum width / height ratio of the figure
+
 
     Display Attributes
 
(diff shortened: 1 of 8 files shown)


jasongrout:
(You can check the after-show message works by creating a tab widget with a bqplot in one of the tabs that doesn't show at the start)

Do you think we should relayout the bqplot only on the *first* show message?

SylvainCorlay:
I don't think so. The browser window may be resized while the tab is invisible...

jasongrout:
Yeah, I guess there are no shortcuts.

jasongrout:
I realized I broke the market map because it assumed `this.el` was the svg. Fixed.

jasongrout:
Rebased on master

jasongrout:
I went through the examples, and affected things seem to work well, except for saving a bqplot image. It just saves a black rectangle as far as I can see. I think once that is fixed, this will be ready to go in.

dmadeka:
@jasongrout Damn it, you beat me to my PR. I thought `white` was a bit too opaque, maybe we can try something like `#fcfcfc` for the light?

jasongrout:
Since the tooltip is often partly transparent, the white didn't look bad to me. But I don't see much difference either way, so I made your change and pushed the new commit.

dmadeka:
btw @jasongrout Does the css for the `Toolbar` work for you on master?

jasongrout:
@ssunkara1 - I just pushed a change that helps a figure work inside an HBox. However, something is still weird. It looks fine in Firefox, but in Chrome, the svg is only about 600px wide, but is blown up to around 1000px wide. It appears that the top-level figure div has the following properties on the element style:
```
align-self: stretch;
flex: 1 1 0%;
min-width: 400px
```
but I can't find anywhere in the code that is being set. Do you know where it may be set? I refreshed without cache, tried incognito, and even firefox.

jasongrout:
Can you make sure to reload (or try incognito), and then upload a screenshot?

jasongrout:
> btw @jasongrout Does the css for the Toolbar work for you on master?

Oh, sorry, I thought you meant on this PR. I'm not checking master - hopefully this PR will be master Very Soon.

ssunkara1:
@jasongrout My question about `HBox` was more why are `HBox` and `VBox` not consistent with each other with regards to what the `el.clientWidth` and `el.clientHeight` are? Shouldn't that be the case?

jasongrout:
> @jasongrout My question about HBox was more why are HBox and VBox not consistent with each other with regards to what the el.clientWidth and el.clientHeight are? Shouldn't that be the case?

Not necessarily, especially with that weird align-self stretch thing I was seeing. Do you have any idea where that is coming from in this PR? Do you see the figure div having those element styles applied?

ssunkara1:
Right now the `max_width`, `max_height` of the `Layout` passed are not respected since we are setting `min_width` for the `svg`. Not sure what is the best way to handle this.

```python
import numpy as np
from bqplot import *
from ipywidgets import Layout

x_ord = OrdinalScale()
y_sc = LinearScale()

bar = Bars(x=np.arange(10), y=np.random.randn(10), scales={'x': x_ord, 'y': y_sc})
Figure(marks=[bar], layout=Layout(max_width='300px', min_width='200px'))

ssunkara1:
> Not necessarily, especially with that weird align-self stretch thing I was seeing. Do you have any idea where that is coming from in this PR? Do you see the figure div having those element styles applied?

@jasongrout The issue are not directly related to this PR. So I think we can deal with it separately. 

But I am not totally comfortable with the same `Figure` looking different if I put it inside a `VBox` as opposed to a `HBox`.

jasongrout:
> Right now the max_width, max_height of the Layout passed are not respected since we are setting min_width for the svg. Not sure what is the best way to handle this.

Perhaps the relayout method is not being called after the Layout properties are applied to the element?

jasongrout:
> Perhaps the relayout method is not being called after the Layout properties are applied to the element?

We should handle this at the widget level - applying Layout properties should trigger a phosphor fit/resize.

jasongrout:
> We should handle this at the widget level - applying Layout properties should trigger a phosphor fit/resize.

ipython/ipywidgets#949 solves this issue.

ssunkara1:
@jasongrout ipython/ipywidgets#949 does solve the `max_width` issue. Thanks.
With that PR, I see that we get a JS error of `set_range` being not defined and the `MarketMap` not displaying correctly. I will try to fix those issues.

ssunkara1:
@jasongrout Made the changes I suggested above in https://github.com/jasongrout/bqplot/pull/1

Have a look. Thanks!

SylvainCorlay:
@jasongrout I made an independent version bump in package.json to get bqplot to compile with jupyterlab 0.11.3.

PR #373, opened 2016-12-05
Fix data attribute

dmadeka:
@SylvainCorlay @jasongrout Does the update work for all versions of Python?

I dont want to do {**x, **y} to merge the two dictionaries.

### DIFF
--- a/bqplot/pyplot.py
+++ b/bqplot/pyplot.py
@@ -243,7 +243,7 @@ def _process_data(*kwarg_names):
             else:
                 data_args = [data[i] if hashable(data, i) else i for i in args]
                 data_kwargs = {
-                   kw: data[kwargs[kw]] if hashable(data, kwargs[kw]) else kwargs[kw] for kw in list(set(kwarg_names).intersection(list(kwargs.keys())))
+                   kw: data[kwargs[kw]] if hashable(data, kwargs[kw]) else kwargs[kw] for kw in set(kwarg_names).intersection(list(kwargs.keys()))
                 }
                 kwargs_update = kwargs.copy()
                 kwargs_update.update(data_kwargs)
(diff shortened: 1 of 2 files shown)


bqplot/pyplot.py
```diff
@@ -231,9 +237,13 @@ def _mark_with_data(*args, **kwargs):
...
                 return func(*args, **kwargs)
             else:
-                data_args = [data[i] if isinstance(i, string_types) else i for i in args]
-                data_kwargs = {kw: data[kwargs[kw]] if isinstance(kwargs[kw], string_types) else kwargs[kw] for kw in list(set(kwarg_names).intersection(list(kwargs.keys())))}
-                return func(*data_args, **data_kwargs)
+                data_args = [data[i] if hashable(i) else i for i in args]
+                data_kwargs = {
+                   kw: data[kwargs[kw]] if hashable(kwargs[kw]) else kwargs[kw] for kw in list(set(kwarg_names).intersection(list(kwargs.keys())))
+                }
+                kwargs_update = kwargs.copy()
```
dmadeka: @SylvainCorlay Is it better to do `dict(ChainMap(data_kwargs, kwargs))` here?

bqplot/pyplot.py
```diff
@@ -94,13 +95,18 @@
...
+
+if PY2:
     string_types = basestring,
+else:
+    string_types = str,
+
+
+def hashable(v):
+    """Determine whether `v` can be hashed."""
+    return hasattr(v, '__hash__')
```
ssunkara1:
```python
def hashable(dic, v):
    """Determine whether `v` can be hashed."""
    try:
        dic[v]
    except (TypeError, KeyError, IndexError):
        return False
    return True
```
This works better for the hashable

PR #383, opened 2016-12-06
Figure as a tooltip for MarketMap doesn't display

ssunkara1:
Triggering `displayed` so that the size of the `Figure` is set.

### DIFF
--- a/js/src/MarketMap.js
+++ b/js/src/MarketMap.js
@@ -655,6 +655,7 @@ var MarketMap = figure.Figure.extend({
             tooltip_widget_creation_promise.then(function(view) {
                 that.tooltip_view = view;
                 that.tooltip_div.node().appendChild(view.el);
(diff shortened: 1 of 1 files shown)


jasongrout:
The market map example notebook last cell works fine for me with #371. Can you check on that PR? If it works for you, I'm curious why it works without this line.

ssunkara1:
I think it is because of lines 107 and 108 in `Figure.js` where the min-width and min-height of the svg are being added. Yesterday after a discussion with @SylvainCorlay I removed them in #370

jasongrout:
No, the min-width and min-height of the *svg* are there now so that the enclosing Figure div has a natural size. Otherwise we were getting the enclosing div to be zero-sized.

jasongrout:
Which I guess is about what you said.

ssunkara1:
@jasongrout This is good to merge :)

jasongrout:
Great, thanks!

PR #384, opened 2016-12-07
Add Viridis for Categorical Scales

dmadeka:
@SylvainCorlay I wanted this PR to be the point where we start discussing how bad our color scales are, and define a roadmap to improve them. 

My suggestions are:
- [ ] Currently, I dont know why, @ssunkara1 , for a linear scale it interpolates linearly between the min, mid, high, rather than interpolating between colors. What was the logic for the 3 colors?
- [ ] We refactor the logic that depends on an arbitrary index to be dictionary like
- [ ] We rely on d3, only interpolating if the user passes colors, that too we should accept n colors

### DIFF
--- a/js/src/ColorUtils.js
+++ b/js/src/ColorUtils.js
@@ -19,7 +19,7 @@ var colorbrewer = require("./colorbrewer");
 var utils = require("./utils");
 
 var color_schemes = [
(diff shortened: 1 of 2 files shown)


dmadeka:
Another option @ssunkara1 @SylvainCorlay is that we import d3-scale-chromatic, one of the smaller sub repos with functions to interpolate viridis, magma, etc. What do you guys think?

jasongrout:
Does d3-scale-chromatic interpolate viridis? It doesn't appear in the readme (from a quick search)

jasongrout:
It does look like https://github.com/d3/d3-scale interpolates the viridis scale.

PR #388, opened 2016-12-15
Market map refdata fix

ssunkara1:
Right now, `MarketMap` does not display if the argument `ref_data` is not passed because of a JS error. This fixes that bug.


SylvainCorlay:
LGTM.

PR #389, opened 2016-12-15
Use the bqplot sphinx extension for the usage example

SylvainCorlay:
This requires the current master branch of ipywidgets (or the ipywidgets 6.0.0b6 when it is out).

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -6,7 +6,9 @@ dependencies:
  - traitlets
  - traittypes
  - widgetsnbextension
- - ipywidgets
+ - ipywidgets>=6.0.0b5
  - numpy
  - pandas
  - bqplot
+ - pip:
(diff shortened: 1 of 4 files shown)


SylvainCorlay:
The second chart does not show the marks

`"Cannot read property 'range' of undefined"` on the ordinal scale.

In `scale.js`: scale is undefined. It seems that `render` was never called on the child scale.

```js
set_range: function(range, padding) {
    this.scale.range(range); 
},
```

SylvainCorlay:
I am going to need to merge that for RTD rendering testing. We can always revert if there is an issue before the next release.

SylvainCorlay:
OK this works http://bqplot.readthedocs.io/en/latest/usage.html I am going to improve the sphinx extension so that we can use `show`

PR #412, opened 2017-01-20
Default size fixes

ssunkara1:
* Firstly, the `min_width` attribute is not as innocuous as it seems. Because `min_width` along with `aspect_ratio` is determining the `min_height` which is setting the el.clientHeight since there is no default natural height.

* I do not like setting the properties on the `svg` and on the `div` separately but the `Layout` seems to clear styling that I set even if that styling is not passed to the `Layout`. For e.g., if `min_height` is not set on the `Layout` object, and I set the style manually on the javascript side, it is still cleared by the `Layout`. So, I have to set the style for the svg.

### DIFF
--- a/js/src/Figure.js
+++ b/js/src/Figure.js
@@ -47,7 +47,7 @@ var Figure = widgets.DOMWidgetView.extend({
         } else if (height_undefined) {
             suggested_height = suggested_width / min_ratio;
         } else if (width_undefined) {
-            suggested_width = suggested_height * min_ratio;
+            suggested_width = suggested_height * max_ratio;
         }
 
         var ratio = suggested_width / suggested_height;
@@ -80,6 +80,9 @@ var Figure = widgets.DOMWidgetView.extend({
         this.width = impl_dimensions["width"];
         this.height = impl_dimensions["height"];
 
+        this.svg.style("min-width", this.width + "px");
+        this.svg.style("min-height", this.height + "px");
+
         this.id = widgets.uuid();
 
         // Dictionary which contains the mapping for each of the marks id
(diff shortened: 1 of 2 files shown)


ssunkara1:
@jasongrout @SylvainCorlay @dmadeka

js/src/Figure.js
```diff
@@ -80,6 +80,9 @@ var Figure = widgets.DOMWidgetView.extend({
         this.width = impl_dimensions["width"];
         this.height = impl_dimensions["height"];
 
+        this.svg.style("min-width", this.width + "px");
+        this.svg.style("min-height", this.height + "px");
```
jasongrout: This seems to be a return to making it so that the svg always grows? What about putting a min-width and min-height for the div in the .less file? Then its overridden by the Layout if desired, but is there if not given in a Layout.

SylvainCorlay: :+1:

ssunkara1: This will not work for the reason that I mentioned above. The `min-height` mentioned in the css will be the binding constraint and even if I increase the `min-width`, the `min-height` will determine the `el.clientHeight`.  And I cannot set the `min-height` programmatically because the `Layout` clears it out.

ssunkara1: Don't know if this is an adequate check but this seems to work fine.

```python
x_ord = OrdinalScale()
y_sc = LinearScale()

bar = Bars(x=x_data[:20], y=y_data[:20], scales={'x': x_ord, 'y': y_sc})
ax_x = Axis(scale=x_ord)
ax_y = Axis(scale=y_sc, orientation='vertical', tick_format='0.2f')

f = Figure(marks=[bar], axes=[ax_x, ax_y])
f2 = Figure(marks=[bar], axes=[ax_x, ax_y])

h = HBox([f, f2])
h

## next cell
h.layout = Layout(width='1600px', height='900px')

ssunkara1: @jasongrout @SylvainCorlay what do you guys think about this?

jasongrout: Thanks. Let's sit down to check this tomorrow.

jasongrout:
@ssunkara1, @dmadeka, and I decided to go ahead and try Sylvain's proposal at #421 and see how well it worked in practice. #412 is still a possibility if it seems better than #421 after we've used it for a while.

PR #424, opened 2017-01-27
Increase max aspect ratio

jasongrout:
We might often have multiple bqplot charts stacked on a 16:9 monitor. This increases the max aspect ratio to fill the space on a 16:9 monitor when we have about 3 charts stacked on top of each other.


jasongrout:
CC @ssunkara1.

PR #429, opened 2017-02-03
Binning in the backend for the Hist mark

rmenegaux:
This will enable plotting histograms without shipping large sets of data back and forth to the frontend.

The binning is done on the python side, using numpy's `histogram` function, the `Hist` mark inheriting from `Bars`.

The resulting `Hist.x` is the array of the bin centers, and `Hist.y` are the bin heights.


dmadeka:
@rmenegaux Is this still WIP?

rmenegaux:
This is ready to be looked at.

The major point of contention here is the name of the scales:
`Hist` is inheriting from `Bars`, so its positional scales are `x` and `y`. Hence the scale for `sample` data is actually `scales['x']`. 
I suggest we either keep it as is, or link a dummy `scales['sample']` to the `scales['x']`, although that might be a little clunky.

Another requirement that might feel counter-intuitive is that the user has to pass a `y` scale, even though he does not pass a `y` data attribute (this `y` is the height of the bars, that is computed internally by the binning algo).
Should we define a default `y` scale?

dmadeka:
@rmenegaux Is this ready?

rmenegaux:
@dmadeka Yes Sir, see previous comment

dmadeka:
@rmenegaux How will the selections etc work with this?

maartenbreddels:
I vote against this, or at least not removing this functionality. Since now you can have all the data on the front end, and do selections there (no backend needed, useful for embedding). If it is possible to do this yourself using Bar's, it is already possible.
I actually would be in favour of making sure the binning is fast, and we can do this for say ~1 million data points.

ssunkara1:
How about we keep the old `Hist` and add this as a separate mark which will be useful for plotting when we have a lot of data? This comes with the caveat that at least initially we do not support selections on the new mark.

maartenbreddels:
Why is the selections involved btw? I'd expect to see either
 * 1 histogram, where only the selected values are binned
 * 2 histograms, the above, overplot on a histogram showing all values.

The 2nd can be emulated by having two histograms, one showing all data, and another showing only a histogram when there is a selection.

ssunkara1:
@maartenbreddels Ahh ok I see what you mean.  
Visually, when the `selected` attribute is set from the python side, changing the display properties of the entire bar even though only a few points falling in that interval are selected is misleading. When we click to select bars in the `Hist` directly in the browser, like using a `BrushSelector` or by `clicking` on the bars, I think changing the style of the entire bar makes sense.

From your two proposals, I like proposal 2 more. I think we should add that at least as an option of viewing the `selected` points in `Hist`.

maartenbreddels:
Great, that is I think what most people would expect. I was linking two Histograms selected attributes, and selecting in one histogram seeing how it looked in the other, while keeping an overview of all the data.

PR #430, opened 2017-02-07
DOM append -> appendChild

jasongrout:
append is only implemented in some of the latest browsers, so causes errors in, for example, Chrome before version 54.


jasongrout:
CC @ssunkara1

SylvainCorlay:
This is a leftover from d3 selections.

ssunkara1:
@SylvainCorlay This is for compatibility with chrome <= 0.54. The newer version of chrome has `.append` on DOM Elements.

jasongrout:
ah, so d3 had the append method, and it was just our luck that Chrome >= 54 also had an append method that functioned the same.

SylvainCorlay:
> ah, so d3 had the append method, and it was just our luck that Chrome >= 54 also had an append method that functioned the same.

Yes, that is what I meant.

PR #434, opened 2017-02-09
Tooltip positioning

ssunkara1:
This is for fixing the `tooltip` positioning. Now the `tooltip` is added directly to the top-level `div` of the `Figure`. 

There is one consequence of this.  Since the top-level div has `overflow : hidden`, the `tooltip` is not displayed if it goes outside the `div`. I had a conversation with @jasongrout regarding this and the consensus was that removing `overflow: hidden` will cause the `div` to measure itself according to its contents which we do not want. But that does not seem to be true at least in the case below.

image_of_hope

Here the `div` still seems to be its actual size even though the child is much larger and is overflowing. In a few resizing tests that I ran, it seems to be working fine. Having said that, the tests are not exhaustive. So, this should be tested thoroughly before being considered for a merge. 

P.S.: I tried adding the `tooltip` to the parent of the `div`. But that causes problems because the `tooltip` is `position: absolute`. Hence the ancestor which is relative determines the location and therefore it is not going to be consistent when it is displayed directly and when it is inside a `HBox`.;


ssunkara1:
ping @jasongrout @SylvainCorlay

jasongrout:
> Here the div still seems to be its actual size even though the child is much larger and is overflowing. In a few resizing tests that I ran, it seems to be working fine. Having said that, the tests are not exhaustive. So, this should be tested thoroughly before being considered for a merge.

We might have been mistaken in making the overflow: hidden. I think the relevant part of the spec is https://www.w3.org/TR/CSS2/visudet.html#Computing_widths_and_margins, which specifies when the content determines the size of the containing box.

SylvainCorlay:
I think that @ssunkara1 's approach is the right one.

jasongrout:
I'm trying to get a minimal example showing that a div's size is not affected by child svg:

http://codepen.io/anon/pen/MJPBNX

This indicates that at least the height is affected by the child svg.

SylvainCorlay:
Is it not affected when positioning is absolute?

jasongrout:
> Is it not affected when positioning is absolute?

If the svg is position absolute, then the div is not affected.

SylvainCorlay:
ok. The tooltip is position absolute

jasongrout:
Yes, the tooltip is absolute position, so it doesn't affect the sizing of the enclosing div. That's not the question.

The question here is about making the containing div's size not be affected by the svg's size, in a way that does not require overflow: hidden on the containing div (which would clip the tooltip). I *think* @ssunkara1 is right that the overflow setting does not change this. I'm trying to understand exactly what makes the enclosing div be sized independent of the svg size.

jasongrout:
It also looks like the flexbox model is completely different in its calculation of sizes, etc: https://www.w3.org/TR/css-flexbox-1/#intro

I still don't understand this fully, but I think you're right that the overflow: hidden doesn't actually affect things here. The code looks good to me, so merging.

Issue #398, opened 2017-01-10
`dtype` for Scatter mark set to `object` when moving points


SylvainCorlay:

ssunkara1:
This has been fixed.

PR #435, opened 2017-02-10
Data type issue fix

ssunkara1:


ssunkara1:
fixes #398

ssunkara1:
I am still converting an array of `dtype` `int` to `dtype` `float` as preserving type `int` leads to problems when the user drags a point to a non-int value.

PR #436, opened 2017-02-16
Adding rotation to scatter scales metadata

rmenegaux:
This was forgotten, and causing the pyplot PanZoom to fail when a rotation scale was specified.


SylvainCorlay:
:+1:

PR #457, opened 2017-03-13
Adding Heatmap to pyplot: plt.heat

rmenegaux:


bqplot/pyplot.py
```diff
@@ -854,6 +854,25 @@ def geo(map_data, **kwargs):
     return _draw_mark(Map, **kwargs)
 
 
+def heat(color, **kwargs):
```
dmadeka: heat_map? What does mpl do?

bqplot/marks.py
```diff
@@ -42,7 +42,7 @@
     )
 from traittypes import Array
 
-from .scales import Scale, OrdinalScale
+from .scales import Scale, OrdinalScale, LinearScale
```
dmadeka: Is Ordinal still used?

dmadeka:
@rmenegaux  Id like a syntax like https://discuss.analyticsvidhya.com/t/how-to-plot-heat-map-in-python/1922/2

``` python
plt.pcolor or plt.map(data, cmap='Reds')
```

@ssunkara1 what do you think?

ssunkara1:
@dmadeka @rmenegaux How about `plt.heatmap` ? I suggest that because that seems the most intuitive one to me.

bqplot/marks.py
rmenegaux: @dmadeka Yes

rmenegaux:
Okay going with heatmap then

PR #459, opened 2017-03-14
Fixing transitions

### LINKED ISSUE #456: Animations need fixing
https://bost.ocks.org/mike/transition/#per-element

According to the above link, there cannot be two transitions on the same selection. So once a new transition is started on a selection which already has a transition going-on, all previous transitions are cancelled. 

```python
# In[1]:
import bqplot
import numpy as np
import bqplot.pyplot as plt

scat_x = np.random.randn(5)
scat_y = np.random.randn(5)
scat_color = np.random.randn(5)

def_scales = {'color': bqplot.ColorScale()}

plt.figure()
scat = plt.scatter(scat_x, scat_y, color=scat_color, scales=def_scales)
plt.show()

# In[2]:
scat.x = np.random.randn(30)
scat.y = np.random.randn(30)
scat.color = np.random.randn(30)
def_scales['color'].min=-5

```

This is the true cause of the bug which was causing #450. 

The way to deal with it could be to attach an id to every transition on the same selection. I will look into this.

ssunkara1:
Added a name to each transition. The convention I followed was to add the name of the  function which contains the transition. In this way, transitions started by other functions are not interrupted. Luckily, we do not have more than 1 transition on the same selection in the same function :)

Fixes #456.

### DIFF
--- a/js/src/Axis.js
+++ b/js/src/Axis.js
@@ -160,7 +160,8 @@ var Axis = widgets.WidgetView.extend({
 
         if(this.g_axisline) {
              this.g_axisline
-                .transition().duration(animate === true ? this.parent.model.get("animation_duration") : 0)
+                .transition("set_tick_values")
+                .duration(animate === true ? this.parent.model.get("animation_duration") : 0)
                 .call(this.axis);
         }
     },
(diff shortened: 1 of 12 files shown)


rmenegaux:
Good convention.
If we have several transitions on the same selection in the same function we can use `transition.transition` so it will not be a problem.

PR #463, opened 2017-03-20
remove marks and axis

maartenbreddels:
I noticed the Mark's `remove` method was not called, this fixes that, and I think the same for axis. As discussed with @jasongrout

### DIFF
--- a/js/src/Figure.js
+++ b/js/src/Figure.js
@@ -635,6 +635,12 @@ var Figure = widgets.DOMWidgetView.extend({
         this.title.text(this.model.get("title"));
     },
 
+    remove: function() {
+        Figure.__super__.remove.apply(this, arguments);
+        this.mark_views.remove()
+        this.axis_views.remove()
+    },
+
     save_png: function() {
 
         var  replaceAll = function (find, replace, str) {
--- a/js/src/Figure.js
+++ b/js/src/Figure.js
@@ -636,9 +636,9 @@ var Figure = widgets.DOMWidgetView.extend({
     },
 
     remove: function() {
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
Should we also call remove on the interaction and the scales?

maartenbreddels:
there is also `interaction_view`, I'm not sure I know bqplot good enough to cover all of these cases.

js/src/Figure.js
```diff
@@ -635,6 +635,12 @@ var Figure = widgets.DOMWidgetView.extend({
         this.title.text(this.model.get("title"));
     },
 
+    remove: function() {
+        Figure.__super__.remove.apply(this, arguments);
+        this.mark_views.remove()
+        this.axis_views.remove()
```
ssunkara1: Should they be called in the reverse order? Generally for a `destructor`, we `destory` the `child` first and then the `parent`, no?

ssunkara1:
I thought we had this before. Don't know where it went! Thanks :)

maartenbreddels:
@ssunkara1 will you handle any other views that need deleting?

jasongrout:
> @ssunkara1 will you handle any other views that need deleting?

+1. If this looks good, perhaps we should merge and iterate.

js/src/Figure.js
maartenbreddels: Yes, I'll do that and see if it works.

ssunkara1:
@maartenbreddels Thanks. This looks good. Merging.

maartenbreddels:
Great :+1:

SylvainCorlay:
Thanks @maartenbreddels

PR #467, opened 2017-04-04
Pie enhancements

ChakriCherukuri:
1. Enhanced transitions of Pie to take into account missing data
2. Made display_labels enum with 'inside', 'outside' and 'none' options
3. Added display_values option (and values_format) to display values
4. Enhanced example noteboks to reflect the above changes


rmenegaux:
Thanks

PR #468, opened 2017-04-07
Added Neural Net Dashboard to examples

dmadeka:
@ssunkara1 @rmenegaux Added a small notebook.

Question: Should we move to Keras 2.0? This notebook is running on our last build: Keras 1.2.2

@davidrosenberg - Would love comments on what we could add to make this better :)


dmadeka:
@ssunkara1 What do you think about extending this to include the Graph and the Neural Net visualization?

ssunkara1:
Yeah I have that from the plotcon demo. I will raise a PR for that.

PR #480, opened 2017-05-31
Added ability to change legend_style and legend_text

madridnick:
@dmadeka, @ssunkara1 
First stab at customizing legend and legend text

### DIFF
--- a/js/src/Figure.js
+++ b/js/src/Figure.js
@@ -536,7 +536,7 @@ var Figure = widgets.DOMWidgetView.extend({
         var legend_disp = 30 + num_series * 7;
         var legend_height = 14;
         var legend_width = 24;
-        var legend_location = this.model.get("legend_location")
+        var legend_location = this.model.get("legend_location");
 
         var legend_g = this.fig_marks.append("g")
           .attr("class", "g_legend");
@@ -558,7 +558,7 @@ var Figure = widgets.DOMWidgetView.extend({
 
                 var coords = that.get_legend_coords(legend_location, legend_width, (count + 1) * (legend_height + 2), 0);
                 if(count !== 1) {
-                    legend_g.append("g")
+                    legend_g.insert("g", ":first-child")
                       .attr("class", "axis")
                     .append("rect")
                       .attr({"y": (legend_height + 2) / 2.0,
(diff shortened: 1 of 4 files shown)


ssunkara1:
Thanks @madridnick 

I notice a small issue. When I set `legend_style`, it applies the style to the rectangle added on top of the legend entries. Because of this, when we set the `fill` property of `legend_style` , the legend is no longer visible.

image 1

To handle this, I think you can add the rectangle initially, before the marks add their legend entries, and set the properties of this rectangle later.

js/src/Figure.js
```diff
@@ -552,15 +564,21 @@ var Figure = widgets.DOMWidgetView.extend({
...
+                      
                 }
                 max_label_len = (legend_location === "top-right" ||
                                  legend_location === "right" ||
                                  legend_location === "bottom-right") ? -(max_label_len + 2) : 1;
                 var em = 16;
                 legend_g.attr("transform", "translate(" + String(coords[0] + max_label_len * em) + " " +
                                                           String(coords[1]) + ") ");
+
+                var legend_rect = legend_g.selectAll("rect");
```
dmadeka: @madridnick Since you don't reuse this variable, you can chain the two statements without an assignment

js/src/Figure.js
```diff
@@ -552,15 +564,21 @@ var Figure = widgets.DOMWidgetView.extend({
...
+                legend_rect.style(that.model.get("legend_style"));
+
+                var legend_text = legend_g.selectAll("text.legendtext");
```
dmadeka: same here

js/src/Figure.js
```diff
@@ -524,7 +536,7 @@ var Figure = widgets.DOMWidgetView.extend({
         var legend_disp = 30 + num_series * 7;
         var legend_height = 14;
         var legend_width = 24;
-        var legend_location = this.model.get("legend_location");
+        var legend_location = this.model.get("legend_location")
```
dmadeka: Not sure why the semi-colon was deleted

dmadeka:
@madridnick This looks great. Just needs one or two more commits, do you want me or @ssunkara1 to push to your branch?

madridnick:
@ssunkara1, @dmadeka 
Fixed the fill property issue. Big thanks to to @cquah for her JS expertise.

js/src/Figure.js
```diff
@@ -237,6 +239,8 @@ var Figure = widgets.DOMWidgetView.extend({
         this.listenTo(this.model, "change:title_style", this.title_style_updated, this);
         this.listenTo(this.model, "change:background_style", this.background_style_updated, this);
         this.listenTo(this.model, "change:layout", this.change_layout, this);
+        this.listenTo(this.model, "change:legend_style", this.legend_style_updated, this);
+        this.listenTo(this.model, "change:legend_text", this.legend_text_updated, this);
```
ssunkara1: The same set of listeners is added twice. You can remove either of them. If you do remove the ones in line 199-200, you need to make `legend_g` an object level variable.

ssunkara1:
Thanks @madridnick There is one minor fix that I just suggested. After that I think this is good to merge.

PR #492, opened 2017-06-28
Fix bqplot.Figure documentation typo

TotalVerb:
At the request of @ssunkara1, this addresses an omission in the fix in #488.


ssunkara1:
Thanks @TotalVerb

PR #502, opened 2017-07-06
Fix bug in Wealth of Nations example

### LINKED ISSUE #493: wealth of nations hover behavior is incorrect
In the wealth of nations example, the trajectory that is highlighted upon hovering over a data point is not correct.

For example in 2008 if I hover over Equatorial Guinea the trajectory for Eritrea is highlighted. Similarly if I hover over Eritrea, the trajectory for Ethiopia is highlighted.

Equatorial Guinea => Eritrea => Ethiopia => Gabon => Ghana => Guinea => Guinea-Bissau => ...

It looks like perhaps something is offset.

dmadeka:
Fixes #493


ssunkara1:
Thanks @dmadeka

PR #503, opened 2017-07-06
Add xlabel and ylabel

dmadeka:
@ssunkara1 @rmenegaux A quick review would be nice, not too sure about this


ssunkara1:
@dmadeka Looks good. Can we not return anything in these functions? In that way, we need not catch the output when we use it in the notebook.

dmadeka:
@ssunkara1 Actually, I did it deliberately - `plt.axes` returns the axes objects, but if you want one - youll have to access the key, here it gives you the one at the key 'x' if you call xlabel directly

matplotlib returns the associated `Text` object, but we don't have that concept - so I picked the corresponding axes

ssunkara1:
I think it is kind of inconsistent with the other functions in pyplot. Like, when we set the title on the figure, or the grids on the axis we do not return it.  any opinion @rmenegaux @SylvainCorlay @ChakriCherukuri

dmadeka:
@ssunkara1 We probably should if the goal is to be consistent with mpl right?

rmenegaux:
I think they should not return anything, as we do not have an axislabel widget.

- `plt.figure()` returns a figure, `.plot()` a mark, `.axes()` an axis, which is expected.
- `plt.grids()`, `.title()`, `.legend()` don't return anything as we do not have corresponding objects

In general the methods that create widgets return the created widgets, but those that modify properties inplace do not return anything.

dmadeka:
@rmenegaux I think they should, if suppressing an output is the only reason we're doing this. matplotlib returns plenty of junk whenever you plot, and its only the last line that creates the issue. Why I think we should is because, people might not want to have to call the axes object to get the axis, maybe they want to just change the xlabel and then keep changing the xlabel or some other attribute, mpl has separate objects which allows that easily, in our case they would keep calling `plt.xlabel` which is a different workflow.

My point is, is suppressing the output the only reason not to return something?

dmadeka:
@rmenegaux @ssunkara1 Removed the return for now. I suggest we open an issue and continue the discussion

ssunkara1:
cool. Sounds good. Thanks!

Issue #508, opened 2017-07-17
Bars not updating properly on data change


dustinmichels:
I have a custom class that extends bars, `class GenDispatchBars(Bars).` It generates a stacked bar plot using data about various "generators" as the y data (and hours 1-24 as the x data).

![image](https://user-images.githubusercontent.com/16504363/28278797-9d34d06a-6adb-11e7-9a64-0de3f445889e.png)

I have a function that primarily updates the object's `self.y` to reflect only a particular generator. When the data is changed, the plot updates automatically but it seems to leave behind some residue of the previous bars. Like this:

![image](https://user-images.githubusercontent.com/16504363/28278905-f1d08466-6adb-11e7-879b-a3690f58a645.png)

or this:

![image](https://user-images.githubusercontent.com/16504363/28278939-1916e808-6adc-11e7-809b-1215d470ddfc.png)

Is there a more idiomatic way to update data than directly changing the attribute? Or is this possibly a bug?

A note: If I run the `fig` cell again, the residue disappears.

![image](https://user-images.githubusercontent.com/16504363/28279076-95c4b9e8-6adc-11e7-80bc-6b7e49ad23f3.png)

Thank you!!

kdheepak:
MWE 

```python
import sys
if sys.version_info[0] < 3: 
    from StringIO import StringIO
else:
    from io import StringIO

import pandas as pd

TESTDATA=StringIO(''',GenCo0,GenCo1\n0,100,0\n1,100,0\n2,100,0\n3,120,0\n4,120,0\n5,120,0\n6,150,0\n7,150,0\n8,150,0\n9,200,0\n10,200,0\n11,200,0\n12,200,100\n13,200,200\n14,200,100\n15,200,0\n16,200,0\n17,200,0\n18,150,0\n19,150,0\n20,150,0\n21,150,0\n22,100,0\n23,100,0\n''')

df = pd.read_csv(TESTDATA, index_col=0)

from bqplot import OrdinalScale, LinearScale, Bars, Figure

m = Bars(x=df.index, y=df.T, scales={'x': OrdinalScale(), 'y': LinearScale()})

# Add it to a Figure to display
fig = Figure(marks=[m])
```

Changing `m.y` does not change the plot as one would expect.

kdheepak:
Not sure if this is the recommended way to do this, but I was able to get around this issue by doing the following.

```python
m.y = []
[...]

PR #509, opened 2017-07-17
Removing extra bars on redraw

rmenegaux:
This fixes issue #508 
Extra bars (when the number of rows in `y` decreases) were not being removed in a redraw


ssunkara1:
Thanks @rmenegaux

PR #513, opened 2017-07-21
Clear the examples.

jasongrout:
We can resave them before release.


dmadeka:
@jasongrout Why do you want to do this? This prevents us from sending nbviewer links to people

jasongrout:
I want to do it because we should be refreshing this content on each release anyway, but I don't want to refresh it yet because we haven't released :).

nbviewer links should go to the stable pages, right?

jasongrout:
It would also be cool if, as part of the update to ipywidgets 7, we could store the arrays in binary, which would save their state as base64-encoded text. That would significantly reduce the size of the examples.

dmadeka:
@jasongrout Im not sure this should have been merged. The discussion was still going on.

And ideally, yes, nbviewer links go to stable pages. But that's not what happens in practice, I've mostly seen people send the one easiest to access - which is usually the master one. I dont see why this should be done in that regard. The notebooks re usually stable

jasongrout:
Sorry if I pulled the trigger too soon. We do have to update - the representation inside the notebook changed, and the saved state representation also changed. There might be more changes in bqplot (like for example the binary array stuff above), so I was hesitant to commit the new data and have a lot of churn on the master branch.

Do you think we should commit the current widget state right now, even if it leads to churn while we develop on master? I think we should just agree that master is master and things are in flux, and commit the widget state perhaps when we hit beta and call for wider feedback.

PR #516, opened 2017-07-26
Tooltips

### LINKED ISSUE #85: tooltips don't automatically adjust position when they go off the edge of the screen
It's relatively straightforward to check positioning of the tooltip to make sure it's visible.  We should do this.

jasongrout:
Uses the Popper.js library to calculate tooltip position.

Fixes #85

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -29,8 +29,9 @@
     "webpack": "^1.12.14"
   },
   "dependencies": {
-    "d3": "^3.5.16",
     "@jupyter-widgets/base": "^0.6.4",
+    "d3": "^3.5.16",
+    "popper.js": "^1.0.0",
     "topojson": "^1.6.24",
     "underscore": "^1.8.3"
   },
--- a/js/package.json
+++ b/js/package.json
@@ -29,7 +29,7 @@
     "webpack": "^1.12.14"
   },
   "dependencies": {
-    "@jupyter-widgets/base": "^0.6.4",
+    "@jupyter-widgets/base": "^1.0.0",
     "d3": "^3.5.16",
     "popper.js": "^1.0.0",
     "topojson": "^1.6.24",
--- a/js/src/Bars.js
+++ b/js/src/Bars.js
@@ -175,7 +175,7 @@ var Bars = mark.Mark.extend({
               interactions.hover !== null) {
                 if(interactions.hover === "tooltip") {
                     this.event_listeners.mouse_over = this.refresh_tooltip;
-                    this.event_listeners.mouse_move = this.show_tooltip;
+                    this.event_listeners.mouse_move = this.move_tooltip;
                     this.event_listeners.mouse_out = this.hide_tooltip;
                 }
             } else {
--- a/js/src/Figure.js
+++ b/js/src/Figure.js
@@ -112,7 +112,7 @@ var Figure = widgets.DOMWidgetView.extend({
             .attr("transform", "translate(" + this.margin.left + "," + this.margin.top + ")");
         this.tooltip_div = d3.select(document.createElement("div"))
             .attr("class", "tooltip_div");
-        this.popper_reference = new popperreference.PopperReference(this.svg.node());
+        this.popper_reference = new popperreference.PositionReference({x: 0, y: 0, width: 20, height: 20});
         this.popper = new popper(this.popper_reference, this.tooltip_div.node(), {
             placement: 'auto',
         });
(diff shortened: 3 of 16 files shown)


ssunkara1:
@jasongrout This does look very promising! :)

jasongrout:
@dmadeka, @ssunkara1 - ready for review, I think.

jasongrout:
@Fezvrasta - just curious, how did you find out we were using Popper.js?

FezVrasta:
@jasongrout I monitor all the new GitHub issues that reference the word "popper" 😉

dmadeka:
@jasongrout How does one do the install? If you change the require, Popper wont be installed by just pip install -e .

npm install failed, whats the other option?

jasongrout:
@dmadeka - sorry, I forgot to commit a new file!

dmadeka:
@jasongrout Two issues, the Tooltip is too close to the mouse and should have its corner to the mouse, not center I guess? Also, are you seeing your figure plus toolbar from plt.show become centered?

jasongrout:
Right now, the tooltip snaps to the element the mouse is pointing at, not the mouse. There are options to offset it and put in an arrow.

Likely the figure+toolbar centering is fixed in https://github.com/jupyter-widgets/ipywidgets/pull/1547?

jasongrout:
> @jasongrout Two issues, the Tooltip is too close to the mouse and should have its corner to the mouse, not center I guess?

The tooltip attempts to avoid the mouse, if possible. If there is not room for the tooltip to be to the side, it may overlap the mouse. Are you seeing the tooltip overlap the mouse where there is enough room for the tooltip to the side?

FezVrasta:
If you want the tooltip to snap to the mouse you can pass a virtual reference element to Popper.js. It's described here:

https://popper.js.org/popper-documentation.html#referenceObject

jasongrout:
@FezVrasta - thanks.

We already pass a virtual reference since we want the tooltip to snap to whatever element we are hovering over: https://github.com/bloomberg/bqplot/pull/516/files#diff-87e9791defd138cea2122bf31305a269R17. Basically, our virtual reference object lets us change the reference element dynamically.

jasongrout:
@FezVrasta - we are seeing a weird flicker issue (basically, the tooltip flickers on and off once or twice on every mouse move, which is where we call for a scheduled update from popper) if the tooltip happens to be forced to overlap the reference element. I haven't tracked down whether that is in our code or popper's, but I'll let you know if it seems to be an issue in popper.

ssunkara1:
@jasongrout The tooltip placement is not quite right for `Lines` and `Pie`. You can check them in the `Mark Interactions` example under Interactions in the example folder.

jasongrout:
Great, thanks

jasongrout:
> The tooltip placement is not quite right for Lines and Pie

The tooltip is placed right beside the object. Are you saying we should always follow the mouse position? The thing about the mouse position is that it often obscures the thing you are hovering over.

It's an easy change to follow the mouse pointer with an offset. Is that preferred? @dmadeka? @rmenegaux? @ssunkara1?

ssunkara1:
@jasongrout I think following the mouse will be better.

jasongrout:
I changed mark interactions to follow the mouse.

We still have an issue in those examples: when the interaction is changed to show the tooltip on click, I'm not sure how to get the event for closing the tooltip. The tooltip just stays up since we never close it. Thoughts?

jasongrout:
(in particular, what event signifies that the tooltip should close when the tooltips are shown on mouse click?)

ssunkara1:
@jasongrout For the `click` tooltip, we close it whenever there is the `background_click` event.

jasongrout:
Actually, I just have code in the Mark.js `show_tooltip` and `hide_tooltip`. Is `hide_tooltip` not called whenever the tooltip needs to disappear, regardless of the event triggering the hide? It appears that `hide_tooltip` is not called when a background click is supposed to close the mark tooltip.

ssunkara1:
which `Mark` are you looking at? In the function `process_interactions`, we do have that for the event `event_listeners.parent_clicked` for `Lines`, `Bars` `Hist` etc..

stonebig:
any hope of quick resolution ?

jasongrout:
@ssunkara1 - what do you think of the PR now? I updated it to work with the ipywidgets 7.0 release.

jasongrout:
or @stonebig - can you test this PR?

stonebig:
it doesn't seem to work any better, but it may be:
- my fault/bad setup,
- a side-effect of ipyleaflet or other,
- a bqplot detail that is managed well only in anaconda / linux.

stonebig:
I get a message in javascript console:
````
lab:1 Uncaught (in promise) Module bqplot, semver range ^0.3.0-alpha.6 is not registered as a widget module
````

PR #518, opened 2017-07-27
model id rename bug fix

### LINKED ISSUE #517: Something in this version seems to break multiple HandDraws on the same figure
I have a Figure where I have a radio button that determines which of two HandDraw objects are supposed to be set as the interactive layer for the Figure.

In ipywidgets 6.0 and bqplot 0.9 this works well!

good_multiline_switching_2

But on the alpha of bqplot and the beta of ipython 7.0 this usage model no longer works:

bad_multiline_switching_2

ssunkara1:
@jasongrout  Because of a change in `ipywidgets' WidgeModel` (https://github.com/jupyter-widgets/ipywidgets/pull/1451) there was a bug in `bqplot` since we were relying on the `id` attribute of `WidgetModel`. This also fixes #517. Thanks @mpacer for reporting this.

### DIFF
--- a/js/src/BarsModel.js
+++ b/js/src/BarsModel.js
@@ -159,9 +159,9 @@ var BarsModel = markmodel.MarkModel.extend({
         });
         if(color_scale && color.length > 0) {
                 if(!this.get("preserve_domain").color) {
-                    color_scale.compute_and_set_domain(color, this.id + "_color");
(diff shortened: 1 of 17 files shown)


rmenegaux:
phew good catch Sri

jasongrout:
Whew, that was a lot! Thanks!

mpacer:
It looks like this fixed #517!

PR #521, opened 2017-07-28
clarify what date format is used for date based tick_formats

mpacer:
I had to dive into the js to figure out that it was expecting d3 based date-string formatting. This should be in the docstring. This might make sense in other places too… but I'm not sure the best way to search exhaustively for this kind of change.

### DIFF
--- a/bqplot/axes.py
+++ b/bqplot/axes.py
@@ -78,7 +78,7 @@ class Axis(BaseAxis):
     label: string (default: '')
         The axis label
     tick_format: string or None (default: '')
-        The tick format for the axis.
+        The tick format for the axis, for dates use d3 string formatting.
     scale: Scale
         The scale represented by the axis
     num_ticks: int or None (default: None)


bqplot/axes.py
```diff
@@ -78,7 +78,7 @@ class Axis(BaseAxis):
     label: string (default: '')
         The axis label
     tick_format: string or None (default: '')
-        The tick format for the axis.
+        The tick format for the axis, for dates use d3 string formatting.
```
dmadeka: Actually, we use the same formats as d3. So maybe a more general statement would be better? Or more detailed documentation? Not sure

ssunkara1:
@mpacer Thanks for this. I thought the `d3` time formatting was similar to the `strftime` function in python as they are both were inspired by the C library. Is there a specific time format which did not work like what you expected?

mpacer:
Yes, some of the unicode date format patterns did not work:

http://unicode.org/reports/tr35/tr35-dates.html#Date_Format_Patterns

Weirdly those has always worked even in bash with the BSD `date` command which ostensibly follows the C locale conventions, so I don't know what to tell you. 

It also just would have been nice to know where to look to figure out how to get what I wanted out of the different objects. 

Also, worth noting is that right now there is no way to use an ISO 8601 compatible date format specification for weeks (since %U ≠ %V = ww; %V was introduced in Python 3.6).

bqplot/axes.py
mpacer: I'm suggesting that this be made explicit here, as well as more generally. If I'm trying to figure out how to use a optional parameter on a function, I'm likely to look at its docstring. It wouldn't have helped to have a more general statement if I wouldn't have known to look for such a more general statement or where to look for it.

It should definitely be a general statement in a tutorial or example somewhere. But, I think of docstrings as more reference documentation than expository/educational documentation. Accordingly, either having information as close as possible to where the information is needed (even if that results in redundant information) or an explicit reference to where the information can be found is usually a good policy.

PR #524, opened 2017-08-01
Update Label.ipynb

dustinmichels:
Changed `font_size` to `default_size` in Label.ipynb, as `font_size` appears to be a deprecated parameter.


ssunkara1:
Good catch @dustinmichels! Thanks..

dustinmichels:
Great, no problem!

Issue #523, opened 2017-08-01
Feature request: Fill between 2 lines


DougRzz:
Hi,

I would like to plot 2 lines and fill in the area between the 2 lines.  See code below where I have used the inside option.  It would be useful to add a 'between' option.  Also, I noticed a bug - the fill opacities were not working for the 'inside' option but worked well with the 'bottom' and 'top' options.  

![image](https://user-images.githubusercontent.com/3510318/28827551-d8bb3cf0-76c5-11e7-99e0-29b35500a561.png)

```python
import ipywidgets as widgets
import bqplot as bq
import numpy as np

x_sc = bq.LinearScale()
y_sc = bq.LinearScale()

x_data = np.arange(100)
y_data = np.random.randn(2, 100)
y_data[1,:] = y_data[1,:] + np.random.randn(1, 100)

line_chart = bq.Lines(x=x_data, y=y_data, scales= {'x': x_sc, 'y': y_sc}, 
                      display_legend=True, labels=["line 1", "line 2", "line 3"],
#                       fill='bottom', # opacity does work with this option
                      fill='inside',   # opacity does not work with this option
#                       fill='between',   # This option does not exist.  Fill between 2 lines feature request
                      fill_opacities = [0.5] *len(x_data)
                     )

ax_x = bq.Axis(scale=x_sc)
ax_y = bq.Axis(scale=y_sc, orientation='vertical', tick_format='0.2f')

fig = bq.Figure(marks=[line_chart], axes=[ax_x, ax_y])
fig.layout.width = '95%'
widgets.VBox([fig])
```

ssunkara1:
@DougRzz : Thanks for reporting this. We are working on adding the `between` option. In the meanwhile, there is a small work-around you can use with the existing `inside` option.

```
def transform_to_area(x_data, y1, y2):
    return np.append(x_data, x_data[::-1]), np.append(y1, y2[::-1])

x_area, y_area = transform_to_area(x_data, y_data[0, :], y_data[1, :])
line_chart = bq.Lines(x=x_area, y=y_area, scales= {'x': x_sc, 'y': y_sc}, ...
```
This should work with fill='inside'.

DougRzz:
Thanks @ssunkara1 .   The work-around is very cool.

PR #525, opened 2017-08-01
Adding fill-between options in Lines

rmenegaux:
Adresses issue #523 

As it stands now, the last line will have a color that is different from the corresponding area (if there are n lines, there are only n-1 areas). A way around that is to set its opacity to 0, or to set the colors manually.
Any ideas on how to address this? Otherwise the PR is good to merge as is.

### DIFF
--- a/js/src/LinesModel.js
+++ b/js/src/LinesModel.js
@@ -98,7 +98,7 @@ var LinesModel = markmodel.MarkModel.extend({
                 });
             } else {
                 this.mark_data = curve_labels.map(function(name, i) {
-                    var xy_data = d3.zip(that.x_data[i], that.y_data[i][j]);
+                    var xy_data = d3.zip(that.x_data[i], that.y_data[i]);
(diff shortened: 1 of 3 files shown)


js/src/LinesModel.js
```diff
@@ -95,11 +98,13 @@ var LinesModel = markmodel.MarkModel.extend({
                 });
             } else {
                 this.mark_data = curve_labels.map(function(name, i) {
-                    var xy_data = d3.zip(that.x_data[i], that.y_data[i]);
+                    var xy_data = d3.zip(that.x_data[i], that.y_data[i][j]);
```
ssunkara1: I get an error saying `j` is not defined.

stonebig:
is there sometihng holding off this merge ? (and the 0.10.0 release)

SylvainCorlay:
Tested locally, this is really nice and a simple solution, merging.

PR #529, opened 2017-08-15
Version update for jupyterlab

maartenbreddels:
Needed for the latest @jupyter-widgets/jupyterlab-manager


jasongrout:
With the ipywidgets 7.0 release, this is now 0.26. When jlab is released on Monday, it will likely be bumped to 0.27.

PR #533, opened 2017-08-22
Fixup loading of popper for jupyter lab

SylvainCorlay:


jasongrout:
Looks good to me. This is because popper apparently changes how it loads itself depending on if requirejs is on the page?

Issue #544, opened 2017-09-06
scatter with extra args makes panzoom have error


jasongrout:
With this code:
```python
from bqplot import pyplot as plt

plt.figure()
plt.scatter([1,2,3], [4,3,2], skew=[4,3,2])
plt.show()
```
I click on the panzoom button and there is a js error: `TypeError: Cannot read property 'dimension' of undefined` at this line: https://github.com/bloomberg/bqplot/blob/005bde1a7f79b46f771ec607d5c0277d91d09722/js/src/Toolbar.js#L130

It seems to think that skew is a scale, and tries to get the dimension for that "skew" scale. This came up when a user was setting an attribute to use for a tooltip field.

ssunkara1:
Thanks pointing this out. As a work-around for displaying data in the tooltip you can use other attributes like `color`, `size` or `opacity`.
A slightly less hacky way to do that will be to listen to `on_hover` and set the data of a HTML table like widget.

Also, I think we should allow users to pass additional data with associated with each point which need not be represented visually in the plot, precisely for `Tooltips` and for controlling other widgets easily than performing a lookup based on index of the `hovered_point` everytime.

jasongrout:
In this usecase, we were trying *not* to have a visual representation of the skew data, hence the extra argument. Why and where is a scale created? I don't see where this happens on the python side. And should it be documented that extra arguments create scales, if that is what it is supposed to do?

ssunkara1:
The scale is also created for the first two attributes for `x` and `y`. Any data attribute when passed to `pyplot` creates a `scale` because it is needed for it to plot.

jasongrout:
Ah, thanks for the explanation, and thanks for fixing the bug above in your PR!

ssunkara1:
Closing this as the fix has been merged.

PR #546, opened 2017-09-07
adding missing attributes to scaled metadata

ssunkara1:
Fixes issue #544. @ChakriCherukuri can you have a look as I changed the `rtype` of `link_color` and also have a look if the scales for `link_color` and `color` should be the same by default or different.


ChakriCherukuri:
looks good @ssunkara1

PR #547, opened 2017-09-14
adding boxplot to pyplot

ssunkara1:


bqplot/pyplot.py
```diff
@@ -816,6 +816,30 @@ def bar(x, y, **kwargs):
...
+    Parameters
+    ----------
+
+    x: numpy.ndarray, 1d
+        The x-coordinates of the data points.
+    y: numpy.ndarray, 2d
+        The data from which the boxes are to be created. Each row of the data
+        corresponds to one box drawn in the plot.
+    options: dict (default: {})
+        Options for the scales to be created. If a scale labeled 'x' is
```
dmadeka: Shouldn't we customize this doc to the mark itself?

bqplot/pyplot.py
```diff
@@ -816,6 +816,30 @@ def bar(x, y, **kwargs):
...
+        required for that mark, options['x'] contains optional keyword
+        arguments for the constructor of the corresponding scale type.
+    axes_options: dict (default: {})
+        Options for the axes to be created. If an axis labeled 'x' is required
```
dmadeka: Same here

bqplot/pyplot.py
```diff
@@ -816,6 +816,30 @@ def bar(x, y, **kwargs):
     kwargs['y'] = y
     return _draw_mark(Bars, **kwargs)
```
dmadeka: Why isn't there a _process_data decorator? It will allow the mpl esque dataframe syntax

ssunkara1: I missed that. But when I was testing it, I think it is better if we do not have it for `boxplot` for now. Because the `y` of `boxplot` is a 2-d array and when we pass a list of column names, the data passed in to the mark is the transpose of what needs to be passed in and hence the plot output is misleading. this is also the case for `Lines`. I feel that is an issue we need to solve separately.

bqplot/pyplot.py
ssunkara1: you mean move to this to the base class itself? didn't completely follow.

bqplot/pyplot.py
ssunkara1: Added `process_data` and created issue #548 for fixing the issue mentioned.

rmenegaux:
@ssunkara1 is this good to merge?

ssunkara1:
yep!!  this is all set.

rmenegaux:
Looks good to me, merging

PR #562, opened 2017-10-27
Tutorials

ChakriCherukuri:
First pass at adding tutorials section, starting with an intro nb with links to tutorials. Added two tutorial nbs. Plan to keep updating this folder with more examples.

### DIFF
--- a/examples/Tutorials/_Introduction.ipynb
+++ b/examples/Tutorials/_Introduction.ipynb
@@ -1,0 +1,41 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "deletable": true,
+    "editable": true
+   },
+   "source": [
+    "### Introduction\n",
+    "`bqplot` is an interactive plotting library for the Jupyter notebook. The real power of `bqplot` is its interactivity. All the attributes of the marks are implemented as <b>traits</b> which let the user register callbacks and listen to their updates. Since `bqplot` is build using the widget framework of `ipywidets` it can be easily integrated with `ipywidgets`.\n",
+    "\n",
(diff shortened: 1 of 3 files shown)


examples/Tutorials/_Introduction.ipynb
```diff
@@ -0,0 +1,41 @@
...
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "deletable": true,
+    "editable": true
+   },
+   "source": [
+    "### Introduction\n",
+    "`bqplot` is an interactive plotting library for the Jupyter notebook. The real power of `bqplot` is its interactivity. All the attributes of the marks are implemented as <b>traits</b> which let the user register callbacks and listen to their updates. Since `bqplot` is build using the widget framework of `ipywidets` it can be easily integrated with `ipywidgets`.\n",
```
dmadeka: Interactive two times in a single paragraph. Marks are not defined - Im not sure this notebook should be included. It can be done as a README.md for this page.

This could be a useful reference:

1) https://github.com/fchollet/deep-learning-with-python-notebooks
2) https://github.com/zackchase/mxnet-the-straight-dope

ChakriCherukuri: @dmadeka , you suggest I add this to the main README with links to tutorial notebooks?

dmadeka: No, each folder can have a README. I suggest cleaning this up a bit and adding it to the README of this folder

dmadeka: https://stackoverflow.com/questions/19020891/is-there-a-way-to-create-a-folder-specific-readme-file-in-github

dmadeka: If we're doing a set of tutorials, I'd like that we do something like the links above - not something adhoc that we can't scale.

PR #563, opened 2017-10-28
Separation of scale name and attribute name

rmenegaux:
Changes the `scaled` metadata attribute so that it can specify a scale name that is different from the attribute's name.

This is useful for example for the Image mark, which has scaled attributes `x0`, `x1` that have the same scale 'x'

### DIFF
--- a/bqplot/marks.py
+++ b/bqplot/marks.py
@@ -66,6 +66,22 @@ def register_mark(key=None):
     return wrap
 
 
+def _check_scaled(scaled):
+    '''
+    Checks whether a trait is scaled or not
+    '''
+    return scaled is not None
+
+
+def _get_scale_name(trait):
+    '''
+    Returns the scale name for the passed trait, as specified
+    by its `scaled` attribute
+    '''
+    scaled = trait.get_metadata('scaled')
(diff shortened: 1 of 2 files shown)


ssunkara1:
cc @maartenbreddels 
@rmenegaux Thanks! What do you think about having `x0`, `x1` merged into a tuple called `x`? I am thinking if we should have scale name as being the same as the name of the data attribute or not.

rmenegaux:
@ssunkara1 I think we should try to keep the scale name the same as the attribute name as much as possible, but that it should not be a hard constraint.
The problem with merging them into a tuple is that you have to constrain the size to be 2, and also that you have to change the whole tuple even if you just would like to change x0.

It depends if you see the scale as attribute-specific or as mark-specific. We'll run into this issue anytime we have separate attributes using the same scale.

maartenbreddels:
I'm fine either way, but I can imagine this situation popping up again in the future.

SylvainCorlay:
Regarding the Image mark, I would be in favor of putting `x0, x1` and `y0, y1` in tuples. Checking the size can be done with a custom validators.

We should keep this proposal around though, if another case where we want to have different names for the variable and the scale.

PR #564, opened 2017-10-30
backporting boxplot width to 0.9

ssunkara1:


ChakriCherukuri:
@ssunkara1 tested this and looks good

PR #565, opened 2017-10-30
porting pyplot for boxplot to 0.9

ssunkara1:


ChakriCherukuri:
@ssunkara1 tested and looks good

PR #574, opened 2017-11-19
First pass at imshow

rmenegaux:
@SylvainCorlay @maartenbreddels 
Takes either an `ipywidgets` `Image` (default), a file name, or a byte string:
```python
import bqplot.pyplot as plt
from ipywidgets import Image

with open('sample.png', 'rb') as f:
    image = Image(value=f.read())
plt.figure()
plt.imshow(image)
plt.show()
```
or
```python
plt.imshow('sample.png', format='filename')
```
or
```python
with open('sample.png', 'rb') as f:
    plt.imshow(f.read(), format='openfile')
```

What do you think should be the default?

The matplotlib `imshow` takes a matrix as input so ideally we should support that too, although it would very close to what our `heatmap` does.

### DIFF
--- a/bqplot/pyplot.py
+++ b/bqplot/pyplot.py
@@ -705,7 +705,7 @@ def plot(*args, **kwargs):
         return _draw_mark(Lines, **kwargs)
 
 
-def imshow(image, format='widget', **kwargs):
+def imshow(image, format, **kwargs):
     """Draw an image in the current context figure.
     Parameters
     ----------
--- a/bqplot/pyplot.py
+++ b/bqplot/pyplot.py
@@ -51,12 +51,13 @@ import sys
 from collections import OrderedDict
 from IPython.display import display
 from ipywidgets import VBox
+from ipywidgets import Image as ipyImage
 from numpy import arange, issubdtype, array, column_stack, shape
 from .figure import Figure
 from .scales import Scale, LinearScale, Mercator
 from .axes import Axis
 from .marks import (
-        Lines, Scatter, Hist, Bars, OHLC, Pie, Map,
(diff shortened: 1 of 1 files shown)


maartenbreddels:
Love it, imshow is a must have ! :)
I'd say stay as close to matplotlib as possible, and add the option of passing a `ipywidgets.Image`. Even if a subset of imshow is close to what heatmap does, people (me at least) love API compatibility.

SylvainCorlay:
Regarding the format

1) small nitpicking suggestion: rename `ipyimage` to `widget` ? 
2) I propose the following logic

```python
    if format == 'widget':
        ipyimage = image
    elif format == 'filename':
        with open(image, 'rb') as f:
            data = f.read()
            ipyimage = ipyImage(value=data)
    else:
        ipyimage = ipyImage(value=image, format=format)
```

so that we pass through the existing formats of the widgets image.

SylvainCorlay:
@rmenegaux any update on the comment that I added? This should be a straightforward change. I think that we should get this out and push out a release, at least for the latest JupyterLab version support.

rmenegaux:
@SylvainCorlay voilà!

SylvainCorlay:
What do you think of *not* having a default value for the format?

If we do this, we will be able to add support for numpy arrays later on, and make it the default, without changing backward compatibility...

rmenegaux:
So make it a non-keyword arg?

SylvainCorlay:
hum, I guess, as soon as we keep it second. I am not very opinionated about this.

SylvainCorlay:
Woot!

PR #575, opened 2017-11-21
Enabled on_element_click for Graph mark

### LINKED ISSUE #500: "on_element_click()" does not seem to work for Graph mark
When constructing a Graph mark, it seems that I can set a callback function using, for example, `graph.on_background_click(print_event)` but setting a callback using `graph.on_click()` or `graph.on_element_click()` does not seem to do anything.

Really loving the Graph mark! Thank you for all your great work!!

ChakriCherukuri:
fixes #500

### DIFF
--- a/examples/Marks/Graph.ipynb
+++ b/examples/Marks/Graph.ipynb
@@ -31,8 +31,7 @@
     "| label |  str | node label | mandatory attribute |\n",
     "| label_display |  {center, outside, none} | label display options | center |\n",
     "| shape |  {circle, ellipse, rect} | node shape | circle |\n",
-    "| shape_attrs |  dict | node SVG attributes | {'r': 15} |\n",
-    "provide link to svg shape attrs\n",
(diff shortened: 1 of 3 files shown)


ChakriCherukuri:
@dustinmichels please review

examples/Marks/Graph.ipynb
```diff
@@ -197,56 +244,108 @@
...
+   "metadata": {
+    "collapsed": true,
+    "deletable": true,
+    "editable": true
+   },
    "outputs": [],
    "source": [
-    "# click on any node to select it (selected attribute gets populated with the selected node)\n",
-    "graph.selected"
+    "import bqplot.pyplot as plt\n",
```
ssunkara1: I think we can move this to the cell right after the tooltip. I think that feels better

ssunkara1:
Thanks @ChakriCherukuri Apart from the small notebook comment, it looks good.

ChakriCherukuri:
@ssunkara1 made changes per your recommendation. Please review

PR #583, opened 2017-12-08
Add JupyterLab integration test to Travis Script

blink1073:
This ensures that the lab extension installs correctly and that the app loads without any console errors or uncaught errors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,8 @@ python:
     - 3.3
     - 3.4
     - 3.5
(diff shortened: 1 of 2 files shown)


blink1073:
Once we have the `jupyter labextension check` command that can be used as well.  cf https://github.com/jupyterlab/jupyterlab/issues/3335

SylvainCorlay:
Thanks. This looks good to me. Merging!

PR #584, opened 2017-12-11
Defaults fixes

ChakriCherukuri:
1. Made default color in bar chart same for all bars (steelblue). Currently bar chart defaults to different colors for each bar which is confusing.
2. Consolidated default colors across marks to have same color (steelblue)
3. Cleaned up Scatter and Bars example notebooks

PS: Chose 'steelblue' as default color since it's similar to the first color of CATEGORY10 and also similar to matplotlib's default color.

### DIFF
--- a/bqplot/marks.py
+++ b/bqplot/marks.py
@@ -513,7 +513,7 @@ class Scatter(_ScatterBase):
     marker: {'circle', 'cross', 'diamond', 'square', 'triangle-down',
              'triangle-up', 'arrow', 'rectangle', 'ellipse'}
         Marker shape
-    colors: list of colors (default: ['DeepSkyBlue'])
+    colors: list of colors (default: ['steelblue'])
         List of colors of the markers. If the list is shorter than the number
         of points, the colors are reused.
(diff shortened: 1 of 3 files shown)


SylvainCorlay:
LGTM. Merging!

ssunkara1:
@ChakriCherukuri Thanks for this. I think the default value of `colors` for `Bar` and `Lines` should be list of values rather than just one, as both of them support plotting 2-D `y` values. We do have it for `Lines` but with this change, `Bar` has only one color.

bqplot/marks.py
```diff
@@ -858,7 +858,7 @@ class Bars(Mark):
         Y respectively.
     type: {'stacked', 'grouped'}
         whether 2-dimensional bar charts should appear grouped or stacked.
-    colors: list of colors (default: CATEGORY10)
+    colors: list of colors (default: ['steelblue'])
```
ssunkara1: For a 1-D `Bar`  to have the same color for each bar, you can change the default value of the attribute `color_mode`. I prefer that rather than having just one default color.

PR #586, opened 2017-12-14
Race condition in the Image

rmenegaux:
The Image was sometimes not rescaling at all (some kind of a race condition). This cleans up and fixes that issue.
Animation is an added bonus.

### DIFF
--- a/js/src/Image.js
+++ b/js/src/Image.js
@@ -36,7 +36,7 @@ var Image = mark.Mark.extend({
 
         return base_render_promise.then(() => {
             this.create_listeners();
-            this.parent.displayed.then(() => {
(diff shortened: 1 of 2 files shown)


maartenbreddels:
That may solve a mysterious bug I've seen where the image indeed doesn't respect the scales. 👍

SylvainCorlay:
@rmenegaux I opened a PR to your branch fixing the race condition for the size of the figure.

SylvainCorlay:
Can you confirm everything is fine for you. The race condition occurred very rarely even before the fix.

rmenegaux:
Yes this fixes the race condition. This is ready to merge now

PR #591, opened 2017-12-21
Backend hist

rmenegaux:
Same as #429, but keeping the old `Hist` and naming the backend Hist `Bins`


bqplot/marks.py
```diff
@@ -931,6 +933,101 @@ def _validate_orientation(self, proposal):
...
+    # Mark decoration
+    icon = 'fa-signal'
+    name = 'Backend Histogram'
+
+    # Scaled Attributes
+    sample = Array([]).tag(
+        sync=True, display_name='Sample', rtype='Number',
+        atype='bqplot.Axis', **array_serialization
+        ).valid(array_squeeze, array_dimension_bounds(1, 1))
+
```
maartenbreddels: sync=False right, I guess this should not be send to the front-end?

rmenegaux: Yes good catch

rmenegaux:
@maartenbreddels updated according to your comments, does it lgty?

maartenbreddels:
It has my blessings 👍

PR #600, opened 2018-01-21
Adding a tutorial for making a feature vector distribution plot

cdknorow:


rmenegaux:
Thanks @cdknorow , there's only a minor bug that I'll fix in an upcoming PR

PR #613, opened 2018-02-16
Fixup binder environment

SylvainCorlay:


dmadeka:
What about `setup.py`?

https://github.com/bloomberg/bqplot/blob/master/setup.py#L184

SylvainCorlay:
These are the dependencies to run the example notebooks. bqplot does not actually depend on sklearn.

dmadeka:
Does the pip install respect the dependencies in the yaml file?

SylvainCorlay:
Not at all. The environmwnt.yml is a manisfest for the binder build

PR #623, opened 2018-03-14
fixed display_labels attr in pie animation

vsub21:
issue with pie animation example; display_labels supplied str value instead of boolean, fixed so running works properly


ChakriCherukuri:
@vsub21 In master 'display_labels' attribute is an enum with three possible values: {'none', 'inside', 'outside'}. It's not a boolean.

vsub21:
Apologies -- just realized I was working of the most recent version of bqplot on py3.4, so my examples did not execute properly. Cheers!

PR #629, opened 2018-03-28
new: Image mark has pixelated trait, to avoid interpolation

maartenbreddels:
For e.g. scientific plots, you want to see the pixels.

### DIFF
--- a/bqplot/marks.py
+++ b/bqplot/marks.py
@@ -1566,6 +1566,7 @@ class Image(Mark):
     _view_name = Unicode('Image').tag(sync=True)
     _model_name = Unicode('ImageModel').tag(sync=True)
     image = Instance(widgets.Image).tag(sync=True, **widget_serialization)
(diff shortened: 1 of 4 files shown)


maartenbreddels:
This can probably followed up by an interpolation argument for imshow, I think 'none' or 'nearest' would be almost the same as this.

rmenegaux:
@maartenbreddels cool! Should it default to `False` though? I'm not sure what the main use of the Image mark is.

maartenbreddels:
mpl is now using it as default, so we may follow that: https://matplotlib.org/users/dflt_style_changes.html#interpolation
On top of that, it will by default show up if we have browser issues ;)

maartenbreddels:
Another option is to use WebGL to draw the image onto a canvas with the same resolution as the screen, then we have much more control over the interpolation.

maartenbreddels:
Replaced by https://github.com/bloomberg/bqplot/pull/630

PR #630, opened 2018-04-03
WIP: imshow is more matplotlib-like, where the colorscale is applied on the frontend

maartenbreddels:
interpolation replaces #629 , which is also more matplotlib like and future proof
if we want to support other interpolation schemes.
Maybe it can be made more bqplot-like, since i'm not sure i understand fully the scales.


maartenbreddels:
This is too slow for larger images (1024^2), I'm working on a WebGL/Threejs version.

rmenegaux:
@maartenbreddels looks good. One disadvantage though: the `Image` mark would not be able to render pngs, jpgs anymore right?

maartenbreddels:
Ok, this uses WebGL, which is quite a bit faster (although the toDataURL isn't the fastest). Also added a opacity.
I'm using a per figure context (not per image), since there is always a limit to the number of contexts (renderers in threejs) you can have.

maartenbreddels:
Just a heads up, this approach needs to change a bit. The idea is that there is like 1 overlay image, that represents OpenGL output, and the image widget is rendered to this svg image. The idea behind this is that if you have a 4k image, it will just render to a say 500x400 image, giving a much better performance. Now the svg image is the same size as the widget image.

mwcraig:
@maartenbreddels -- would you be OK with me moving the stuff in here related to interpolation to a new PR (or could pull from #629).

Agree that a more performant image viewer would be nice, but in the meantime just getting pixelated or interpolation would be good.

maartenbreddels:
I propose to have similar future interface as what I describe in https://github.com/jupyter-widgets/ipywidgets/issues/2114 so that we can have support for png/jpeg, as well as intensity and rgb(a) 8 bit data. The intensity does requires a colorscale, so that part can only be translated to the ipywidget.Image widget once scales move to the ipyscales package.

maartenbreddels:
@mwcraig that would be ok for me btw!

maartenbreddels:
Although future dev on this PR may make it irrelevant actually

maartenbreddels:
Rebased to https://github.com/bloomberg/bqplot/pull/684 (so do not merge until that is accepted).

This new method is way more efficient. An canvas 'overlay' is produced, which matches the size of the viewport. Now the image (which may be huge) will be drawn as texture on top of this overlap.
(Before the canvas had the same size as the overlay, meaning a 2k image would result in a 2k canvas, really slow!).
An even better approach would be to really do the overlay in the html-DOM so we do not have to grab the resulting image and put it in the svg-DOM.
Also, we now have better control over interpolation, since it is done by webgl, which will probably always work.

TODO
 - [ ] Get #684 merged. 
 - [ ] Drop the interpolation css part
 
Improvements/ideas: we need some way to add extra scales to the image, for instance first take the log, and then do a colormap.

maartenbreddels:
Closing this in favor of https://github.com/glue-viz/glue-jupyter/pull/50 , making a new astro specific widget that uses bqplot: https://github.com/glue-viz/ipyastroimage

PR #631, opened 2018-04-10
Scatter scale size range should starts at 0

maartenbreddels:
The minimum of the range of the size_scale is no zero, which although might be convenient in some cases does not to predictable behaviour. For instance, see below where the min of the scale is set to zero, the Scatter's size equals y, but because of the minimum size of 10 (at the minimum of the domain), goes below zero (as it extrapolates):
screen shot 2018-04-10 at 13 34 56
Instead, if simply set to zero, it will lead to the expected behaviour:
screen shot 2018-04-10 at 13 35 35
If a users wants to have a minimum size, this can easily be computed by changing the domain accordingly.

### DIFF
--- a/js/src/ScatterBase.js
+++ b/js/src/ScatterBase.js
@@ -18,9 +18,6 @@ var _ = require("underscore");
 var utils = require("./utils");
 var mark = require("./Mark");
 
-var min_size = 10;
-
-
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
I am 👍 on this. Leaving this open a little bit so that other can chime in.

PR #638, opened 2018-04-26
Selectors example notebook

rmenegaux:
This PR adds a simple example notebook for selectors and fixes a couple of issues:
- The IndexSelector `color` attribute did not work on initialisation
- Selectors did not listen to changes in `marks`


SylvainCorlay:
@jasongrout are you aware of the JupyterLab extension issues that we are observing here with Python 2?

jasongrout:
> @jasongrout are you aware of the JupyterLab extension issues that we are observing here with Python 2?

No. What are you observing?

SylvainCorlay:
> No. What are you observing?

The build failure on python 2, which appears to be legitimate.

jasongrout:
Oh wow: `The log length has exceeded the limit of 4 MB` - haven't seen that before on Travis!

I'm trying to rebuild something that passed previously to understand it better.

jasongrout:
Yep, #644 solves the issue.

Separately, I've opened https://github.com/jupyterlab/jupyterlab/issues/4513 that may fix the root issue.

SylvainCorlay:
Thanks for fixing the issue @jasongrout !

PR #645, opened 2018-05-11
Pep8 cleanup

ChakriCherukuri:
1. Fixed PEP8 warnings in the python files
2. Tested to make sure all the example notebooks are running fine

### DIFF
--- a/bqplot/__init__.py
+++ b/bqplot/__init__.py
@@ -21,7 +21,10 @@ BQPlot Package
 .. currentmodule:: bqplot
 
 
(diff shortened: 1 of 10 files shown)


dmadeka:
Looks like the commit history is pretty broken here

kaiayoung:
LGTM, reviewed as requested by Chakri

jasongrout:
> Looks like the commit history is pretty broken here

Can you elaborate, @dmadeka? It looks like each commit is fixing pep8 warnings in a different file.

dmadeka:
@jasongrout Yup, but they're all titled the same. So it looks like 10 straight commits of the same name, which isnt very informative. It might make sense to add more details to the commit names, or squashing them into a single comment

PR #647, opened 2018-05-16
Using pyplot in marks example notebooks

ChakriCherukuri:
Use pyplot in all the mark example notebooks. Code reduced significantly!
Tested all notebooks to make sure they run without any errors and reproduce previous results


kaiayoung:
I'm also a bit hesitant to completely eliminate the object model examples, there are a few use cases that `pyplot` doesn't currently cover.

dmadeka:
+1 to @kaiayoung , also given our complete API - wouldnt it make sense to have multiple examples and then interactively change the plot? We could say - `Bars`, `pyplot`, 'pandas` and `Object Model`

ChakriCherukuri:
@dmadeka and @kaiayoung sorry I didn't mean to completely remove the object model from examples.  In fact a few marks are not exposed in pyplot (MarketMap, Graph etc.). **Where ever possible we should encourage users to use pyplot**. That's what I did in my commits. I noticed just one instance  where it was not possible to use pyplot and I used the object model. 

Currently the format is to use object model in all the examples and just have one nb for pyplot which doesn't cover most of the mark examples in detail. IMHO we should reverse this format to use pyplot in all the example nbs and have one nb explaining the object model for doing complex uses cases.

Also this gives us the opportunity to enhance pyplot (which should be very easy) to take care of all the use cases. 

I am pretty sure very few will use bqplot if their first experience with the library is the object model. That's what I am hearing from both internal and external users who keep asking me if there's a simpler API. Like any good library, we should just expose the user to the minimalistic API and give them the lower level object model only for doing advanced stuff. I rarely use the object model myself even for building rich dashboards!

ChakriCherukuri:
@kaiayoung per our discussion closing this PR. Will submit a new one where I add (instead of update) pyplot examples

PR #650, opened 2018-05-18
Added Mark example notebooks using pyplot

ChakriCherukuri:
1. Added two new folders in the Marks folder: Object Model and Pyplot
2. Moved the existing notebooks to 'Object Model' folder
3. Added pyplot examples to the Pyplot folder
4. Updated the Index notebook to provide the links to both the Object Model and Pyplot implementations


ChakriCherukuri:
@rmenegaux can you please review?

PR #656, opened 2018-05-25
Improve classic notebook layout

### LINKED ISSUE #641: Listen to window resize in the classic notebook for bqplot relayout

SylvainCorlay:
This PR makes three changes related to layout

 1. It fixes #641, meaning that if you set the layout of a figure to be e.g. 100%, the figure was not being re-laid out like it was in JupyterLab.
 2. ~~It makes the natural width of the Figure 100%. Natural min-width and min-height are 480px.~~
 3. Min and max aspect ratio are set to 0 and 100 (an arbitrary large number corresponding to something very oblong because JSON does not support IEEE infinities).

- Item (1) is a bug fix, which was impacting users of the classic notebook.
- Item (3) is a minor change that should not have much impact expect in complicated JupyterLab layouts.
- ~~item (2) is a major change that may change people's layouts in applications. However I like how charts now take up more real estate by default.~~

Screencast of figure getting the relayout upon window resize in the classic notebook. (when setting `layout.width='100%'`)

full-width

### DIFF
--- a/bqplot/figure.py
+++ b/bqplot/figure.py
@@ -136,10 +136,9 @@ class Figure(DOMWidget):
     # min width is based on hardcoded padding values
     layout = LayoutTraitType(kw=dict(min_width='125px'))\
         .tag(sync=True, **widget_serialization)
-    min_aspect_ratio = Float(1.0).tag(sync=True)
-    # Max aspect ratio is such that we can have 3 charts stacked vertically
-    # on a 16:9 monitor: 16/9*3 ~ 5.333
-    max_aspect_ratio = Float(6.0).tag(sync=True)
+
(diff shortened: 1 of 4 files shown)


SylvainCorlay:
cc @ChakriCherukuri 

I think that you should be interested in this. If the layout change is an issue, I can separate the bug fix (adding a resize event listener) in another PR.

SylvainCorlay:
Dynamic width in JupyterLab: 

full-width-jlab

SylvainCorlay:
@ChakriCherukuri I removed item (2) so that it is backward compatible and can be merged quickly and will open another PR with the layout change.

To get the result of the screencast, one can still do `fig.layout.width = '100%'`

PR #661, opened 2018-06-15
ScatterGL, a million scatter points using WebGL

maartenbreddels:
~~This continues #630, not sure what the best strategy is, history wise etc.~~

cc @astrofrog
For https://github.com/glue-viz/glue-jupyter we need bqplot to handle larger scatter plots, say up to 100k-1M. This PR shows it is possible to do using WebGL (based on ipyvolume code), and I think a first good attempt. It doesn't support yet all features the Scatter mark has, for instance, only linear scales. 

The demo shows it runs pretty smooth, except that at 1 million, the non-binary serialization causes quite a lag. 
scattermega_example


SylvainCorlay:
This is really cool!

mprogram:
just FYI, I'm getting the following trace with `ipywidgets = 7.3.1` after the plot in the second input:

`---------------------------------------------------------------------------`
`KeyError                                  Traceback (most recent call last)`
`/usr/lib/python3/dist-packages/ipywidgets/widgets/widget.py in _handle_msg(self, msg)`
`    667                 if 'buffer_paths' in data:`
`    668                     _put_buffers(state, data['buffer_paths'], msg['buffers'])`
`--> 669                 self.set_state(state)`
`    670 `
`    671         # Handle a state request.`
``
`/usr/lib/python3/dist-packages/ipywidgets/widgets/widget.py in set_state(self, sync_data)`
`    537                     from_json = self.trait_metadata(name, 'from_json',`
`    538                                                     self._trait_from_json)`
`--> 539                     self.set_trait(name, from_json(sync_data[name], self))`
`    540 `
`    541     def send(self, content, buffers=None):`
``
`/usr/lib/python3/dist-packages/bqplot/traits.py in array_from_json_binary(value, obj)`
`    168 def array_from_json_binary(value, obj=None):`
`    169     if value is not None:`
`--> 170         return np.frombuffer(value['buffer'], dtype=value['dtype']).reshape(value['shape'])`
`    171 `
`    172 def array_from_binary_json(value, obj=None):`
``
`KeyError: 'buffer'`

It doesn't affect functionality though (as of commit `36e1209`).

mprogram:
I truly like the implementation. This is just to mention, that bqplot/examples/Advanced Plotting/Animations.ipynb seems not to be working with this branch.

maartenbreddels:
Thanks, it is still WIP, but thanks for the feedback, I'll need to check that. This branch continues from #630, which continues #684, which continues from #680 (which just got merged), I wouldn't be surprised if #684 breaks it already. 
But thanks for the feedback, keep it coming!

maartenbreddels:
@mprogram those issues are partly resolved in #684, and I rebased to that.

mprogram:
I've merged locally #684, #630, #696 (`rmenegaux:latex`), #661, #701, #705 (`davidbrochart:drag_map`)  #706 (`box_plot_outlier_detection`), – in the present order, and to mention all of my local merges – resolving trivial interim conflicts. However, I could no longer run the above example, it doesn't show anything. In addition, I'm getting the following warning under `jupyter-lab build`: 
```log
WARNING in d3
  Multiple versions of d3 found:
    3.5.17 ./~/d3 from ./~/bqplot/src/ScatterMegaModel.js
    4.6.0 ./~/ipyvolume/~/d3 from ./~/ipyvolume/lib/figure.js
```
(with the master branch of `ipyvolume`)

The above referenced example `bqplot/examples/Advanced Plotting/Animations.ipynb`, however, runs just fine with all the merges, except of the pie animation, for which it further differs if the notebook is run cell-by-cell, or from the 'run all cells' menu.

_(for the curious like me: the comment is about WIP, it may not be related to a finished release)_

maartenbreddels:
Some findings:
I've been doing some benchmarking/comparison, and it seems that Chrome is quite slow with large dataframes over websocket, and firefox is really fast, using a million points with firefox is no problem at all, it renders it almost instantly (the non-smoothness is mostly the limited captured fps):
scatter-mega-1e6
1 million is possible with chrome, but it just takes a few seconds for each ~4m array to transfer.
And 10 million is even possible with firefox, not super comfortable/smooth though:
scatter-mega-1e7

maartenbreddels:
The issue with chrome exists only when the dev console is open, which means that profiling (which opens de dev console) gives a distorted performance.

DougRzz:
@maartenbreddels On a vaguely related point, I created a set of linked plots to visualise and interactively edit 8 dimensions.  It was just for fun (sad I know).   Basically, it involves jslinking the skew, rotation, color, skew and opacity attributes with the x and y attributes.   The animated gif shows this implimented with the standard Scatter mark. With 25 points it started to get a little sluggish.  Would the ScatterMega handle many more points for this type of linked plots?  

Would love to see this PR in BQplot master sometime soon :D


 dimensionalViewer_8D

dfreeman06:
I have been also developing something similar. 
scatterplotmatrix

DougRzz:
@dfreeman06   That is really nice.   I was planning on doing something very similar.  I use Bokeh at the moment for this sort of thing because it performs reasonably well for large numbers of points.  But I am sure ScatterMega will be even better!
To get all the figure plot areas the same, I create extra figure widgets which just contain a vertical or horizontal axes.  Are you ok to share the code?

martinRenou:
@maartenbreddels I could rebase without too many difficulties. I'll do the same changes to `ScatterGL` and `ScatterGLModel` as we did for the other files (es6 classes, var -> const etc...)

martinRenou:
@maartenbreddels I'm done concerning the TypeScript/ES6 migration. The TypeScript build fails for tests though... I'm not sure how to fix it. Would you mind taking a look at it?

Errors are concerning ThreeJS. The weird thing is that it compiles fine when doing `npm install`, but not when doing `npm run test`. 
```
ERROR in /home/travis/build/bloomberg/bqplot/js/src/ScatterGL.ts
[tsl] ERROR in /home/travis/build/bloomberg/bqplot/js/src/ScatterGL.ts(29,23)
      TS2350: Only a void function can be called with the 'new' keyword.
ERROR in /home/travis/build/bloomberg/bqplot/js/src/ScatterGL.ts
[tsl] ERROR in /home/travis/build/bloomberg/bqplot/js/src/ScatterGL.ts(52,22)
      TS2339: Property 'needsUpdate' does not exist on type 'DataTexture'.
ERROR in /home/travis/build/bloomberg/bqplot/js/src/ScatterGL.ts
[tsl] ERROR in /home/travis/build/bloomberg/bqplot/js/src/ScatterGL.ts(201,66)
      TS2339: Property 'fromGeometry' does not exist on type 'BufferGeometry'.
ERROR in /home/travis/build/bloomberg/bqplot/js/src/ScatterGL.ts
[tsl] ERROR in /home/travis/build/bloomberg/bqplot/js/src/ScatterGL.ts(302,35)
      TS2350: Only a void function can be called with the 'new' keyword.
ERROR in /home/travis/build/bloomberg/bqplot/js/src/glattributes.ts
[tsl] ERROR in /home/travis/build/bloomberg/bqplot/js/src/glattributes.ts(36,29)
      TS2350: Only a void function can be called with the 'new' keyword.
ERROR in /home/travis/build/bloomberg/bqplot/js/src/glattributes.ts
[tsl] ERROR in /home/travis/build/bloomberg/bqplot/js/src/glattributes.ts(53,29)
      TS2350: Only a void function can be called with the 'new' keyword.
ERROR in /home/travis/build/bloomberg/bqplot/js/src/glattributes.ts
[tsl] ERROR in /home/travis/build/bloomberg/bqplot/js/src/glattributes.ts(282,22)
      TS2339: Property 'normalized' does not exist on type 'InstancedBufferAttribute'.
ERROR in /home/travis/build/bloomberg/bqplot/js/src/glattributes.ts
[tsl] ERROR in /home/travis/build/bloomberg/bqplot/js/src/glattributes.ts(295,22)
      TS2339: Property 'normalized' does not exist on type 'InstancedBufferAttribute'.
```

martinRenou:
I think the difference is that we are using `ts-loader` when building the tests

astrofrog:
As discussed on Gitter, the following doesn't work with ``ScatterGL`` but works with ``Scatter``:

```python
import bqplot
scale_x = bqplot.LinearScale(min=0, max=1)
scale_y = bqplot.LinearScale(min=0, max=1)
scale_color = bqplot.ColorScale()
scales = {'x': scale_x, 'y': scale_y, 'color': scale_color}
axis_x = bqplot.Axis(scale=scale_x, grid_lines='solid', label='x')
axis_y = bqplot.Axis(scale=scale_y, orientation='vertical', tick_format='0.2f', grid_lines='solid', label='y')
figure = bqplot.Figure(scales=scales, animation_duration=0, axes=[axis_x, axis_y])
scatter = bqplot.ScatterGL(scales=scales, x=[0, 0.33, 0.66, 1], y=[0, 0.3, 0.7, 1])
scatter.color = [10., 20., 30., 40]
scale_color.min = 0
scale_color.max = 50
scale_color.scheme = 'RdYlGn'
figure.marks = list(figure.marks) + [scatter]
figure
```

**ScatterGL**
Screenshot from 2019-04-26 18-16-27


**Scatter**
Screenshot from 2019-04-26 18-16-37

js/src/ScatterGL.ts
```diff
@@ -0,0 +1,981 @@
...
+               .attr('width', fig.plotarea_width)
+               .attr('height', fig.plotarea_height)
+        this.camera.left = 0
+        const x_scale = this.scales.x ? this.scales.x : this.parent.scale_x;
+        const y_scale = this.scales.y ? this.scales.y : this.parent.scale_y;
+
+        const range_x = this.parent.padded_range("x", x_scale.model);
+        const range_y = this.parent.padded_range("y", y_scale.model);
+
+        this.scatter_material.uniforms['colormap'].value = create_colormap(this.scales.color)
```
martinRenou: Is it reasonable to create the colormap texture for each frame?

Another (more optimized?) strategy would be to create it only when the colorscale value changes.

js/shaders/scatter-vertex.glsl
```diff
@@ -0,0 +1,214 @@
...
+    // times 4 because of the normalized coordinates, and radius vs diameter use
+    pixel_size = sqrt(mix(SCALE_SIZE(size_previous), SCALE_SIZE(size), animation_time_size)) * marker_scale * 4.;
+    // we draw larger than the size for the stroke_width (on both side)
+    float s = pixel_size + 2.0 * stroke_width;
+    vUv = uv;
+    float angle = SCALE_ROTATION(mix(rotation_previous, rotation, animation_time_rotation));
+    vec3 model_pos = rotate_xy(position, 1.) * s + center_pixels;
+
+#ifdef USE_COLORMAP
+    float color_index = color; //(color - domain_color[0]) / (domain_color[1] - domain_color[0]);
```
martinRenou: The `domain_color` value is never used and never updated, which results in the colormap use case not working. I'll see if I can fix it.

martinRenou: It is defined as an `attribute` in shaders, but I guess it should be a `uniform`

martinRenou: Fixed

martinRenou:
@maartenbreddels I fixed and cleaned a bit the arrow implementation.

Before:
triangle_GL2

After:
triangle_GL

martinRenou:
@maartenbreddels I also added a small ellipse function in the fragment shader which allows to simply create an ellipse, we could add it as a marker type later :)

ellipse_GL

martinRenou:
We need to check whether we would have performances improvements when using `DYNAMIC_DRAW` with the `THREE.InstancedBufferAttribute`s. See https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/bufferData#Parameters

js/src/ScatterGL.ts
```diff
@@ -0,0 +1,913 @@
...
+            blendEquation: THREE.AddEquation,
+            blendSrc: THREE.OneFactor,
+            blendDst: THREE.OneMinusSrcAlphaFactor,
+
+            blendEquationAlpha: THREE.AddEquation,
+            blendSrcAlpha: THREE.OneFactor,
+            blendDstAlpha: THREE.OneMinusSrcAlphaFactor,
+        });
+
+        this.buffer_marker_geometry = new THREE.BufferGeometry().fromGeometry(new THREE.PlaneGeometry());
```
martinRenou: This should be replaced by `this.buffer_marker_geometry = new THREE.PlaneBufferGeometry();`, not only because it improves readability, but also because it reduces the number of vertices (12 vs 18 vertices), and there is an unused extra `color` attribute in the `PlaneGeometry`.

js/src/ScatterGL.ts
martinRenou: fixed

js/shaders/scatter-fragment.glsl
```diff
@@ -33,39 +33,43 @@ vec2 rotate_xy(vec2 x, float angle) {
 
 void main(void) {
     vec2 pixel = (vUv - 0.5) * (pixel_size + 2.0 * stroke_width);
+    float fill_weight = 0.0;
 
 #if FAST_DRAW == FAST_CIRCLE
-    float r = length(pixel);
-    float fill_weight = 1.0 - smoothstep(pixel_size/2.0 - stroke_width - SMOOTH_PIXELS, pixel_size/2.0 - stroke_width + SMOOTH_PIXELS, r);
+    float r = length(pixel); // This uses `sqrt` which is slow, there might be room for performance improvements
```
maartenbreddels: Yes, we could work in 'squared' space.

js/shaders/scatter-fragment.glsl
```diff
@@ -33,39 +33,43 @@ vec2 rotate_xy(vec2 x, float angle) {
...
-    fill_weight *= (fill ? 1.0 : 0.0);
-
 #elif FAST_DRAW == FAST_SQUARE
-    float stroke_weight_inverse = smoothstep(-SMOOTH_PIXELS, SMOOTH_PIXELS, -(abs(pixel.x) - pixel_size/2.0 + stroke_width/2.0))
-                                * smoothstep(-SMOOTH_PIXELS, SMOOTH_PIXELS, -(abs(pixel.y) - pixel_size/2.0 + stroke_width/2.0));
-    float stroke_weight = 1.0 - stroke_weight_inverse;
-    float fill_weight =  1.0 * smoothstep(-SMOOTH_PIXELS, SMOOTH_PIXELS, -(abs(pixel.x) - pixel_size/2.0 + stroke_width/2.0))
-                             * smoothstep(-SMOOTH_PIXELS, SMOOTH_PIXELS, -(abs(pixel.y) - pixel_size/2.0 + stroke_width/2.0));
-    fill_weight *= (fill ? 1.0 : 0.0);
+    if (fill) {
```
maartenbreddels: Did you benchmark this, I assumed a multiply to be faster than an if.

js/shaders/scatter-fragment.glsl
martinRenou: Yes that's what I did later on :) on this commit 516013c

js/shaders/scatter-fragment.glsl
martinRenou: I wanted to replace it with an `#if FILL`, but I changed my mind and put back the `fill_weight *= (fill ? 1.0 : 0.0);`, that's on the current status of the branch

martinRenou: I guess `#if FILL` would not be a bad idea?

js/shaders/scatter-fragment.glsl
```diff
@@ -35,10 +35,13 @@ vec2 rotate_xy(vec2 x, float angle) {
 
 /*
  * Returns 1.0 if pixel inside of a circle (0.0 otherwise) given the circle radius and the
- * pixel distance from the circle center.
+ * pixel position.
  */
-float circle(in float radius, in float dist) {
-    return 1.0 - smoothstep(radius - SMOOTH_PIXELS, radius + SMOOTH_PIXELS, dist);
+float circle(in float radius, in vec2 pixel_position) {
+    float d = pow(pixel_position.x, 2.0) + pow(pixel_position.y, 2.0);
```
maartenbreddels: isn't this equivalent to `dot(pixel_position, pixel_position)`, which might be faster it the optimization path doesn't recognize it.

martinRenou: It is! Good catch

maartenbreddels:
Very *very* nice refactor @martinRenou, great work.

js/shaders/scatter-fragment.glsl
maartenbreddels: I'm not sure why I did not do that. I think that is much better.

martinRenou:
Well, you did the hard work ;)

martinRenou:
@maartenbreddels I quickly added a "cross" implementation in the shaders. It's not perfect, no using the smoothstep so we might have some aliasing when the cross is turning.

cross

martinRenou:
We are not taking the `skew` into account

bqplot/marks.py
```diff
@@ -522,8 +522,9 @@ class _ScatterBase(Mark):
         'opacity': {'dimension': 'opacity'},
         'rotation': {'dimension': 'rotation'}
     }).tag(sync=True)
-    default_opacities = List(trait=Float(1.0, min=0, max=1, allow_none=True))\
-        .tag(sync=True, display_name='Opacities')
+    default_opacities = Array(None, allow_none=True)\
```
martinRenou: I'm not entirely sure, but I think this change breaks the `default_opacities` in the default Scatter implementation

martinRenou: Fixed c167c2a

maartenbreddels:
This is good to go I think!

PR #662, opened 2018-06-15
aligning label to be in the center

ChakriCherukuri:
This fix aligns labels to be in the center of a rectangle (for e.g. in GridHeatMap if we add labels they are not aligned in center)


SylvainCorlay:
Hey, this looks good to me. Can you make a quick screenshot for the record?

ChakriCherukuri:
before
after

ChakriCherukuri:
before the fix and after the fix

SylvainCorlay:
OK, thanks!

PR #665, opened 2018-06-26
fix for box plot x-axis alignment with ordinal scale

mbektasbbg:


SylvainCorlay:
LGTM thanks @mbektasbbg !

PR #666, opened 2018-06-26
auto box width calculation for box plot

mbektasbbg:


SylvainCorlay:
Thanks for tackling this!

Since you change the default value for `box_width`, you also need to update the default value in the front-end in `BoxplotModel`.

https://github.com/bloomberg/bqplot/blob/master/js/src/BoxplotModel.js#L37

mbektasbbg:
thanks @SylvainCorlay ! I overlooked js model update need. I will push a new commit soon with the change

SylvainCorlay:
LGTM!

PR #669, opened 2018-07-03
Add tick_transform parameter

kaiayoung:
I’ll follow up with a more comprehensive reworking later, but wanted to at least get the main fixes in for now:
- Added `tick_transform` parameter, which allows users to rotate the tick axis labels: 
   - `tick_transform = {‘rotate’: -35}`
   `tick_style = {'text-anchor': 'end'}`
- Fixed several state management issues that caused the ticks to be redrawn incorrectly if attributes such as `tick_style` are modified after the plot was created.

### DIFF
--- a/bqplot/axes.py
+++ b/bqplot/axes.py
@@ -111,6 +111,8 @@ class Axis(BaseAxis):
         Dictionary containing the CSS-style of the text for the ticks.
         For example: font-size of the text can be changed by passing
         `{'font-size': 14}`
+    tick_rotate: int (default: 0)
+        Degrees to rotate tick labels by.
     """
     icon = 'fa-arrows'
     orientation = Enum(['horizontal', 'vertical'], default_value='horizontal')\
@@ -136,6 +138,7 @@ class Axis(BaseAxis):
 
     visible = Bool(True).tag(sync=True)
     tick_style = Dict().tag(sync=True)
+    tick_rotate = Int(0).tag(sync=True)
 
     _view_name = Unicode('Axis').tag(sync=True)
     _model_name = Unicode('AxisModel').tag(sync=True)
--- a/js/src/Axis.js
+++ b/js/src/Axis.js
@@ -23,7 +23,7 @@ Math.log10 = Math.log10 || function (x) {
     return Math.log(x) / Math.LN10;
 };
 
-var units_array = ["em", "ex", "px"];
+var UNITS_ARRAY = ["em", "ex", "px"];
 
 var Axis = widgets.WidgetView.extend({
 
@@ -35,21 +35,14 @@ var Axis = widgets.WidgetView.extend({
 
     render: function() {
         this.d3el.style("display", this.model.get("visible") ? "inline" : "none");
-
         this.parent = this.options.parent;
         this.margin = this.parent.margin;
-        this.vertical = this.model.get("orientation") === "vertical";
         this.height = this.parent.height - (this.margin.top + this.margin.bottom);
         this.width = this.parent.width - (this.margin.left + this.margin.right);
 
-        var scale_promise = this.set_scale(this.model.get("scale"));
-        this.side = this.model.get("side");
-        this.padding = this.model.get("padding");
-        this.num_ticks = this.model.get("num_ticks");
-        this.label_loc = this.model.get("label_location");
-        this.label_offset = this.extract_label_offset(this.model.get("label_offset"));
+        var scale_promise = this.set_scale_promise(this.model.get("scale"));
+        var offset_promise = this.get_offset_promise();
 
-        var offset_promise = this.get_offset();
         var that = this;
         Promise.all([scale_promise, offset_promise]).then(function() {
             that.create_listeners();
@@ -60,6 +53,8 @@ var Axis = widgets.WidgetView.extend({
     },
 
     create_listeners: function() {
+        // Creates all event listeners
+
         this.listenTo(this.model, "change:scale", function(model, value) {
             this.update_scale(model.previous("scale"), value);
             // TODO: rescale_axis does too many things. Decompose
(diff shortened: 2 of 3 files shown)


SylvainCorlay:
Thanks for getting started on this! This looks pretty good.

Quick nitpicking comment: regarding the changes from `snake_case` to `camelCase`, I am not sure we should partially change things like this. It would make the patch easier to read if we kept a uniform notation.

(I would not be opposed to adopt camelCase, but a better occasion would be another PR where we already have to get over the entire code base mechanically, like a JS to TS transition).

kaiayoung:
PR has been updated to use snake case instead.
@SylvainCorlay @ChakriCherukuri

js/src/Axis.js
```diff
@@ -514,81 +565,73 @@ var Axis = widgets.WidgetView.extend({
...
+        this.rescaleAxis();
     },
 
     update_visibility: function(model, visible) {
         this.d3el.style("display", visible ? "inline" : "none");
     },
 
     get_ticks_from_array_or_length: function(data_array) {
-        // This function is to be called when the ticks are passed explicitly
+        // This function is to be called when the ticks are passed explicilty
```
SylvainCorlay: `explicitly`

js/src/Axis.js
```diff
@@ -514,81 +565,73 @@ var Axis = widgets.WidgetView.extend({
...
+        this.apply_tick_styling();
     },
 
     parent_margin_updated: function() {
         // sets the new dimensions of the g element for the axis.
         this.margin = this.parent.margin;
         this.width = this.parent.width - this.margin.left - this.margin.right;
         this.height = this.parent.height - this.margin.top - this.margin.bottom;
-        this.rescale_axis();
+        this.rescaleAxis();
```
SylvainCorlay: you missed a camel!

js/src/Axis.js
```diff
@@ -514,81 +565,73 @@ var Axis = widgets.WidgetView.extend({
...
     get_ticks_from_array_or_length: function(data_array) {
-        // This function is to be called when the ticks are passed explicitly
+        // This function is to be called when the ticks are passed explicilty
         // or the number of ticks to be drawn.
         // Have to do different things based on the type of the scale.
         // If an array is passed, then just scale and return equally spaced
         // points in the array. This is the way it is done for ordinal
         // scales.
         var step, max;
+        let num_ticks = this.model.get("num_ticks");
```
SylvainCorlay: +1 on making this a local variable.

js/src/Axis.js
```diff
@@ -35,22 +35,15 @@ var Axis = widgets.WidgetView.extend({
...
         this.height = this.parent.height - (this.margin.top + this.margin.bottom);
         this.width = this.parent.width - (this.margin.left + this.margin.right);
 
-        var scale_promise = this.set_scale(this.model.get("scale"));
-        this.side = this.model.get("side");
-        this.padding = this.model.get("padding");
-        this.num_ticks = this.model.get("num_ticks");
-        this.label_loc = this.model.get("label_location");
-        this.label_offset = this.extract_label_offset(this.model.get("label_offset"));
+        let scale_promise = this.set_scale_promise(this.model.get("scale"));
```
SylvainCorlay: Ping @jasongrout. Wonder if we should use babel to allow the use of ES6 syntax in bqplot?

(`const`, `let`).

kaiayoung:
@SylvainCorlay ES6 syntax has been removed.

bqplot/axes.py
```diff
@@ -111,6 +111,10 @@ class Axis(BaseAxis):
         Dictionary containing the CSS-style of the text for the ticks.
         For example: font-size of the text can be changed by passing
         `{'font-size': 14}`
+    tick_transform: Dict (default: {})
+        Dictionary containing the SVG transform attributes to be passed to the
+        axis ticks.
+        For example: `{'rotation': 90}`
```
SylvainCorlay: Typo: `rotation` -> `rotate`.

js/src/Axis.js
```diff
@@ -60,68 +53,78 @@ var Axis = widgets.WidgetView.extend({
...
     create_listeners: function() {
+        // Creates all event listeners
+
         this.listenTo(this.model, "change:scale", function(model, value) {
             this.update_scale(model.previous("scale"), value);
-            // TODO: rescale_axis does too many things. Decompose
+            // TODO: resacle_axis does too many things. Decompose
             this.axis.scale(this.axis_scale.scale); // TODO: this is in redraw_axisline
-            this.rescale_axis();
+            this.resacle_axis();
```
SylvainCorlay: typo.

js/src/Axis.js
```diff
@@ -60,68 +53,78 @@ var Axis = widgets.WidgetView.extend({
...
-                that.update_grid_lines();
-            });
-        }, this);
         this.parent.on("margin_updated", this.parent_margin_updated, this);
     },
 
+    update_offset: function() {
+        var offset_creation_promise = this.get_offset_promise();
+        var that = this;
+        offset_creation_promise.then(function() {
```
maartenbreddels: Should we use arrow functions? only IE11 doesn't support it?
cc @SylvainCorlay @jasongrout
Or should we keep this before moving to ES6+babel?

js/src/Axis.js
```diff
@@ -235,64 +280,71 @@ var Axis = widgets.WidgetView.extend({
...
-            .attr(this.get_label_attributes())
-            .style(this.get_text_styling())
-            .text(this.model.get("label"));
-
-        this.set_tick_values();
+            .attr("class", "axislabel");
+        
+        // Apply custom settings
+        this.update_label_settings();
+        this.update_label_settings();
```
maartenbreddels: repeated line

js/src/Axis.js
```diff
@@ -235,64 +280,71 @@ var Axis = widgets.WidgetView.extend({
...
         this.update_grid_lines();
-        this.tickstyle_changed();
-        this.update_color();
-        this.update_label();
+        this.update_line_color();
+        this.apply_tick_styling();
     },
 
-    get_offset: function() {
+    get_offset_promise: function() {
```
maartenbreddels: offtopic: Good reason to go ES6! (async/await)

js/src/Axis.js
jasongrout: > Or should we keep this before moving to ES6+babel?

Or Typescript, which also compiles down to either es5 or es6.

SylvainCorlay: I think moving to typescript is likely to make us lose contributors.

Maybe the separated scales are a good place for typescript.

maartenbreddels: Yes, I'd say ES6 for code, typescript for unittests maybe?
But should we move to fat arrows functions yet?

jasongrout: We can move to typescript as *just* as es6->es5 translator. Or we can just use es6 and insist the user has a recent browser. I'm just saying we don't have to use Babel. The nice thing about using TS as an es6->es5 translator is that it's easy to gradually introduce typing as it makes sense, and we also automatically get type checking for widgets stuff.

My guess is that just using TS to compile even our existing js will show us many bugs :). That's been my experience with converting codebases, anyway.

maartenbreddels: That is true, it caught some bugs in ipyvolume as well, so stay with js, but use ts for compiling to es5 sounds like a good idea.

kaiayoung:
@maartenbreddels Changed `tick_transform` to `tick_rotate`.

SylvainCorlay:
Merging!

SylvainCorlay:
Thanks Kaia for this PR and sorry for all the nitpicking!

maartenbreddels:
Cool, thanks @kaiayoung!

SylvainCorlay:
This is breaking a number of things unfortunately. Fixing up and opening a PR.

kaiayoung:
@SylvainCorlay What are the problems that you are seeing?

SylvainCorlay:
The formatting of ticks of axis is not set properly.

`tickValues` and `tickFormat` on the d3 axis should still be camelCase but
it is not sufficient to fix it all so there is a bit more work.

I was working on this earlier this afternoon but I had to leave before I
could complete.

S.

PR #673, opened 2018-07-11
Fix widget in the documentation

martinRenou:
The show function now returns the displayed widget, it is needed by `jupyter-sphinx` because it checks the return value of the last line of code: https://github.com/jupyter-widgets/jupyter-sphinx/blob/master/jupyter_sphinx/embed_widgets.py#L208

### DIFF
--- a/bqplot/pyplot.py
+++ b/bqplot/pyplot.py
@@ -153,9 +153,12 @@ def show(key=None, display_toolbar=True):
     if display_toolbar:
         if not hasattr(figure, 'pyplot'):
             figure.pyplot = Toolbar(figure=figure)
-        display(VBox([figure, figure.pyplot]))
+        vbox = VBox([figure, figure.pyplot])
+        display(vbox)
+        return vbox
     else:
         display(figure)
+        return figure
 
 
 def figure(key=None, fig=None, **kwargs):
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -19,7 +19,7 @@ Using the pyplot API
     x = np.linspace(0.0, 10.0, n)
     y = np.cumsum(np.random.randn(n))
     plt.plot(x, y)
-    plt.show()
+    plt.show(display_toolbar=False)
 
 Using the bqplot internal object model
 


SylvainCorlay:
Thanks! This should not be needed because we override the _ipython_display_ of Widget in jupyter_sphinx.

This used to work but somehow, the changes in ipython's display mechanism broke the bwhavior. Cc @carreau.

dmadeka:
If we do this, wouldn't it mean that `plt.show` has to be the last line of a cell to display the plot?

martinRenou:
Ok, so this issue is more complicated than I thought. 

In `jupyter_sphinx` I decided to directly overwrite the `IPython.display.display` function in order to not depend on the `display` implementation anymore, it works fine (See my implementation here: https://github.com/martinRenou/jupyter-sphinx/blob/fix_display/jupyter_sphinx/embed_widgets.py#L138). 

But that's not the only issue resulting in this "widget not displayed" issue in the bqplot documentation. The problem is that when building the "autosummary" using "sphinx.ext.autosummary" extension, `bqplot` module is loaded, and it happens before the `.. ipywidgets-setup:: ...` code is executed, so the overwrite of `IPython.display.display` is not effective on bqplot.

So I could display the widget by using my implementation of the overwrite of display AND by removing the autosummary, but removing the autosummary is not a solution.

I have no idea on how to fix it properly. I couldn't find a way to "force" the overwrite of the display, even for modules already imported. Any idea @SylvainCorlay ?

SylvainCorlay:
OK a first answer is that we should probably drop autosummary because most of the informations that it displays are not very userful.

Beyond the case of  bqplot, we should make it work even when autosummary is used...

One way could be to do the monkey patching earlier, such as when `ipywidgets` is imported by the extension. Although, we would (maybe) need to monkey patch it with a hook that allows us to 
 - not display anything in the case of the setup code.
 - do what we already do in the case of the widget code.

martinRenou:
I'll open an issue in `jupyter_sphinx`, or a PR if I can. Let's close this one

PR #680, opened 2018-07-17
Set up testing framework using karma and headless chrome

maartenbreddels:
I'm still not happy with this setup, it doesn't lend itself to TDD since it takes 10 seconds for the first tests to run, and on a file change can take 5 seconds. Also, when typescript gives and error, karma quits directly.
Another setup, which is much faster, using webpack results in having no sourcemapping. For this setup (which I'm not using here) I had to move to webpack 4 as well. If this is an issue I can check if this all works with webpack 2.


maartenbreddels:
Ok, got it working with webpack+ts-loader, which works much more nicely.
Workflow:
 * run `$ (cd js; ./node_modules/.bin/karma start)` (will keep chrome open)
 * edit file, save it, karma detects it and reruns

If you need debugging, change HeadlessChrome to Chrome in karma.conf.json, and remove the --headless flag, click the 'Debug' button in the chrome window, and you should have source mapping as well.

maartenbreddels:
This seems to work nicely now, a review would be good.

js/webpack.config.js
```diff
@@ -1,22 +1,24 @@
...
-    { test: /\.css$/, loader: "style-loader!css-loader" },
-    { test: /\.less$/, loader: "style-loader!css-loader!less-loader" },
-    { test: /\.json$/, loader: "json-loader" },
-    { test: /\.(jpg|png|gif)$/, loader: "file" },
+const path = require('path');
+
+var rules = [
+    { test: /\.css$/, use: [{loader: "style-loader"}, {loader: "css-loader" }]},
+    { test: /\.less$/, use: [{loader: "style-loader"}, {loader: "css-loader" }, {loader: "less-loader" }]},
+    // { test: /\.json$/, loader: "json-loader" },
```
SylvainCorlay: Do we still use the JSON loading in bqplot?

js/test/dummy-manager.ts
```diff
@@ -0,0 +1,111 @@
...
+    constructor(library: any) {
+        super();
+        this.el = window.document.createElement('div');
+        window.document.body.appendChild(this.el)
+        this.library = library;
+    }
+
+    display_view(msg: services.KernelMessage.IMessage, view: Backbone.View<Backbone.Model>, options: any) {
+        // TODO: make this a spy
+        // TODO: return an html element
```
SylvainCorlay: Looks like the second TODO is satisfied :)

SylvainCorlay:
This looks good to me!

I like the idea of writing more pure JS tests before any full integration testing.

js/webpack.config.js
maartenbreddels: I'll have to look into it, but webpack itself failed with this enabled, and everything works without.

maartenbreddels: webpack 4 has a default loader, and the json loader is used to get the version from package.json

maartenbreddels:
Ok, this has worked quite well for me, I think it's ready for review/merge.

SylvainCorlay:
I am happy to merge this know that the release is out. There is a conflict due to the 0.11 release.

maartenbreddels:
ok, it was a trivial conflict, solved now.

PR #684, opened 2018-07-19
Binary serialization

maartenbreddels:
~~Continues from #680 (if that is merged I'll rebase)~~

This adds binary serialization to many of the Marks (scatter and line seems to work).
- [x] basic binary serialization 
- [x] support for dates (this is using ms resolution, and float64 to transfer)
- [ ] see if we can reuse the front-end code and feed this back to https://github.com/vidartf/ipydatawidgets/tree/master/packages/serializers/src (cc @vidarf )
- [x] check all marks (histogram doesn't seem to work)
- [x] use serializers instead of on the fly deserialization
- [x] find a way for datetime's to survive the roundtrip
- [x] `Lines`
- [x] `Label` (and check other text arrays)
- [x] `Bars`
- [x] `Bins`
- [x] `Boxplot`
- [x] `Candles`
- [x] `GridHeatMap`
- [x] `HeatMap`
- [x] `Hist`
- [x] `MarketMap`
- [x] `Pie`
- [x] Selectors
- [ ] Remove `BaseModel` and `get_typed_field/set_typed_field`
 - [ ] Since date is now a number, tick_format in Axis fails.

Since some of the code is coming from ipyvolume, I left in support for 'keyframes', i.e. the data arrays can be or one dimension higher, allowing for full frontend based animation. Do we want to go that route?

Also, do we want to use ndarray? For 1d arrays I'm fine with TypedArrays for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,8 @@ script:
         sh -e /etc/init.d/xvfb start || true;
         jupyter labextension install js @jupyter-widgets/jupyterlab-manager;
         python -m jupyterlab.selenium_check;
+        (rm -rf js/node_modules);
+        (cd js; npm install);
         (cd js; ./node_modules/.bin/karma start)
       fi
 after_success:
--- a/bqplot/marks.py
+++ b/bqplot/marks.py
@@ -51,8 +51,8 @@ from traittypes import Array
 from numpy import histogram
 
 from .scales import Scale, OrdinalScale, LinearScale
-from .traits import (Date, array_serialization,
-                     array_squeeze, array_dimension_bounds)
+from .traits import (Date, array_serialization, 
+                     array_squeeze, array_dimension_bounds, array_supported_kinds)
 from ._version import __frontend_version__
 from .colorschemes import CATEGORY10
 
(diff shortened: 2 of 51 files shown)


rmenegaux:
@maartenbreddels I tested all the marks, the list is below.

One problem is that arrays of text (e.g. for Label and MarketMap) are not handled by the binary serializer, they should be changed to the 'classic' one. I haven't looked into the JS errors for the other Marks though...

#### Work
- FlexLine
- Graph
- Scatter
- Image
- Map
- Lines (most cases)

#### Don't work (serializing error because data is not float/int/date)
- Lines (several lines w/ different number of points)
   ```python 
    Lines(x=[[0, 1], [0, 1, 2]], y=[[0, 1], [1, 0, -1],)
   ```
   because `np.asarray([[0, 1], [0, 1, 2]]).dtype = ‘O’`
- Label (problem with the label text)
- Market Map

#### Don't work (JS error)
- Bars
- Bins
- Boxplot
- Candles
- GridHeatMap
- HeatMap
- Hist
- Market Map

maartenbreddels:
Ok, I've looked into it, and in many places there is the expectation that the data can be 2d, but of irregular shape. That means that you can't really convert it to a numpy array (ok, of type object).
I propose two solutions:
 * only opt in for binary serialization for Scatter, Image, and the rest that work.
 * Have a new trait type, ListArray that will allow an array, or list of arrays, which is I think an ok solution, but some more work.

So it depends if we wanna solve this completely now, or have some results now quickly, and iterate later on. Thoughts @SylvainCorlay ?

maartenbreddels:
Discussed this with Sylvain, the support for irregular arrays was an accident, and there are two options:
 * not support it
 * fill with nans

I'll go for the first option first, and then when everything seems to work, I'll work on the 2nd item.

SylvainCorlay:
> - not support it
> - fill with nans

In any case, the "fill with nan" thingy would be a backend change only.

maartenbreddels:
It's 90% done I think. I check many of the notebooks, and there are only a few issues (non-breaking). I think we can totally can get rid of `BaseModel` as well.

Issue #678, opened 2018-07-12
Allowing LaTeX in bqplot Axis or Figure variables


JuanCab:
This is a feature request:

Any chance we could see support for LaTeX in the bqplot.Axis.label and bqplot.Figure.title traits?  This is a feature of ipywidgets.Label and ipywidgets.HTML and matplotlib that I exploit extensively and I would love to see the ability move over to ballot.  I am certain it is not trivial, but I through I would ask.

maartenbreddels:
Great to see movement on this, keeping an eye on it for ipyvolume!

lukasbongartz:
Has there be any movement in that regard?

PR #696, opened 2018-08-02
(WIP) LateX rendering

rmenegaux:
This aims to enable LateX rendering to titles, legends, ticks and labels (#678).

### DIFF
--- a/js/src/Figure.js
+++ b/js/src/Figure.js
@@ -18,6 +18,7 @@ var d3 = require("d3");
 var _ = require("underscore");
 var popperreference = require("./PopperReference");
 var popper = require("popper.js");
+var utils = require("./utils.js")
 
 if (popper.__esModule) {
     popper = popper.default;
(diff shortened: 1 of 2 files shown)


rmenegaux:
With the current code, MathJax runs but does not render anything. 

```python
import bqplot.pyplot as plt

plt.figure(title='L$\\alpha$tex')
plt.show()
```
outputs 'Ltex'.

Here is what the html looks like:
screen shot 2018-08-02 at 21 05 21

However, it does render correctly if I wrap the svg text in a foreignElement (commented JS code in the PR). That is something we might want to avoid though because it would mean manually rendering axis ticks rather than using d3's inbuilt ones.

@jasongrout any ideas on this?

rmenegaux:
Yes that's right. The issue is the same in both notebook and lab.

rmenegaux:
So the problem is that SVG refuses to render anything that is not an SVG element, unless it is inside a foreignObject. 
**Good news**: there is a MathJax configuration which outputs SVG, which would be the beginning of the answer to all our problems.
**Bad news**: that configuration is set by the notebook (or the lab) and cannot be changed afterwards.  Unfortunately for us, the current MathJax output configuration is common html/css.

Unless the notebook MathJax configuration is changed, it seems we have 2 choices:
- Wrap the Tex in a foreignObject tag. That would work fine for the figure title, legend and axis labels, but would be more problematic for axis ticks and the Label Mark (we'd have to hack the d3 ticks or write our own, and even then SVG transforms like scale, rotation would not apply)
- Manually convert the MathJax output into something digestible by SVG. That would be a lot of work and probably ugly.

Thoughts?

maartenbreddels:
I thought of https://khan.github.io/KaTeX/ but it does not have [svg output](https://github.com/Khan/KaTeX/issues/375)
The alternative would be (similar to what I do with scattermega) to render to an image, and include that.

jasongrout:
MathJax 3 (in beta) is also way more flexible in these regards. However, we probably shouldn't be in the business of providing a mathjax implementation?

maartenbreddels:
@jasongrout any idea on how beta is beta, and what their timeline is?
Also, what do you mean with the last sentence? That bqplot shouldn't include a latex renderer (mathjax) but that the environment should provide it?

jasongrout:
> That bqplot shouldn't include a latex renderer (mathjax) but that the environment should provide it?

Yes. To me that makes sense.

jasongrout:
> @jasongrout any idea on how beta is beta, and what their timeline is?

@dpvc is one of the most careful individuals I've ever met (that's a good thing!). I don't know the precise timeline for mathjax 3, but there has been steady progress, and given past experience, I am confident that whatever is implemented/published is pretty high quality.

edwinytgoh:
Hi there! I was wondering what the current status of this WIP is?

JuanCab:
Just checking to see if any movement is possible on this feature request (has the state of MathJax support for SVG changed)?

PR #701, opened 2018-08-09
Add setter to BrushSelector.selected

maartenbreddels:
Added a setter for backwards compatibility. Although I've changed the type to be always an array (to avoid edge cases). Also added a unittest using py.test.


maartenbreddels:
Actually, maybe this should be done with traitlets, since now observe doesn't work. It is used in one of the examples (examples/Interactions/Selectors.ipynb)

rmenegaux:
Lgtm, ready to merge @maartenbreddels?

maartenbreddels:
:+1:

(from mobile phone)

PR #705, opened 2018-08-12
Drag map

davidbrochart:
This implements dragging to rotate the map.


davidbrochart:
It doesn't play so well with the zoom (the map gets off).

davidbrochart:
Since we can drag the map now, zoom doesn't need to depend on mouse cursor position.

PR #714, opened 2018-09-10
WebGL support for bqplot (threejs based)

maartenbreddels:
This adds support for WebGL based on threejs (although I think we can also support regl). This will be the basis for #661 and #630.

It works by overlaying a canvas on top of the svg (for performance reasons). However, when a screenshot is needed, the canvas is converted to an image and embedded.

### DIFF
--- a/js/src/Figure.js
+++ b/js/src/Figure.js
@@ -18,6 +18,9 @@ var d3 = require("d3");
 var _ = require("underscore");
 var popperreference = require("./PopperReference");
 var popper = require("popper.js");
+var THREE = require('three')
+
+THREE.ShaderChunk['scales'] = require('raw-loader!../shaders/scales.glsl')
 
 if (popper.__esModule) {
     popper = popper.default;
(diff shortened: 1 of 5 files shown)


maartenbreddels:
@SylvainCorlay do you feel like merging this or should we first attempt to recover from lost webgl context, similar to what jupyter-threejs is doing (cc @vidartf), and make ipyvolume, jupyter-threejs and bqplot use a similar strategy for recovering.
We may also want to align this with the plotly folks (cc @jonmmease @jackparmer) so that we are not all fighting for a webgl context.

js/src/Figure.js
```diff
@@ -18,6 +18,9 @@ var d3 = require("d3");
 var _ = require("underscore");
 var popperreference = require("./PopperReference");
 var popper = require("popper.js");
+var THREE = require('three')
+
+THREE.ShaderChunk['scales'] = require('raw-loader!../shaders/scales.glsl')
```
SylvainCorlay: should we namespace scales?

SylvainCorlay: bqscales?

PR #724, opened 2018-09-28
WIP: Added a more general ToolbarWidget

maartenbreddels:
This is more of an exploration of doing toolbars (for bqplot and ipyvolume) a different way.

By making a ToolbarWidget that is a container (it has a single child widget), it can create a view of that child widget and send it messages (save to png, fullscreen).
What is does now as well, is keep a list of Interacts, and the current index and current value, which can be linked to a ToggleButtons widgets. I think a more general approach of this is to have a 'WidgetList' widget, that will keep a list, an index, and a current widget.

cc @astrofrog
In the attached screenshot (or the toolbar notebook) I used the glue icon using this PR https://github.com/jupyter-widgets/ipywidgets/pull/2182
screen shot 2018-09-28 at 16 31 19

Maybe we should talk about using/including/modifying these icons, for a uniform UI for bqplot/ipyvolume/glue-jupyter

### DIFF
--- a/bqplot/toolbar.py
+++ b/bqplot/toolbar.py
@@ -26,10 +26,10 @@ Toolbar
    Toolbar
 """
 
-from traitlets import Unicode, Instance, Bool
-from ipywidgets import DOMWidget, register, widget_serialization
+from traitlets import Unicode, Instance, Bool, CInt, List, Tuple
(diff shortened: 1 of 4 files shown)


maartenbreddels:
cleaned up version, using https://github.com/jupyter-widgets/ipywidgets/pull/2232

screen shot 2018-10-05 at 20 13 14

PR #740, opened 2018-10-26
New: scaling up of png images and have a good default for high dpi screens

maartenbreddels:
by default, it matches what you see on the screen, for instance on a mac with `window.devicePixelRatio == 2` it will give you an image of twice the resolution.


maartenbreddels:
Left: previous result (scale=1). Right: new result (scale=2), much sharper!
screen shot 2018-10-26 at 16 00 38

SylvainCorlay:
Backported to 0.11.x

PR #742, opened 2018-10-29
fixed an issue with marker strings ("default_colors" renamed to "colors")

### LINKED ISSUE #741: Issue with pyplot marker strings
Pyplot marker strings currently not working as expected. This is because the attribute "default_colors" was recently renamed to "colors"

ChakriCherukuri:
fixes #741


SylvainCorlay:
Thanks!

PR #744, opened 2018-10-29
MarketMap renamed to TileMap

### LINKED ISSUE #743: Rename MarketMap to TileMap
MarketMap name is confusing and not many people know what exactly it does. TileMap is a more relevant name

ChakriCherukuri:
1. Renamed MarketMap to TileMap
2. Tested code and all example notebooks

fixes #743


SylvainCorlay:
This looks good to me, although it is a backward-incompatible change.

We are probably going to tag a bug-fix patch release before moving forward with this.

ChakriCherukuri:
@SylvainCorlay do you want me to add the MarketMap back (while keeping the TileMap) with deprecation warning?

SylvainCorlay:
I think it is fine as is. I just want to tag a conservative patch release very soon. Then we can start with breaking changes in master (theming, renamings, megamarks).

PR #748, opened 2018-11-05
Scale number of suggested axis ticks for DateScale

kaiayoung:
D3's default number of axis ticks often results in x axis tick labels overlapping each other when when displaying date information on a figure with a narrow width. For example, when multiple plots are displayed in a grid using the ipywidgets `GridBox` widget. This PR reduces the number of suggested ticks if the width falls below a specified threshold, and the user did not specify their own number.

### DIFF
--- a/js/src/Axis.js
+++ b/js/src/Axis.js
@@ -23,6 +23,7 @@ Math.log10 = Math.log10 || function (x) {
     return Math.log(x) / Math.LN10;
 };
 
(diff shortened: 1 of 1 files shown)


kaiayoung:
@maartenbreddels for awareness. Not sure why it won't let me add you as a reviewer?

maartenbreddels:
I do not have commit right, happy to review it thought!

(from mobile phone)

SylvainCorlay:
backported to 0.11.x

PR #750, opened 2018-11-08
Support arrays with object type if they are all strings

maartenbreddels:
Since pandas is using dtype=object for strings, we need to support that again.
This makes the Wealth of Nations work again.


ChakriCherukuri:
LGTM

PR #751, opened 2018-11-08
Fix/workaround for when nan's enter brush selector causing infinite loop

maartenbreddels:
Because of https://github.com/jupyter-widgets/traittypes/issues/30 selected and selected_x keep updating each other, if they contain a nan.


maartenbreddels:
Even though it should not happen, it exposed a different bug, I don't think we should merge this. Other PR coming to fix the underlying bug.

PR #752, opened 2018-11-08
Extend fill_opacities to patches

### LINKED ISSUE #747: Lines.fill_opacities ignored for fill='inside'
Hi!
I noticed in [Lines.js:197](https://github.com/bloomberg/bqplot/blob/a4917aedbc8caab8b6fcb966b0ae039bd261edba/js/src/Lines.js#L197) that there is explicit logic to fill with the line opacity _only_ in the case of using `fill='inside'`. The end result is that it is then not possible to fill lines inside with transparency (ultimately leading to weirdness when trying to use interactions).

This otherwise works as expected for the other fill types. Is there a reason the logic has been implemented this way? It seems counter intuitive to me.

rmenegaux:
Fixes #747

Also adds an example of an area chart in the Lines example notebook.

### DIFF
--- a/examples/Marks/Object Model/Lines.ipynb
+++ b/examples/Marks/Object Model/Lines.ipynb
@@ -82,7 +82,9 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {},
+   "metadata": {
+    "collapsed": true
+   },
    "outputs": [],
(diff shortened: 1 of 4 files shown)


SylvainCorlay:
backported to 0.11.x

Issue #710, opened 2018-08-17
Boxplot not responding to panning


renanxcortes:
Hello, I'm trying to develop some graphs using bqplot and I have a question regarding boxplots. Here's a toy example of I was trying to make:

`import bqplot.pyplot as plt
x = [1]
fig = plt.figure()
boxes = plt.boxplot(x, [[3,7,6,4,8,1,4]], box_fill_color='lightblue', outlier_fill_color='black')
plt.ylabel('Teste')
plt.show()
`

However, when I trying to pan the graph using the toolbar button, the boxplot doesn't move, although the axes move which can lead to visual misinterpretation. Is there a way to solve that?

Issue #698, opened 2018-08-06
Boxplot mark issues


kaiayoung:
Some noted issues with the Boxplot mark:

**Race condition occasionally prevents mark from being displayed.**
- Run the following code and run the cell several times, about 1/10 times, the mark does not display.
```
from bqplot import *
import numpy as np
sc_x, sc_y = LinearScale(), LinearScale()
ax_x = bq.Axis(scale=sc_x)
ax_y = bq.Axis(scale=sc_y, orientation='vertical')
boxes = bq.Boxplot(x=[1,2,3], y=[[1,2,3], [1,2,3], [1,2,3]], scales={'x': sc_x, 'y': sc_y})
Figure(axes=[ax_x, ax_y], marks=[boxes])
```
**Assigning new data to the x/y attributes of a mark mark does not redraw the mark**. 
- Run the below code to observe this behavior:

`boxes.y = np.random.randn(3,3)`

**Missing features that are supported in other marks:**

- Tooltips are not supported
- Hover/click actions are not supported
- Selections are not supported

PR #760, opened 2018-12-13
Fix boxplot redraw issues

kaiayoung:
The boxplot mark was previously not responding to various state changes (new data, domain changing, etc), as not all of the selections were rebinding the data to the DOM elements. 
Fixes issue in #710, and some issues noted in #698.


kaiayoung:
Just noted a needed tweak in the outliers, will follow up with another commit soon.

SylvainCorlay:
backported to 0.11.x

PR #769, opened 2019-01-10
Upgraded to d3 v5

JohanMabille:


JohanMabille:
@SylvainCorlay I think this one is ready. I'll fix the problems on master unrelated to d3 upgade in dedicated PRs.

js/src/bqplot.less
```diff
@@ -192,15 +195,18 @@
...
     .indsel {
         stroke: red;
     }
-    .brushintsel, .intsel {
-        fill-opacity: 0.2;
+    .brushintsel {
+        .selection {
+            fill: var(--bq-selector-fill);
+            stroke: none;
+        }
```
JohanMabille: Note for self: transfer fill-opacity from code here.

JohanMabille:
@SylvainCorlay I had to changed the shortcut for removing multi brush selectors because ctrl + alt + shift does not work with xfce4 (either alt or shift is not detected depending on the order used to press the keys). Now ctrl + shift + click removes the brushes.

The brush is similar to the one in master, however there is still a problem (on both master and this PR): the points selected by an inactive brush are rendered as non selected points. This will be fixed in a dedicated PR.

Issue #770, opened 2019-01-11
Error thrown when plotting line with ordinal x values


ChakriCherukuri:
```python
import bqplot.pyplot as plt

f = plt.figure()
plt.plot(list('ABC'), [1, 2, 3], 'ms-')
f
```
Running the above code throws the following error:
TraitError: Array type not supported for trait x of class <class 'bqplot.marks.Lines'>: expected a             array of kind in list ['b', 'i', 'u', 'f', 'M'] and got an array of type <U1 (kind U)

This has to do with 'array_supported_kinds()' which was added to x and y attributes of various marks. @maartenbreddels can you please check?

maartenbreddels:
Hi Chakri,

Indeed a bug, I'm working on a PR to test/fix this.

cheers,

Maarten

SylvainCorlay:
Fixed by #772.

PR #772, opened 2019-01-17
Ordinal fix/support for binary

maartenbreddels:
As noticed in #770, since the move to binary transfer, ordinal values are not supported.


maartenbreddels:
On both the frontend and backend this is now supported, I'll also have to check other Marks.

maartenbreddels:
Bars y values can also be 2d, but I even in the 0.11.x branch ordinal values for the y values do not seem to work so I think we also don't have to support it now in master. If agreed, we can consider #770 fixed with this.

PR #775, opened 2019-01-21
Silence brush events on programmatic changes

### LINKED ISSUE #771: BrushIntervalSelector brushing initialized to True
The `brushing` attribute seems to be initialized to True, but should be False to start with:

```python
from bqplot import LinearScale, Axis, Lines, Figure
from bqplot.interacts import BrushIntervalSelector
import numpy as np

# generate data
prices = np.cumsum(np.random.randn(150, 2).dot([[0.5, 0.4], [0.4, 1.0]]), axis=0) + 100
prices1 = prices[:,0]
prices2 = prices[:,1]
domain = list(range(len(prices1)))

# Build plot
x_scale = LinearScale()
y_scale = LinearScale()
line1 = Lines(x=domain, y=prices1, scales=dict(x=x_scale, y=y_scale))
brush = BrushIntervalSelector(scale=x_scale, marks=[line1])
x_ax = Axis(label='X', scale=x_scale)
y_ax = Axis(label='Y', scale=y_scale, orientation='vertical')
fig1 = Figure(marks=[line1], axes=[x_ax, y_ax], interaction=brush)

# display plot
fig1

```
Then
```
brush.brushing
```
gives True, but should give False.

Furthermore, setting the `.selected` value only displays an updated selection every other time, and it seems that the `.brushing` attribute toggles between True and False:

```
brush.selected = [10, 20] # setting the selected range only displays the new range every other time
```

CC @drbitboy who brought up this issue in Gitter.

rmenegaux:
Stop calling `brush_start/move/end` on programmatic brush changes, and just redraw the brush. I can't think of a use case where we would want those events to be triggered. 

Fixes #771

### DIFF
--- a/js/src/BrushSelector.js
+++ b/js/src/BrushSelector.js
@@ -85,10 +85,10 @@ var BaseBrushSelector = {
     },
 
     _update_brush: function() {
-        // Programmatically setting the brush does not redraw it. It is
-        // being redrawn below
+        // Redraw the brush
         this.brushsel = this.d3el.call(this.brush);
-        this.d3el.call(this.brush.event);
+        // Trigger brushstart-move-end events
+        // this.d3el.call(this.brush.event);
     },
 
     clear_brush: function() {


js/src/BrushSelector.js
```diff
@@ -85,10 +85,10 @@ var BaseBrushSelector = {
     },
 
     _update_brush: function() {
-        // Programmatically setting the brush does not redraw it. It is
-        // being redrawn below
+        // Redraw the brush
         this.brushsel = this.d3el.call(this.brush);
-        this.d3el.call(this.brush.event);
+        // Trigger brushstart-move-end events
+        // this.d3el.call(this.brush.event);
```
jasongrout: Should we just delete these two lines of code, if we're not using them anymore?

PR #778, opened 2019-01-29
fixed data file locations and Layout errors

ChakriCherukuri:
Fixed the following issues which are throwing errors:
1. path to data_files
2.  using Layout class


SylvainCorlay:
Thx

PR #781, opened 2019-02-05
examples: correct link for FlexLine.ipynb

DentonGentry:
Difference in capitalization results in a 404.


ChakriCherukuri:
Thanks @DentonGentry

PR #785, opened 2019-02-14
Tile map fixes/enhancements

ChakriCherukuri:
1. Cleaned up selections (clicking and selecting tiles) to take care of special characters in names
2. Added a bool option to freeze/unfreeze tooltips. This way user can decide when to freeze the tooltips or when to make them follow the mouse


ChakriCherukuri:
@kaiayoung

Issue #144, opened 2016-05-13
The `process_interactions` code is very similar across Marks 


rmenegaux:
Shouldn't we move most of it to `Mark.js` ?

PR #794, opened 2019-03-22
Refactor `process_interactions` into the `Mark` base class, and provide bugfixes for new d3.

agoose77:
This PR moves the `process_interactions` function into `Mark`, in order to reduce code duplication. By calling `Mark.process_interactions` from `__super__`, this does mean that some branches are executed twice - once for the base class, and then overridden in the child, but I think this is acceptable. Most of this overriding is caused by the child class hooking into the click event, which could be removed by using a method overload instead.

I introduced a small helper function in `Mark` which I didn't re-use it in the other mark classes, to test for `value !== null && value !== undefined`. Having spoken with @maartenbreddels, a longer term solution would be to use lodash, but I thought that beyond the scope of this PR.

As well as refactoring this code, there are also some bugfixes which prevented testing along the way. Although it's bad practice to mix goals in a PR, my opinion was that they were small changes and I'm pretty sure that they're broken in master. As part of this, I've re-used the runtime resolution of `d3.event` with `d3.getEvent`. I'm under the impression that webpack is causing this bug, but I'm not proficient enough with webpack to really know how to solve it.

### DIFF
--- a/js/src/Bars.js
+++ b/js/src/Bars.js
@@ -163,8 +163,6 @@ var Bars = mark.Mark.extend({
                 this.event_listeners.parent_clicked = this.reset_selection;
                 this.event_listeners.element_clicked = this.bar_click_handler;
             }
-        } else {
-            this.reset_click();
         }
 
     },
--- a/js/src/Bars.js
+++ b/js/src/Bars.js
@@ -153,16 +153,11 @@ var Bars = mark.Mark.extend({
         });
     },
 
-    process_interactions: function() {
-        Bars.__super__.process_interactions.apply(this);
-        const interactions = this.model.get("interactions");
-
-        if(interactions.click !== undefined &&
-          interactions.click !== null) {
-            if (interactions.click === "select") {
-                this.event_listeners.parent_clicked = this.reset_selection;
-                this.event_listeners.element_clicked = this.bar_click_handler;
-            }
+    process_click: function(interaction) {
+        Bars.__super__.process_click.apply(this, [interaction]);
+        if (interaction === "select") {
+            this.event_listeners.parent_clicked = this.reset_selection;
+            this.event_listeners.element_clicked = this.bar_click_handler;
         }
 
     },
--- a/js/src/Bars.js
+++ b/js/src/Bars.js
@@ -14,6 +14,9 @@
  */
 
 var d3 = Object.assign({}, require("d3-array"), require("d3-scale"), require("d3-selection-multi"));
+// Hack to fix problem with webpack providing multiple d3 objects
+d3.getEvent = function(){return require("d3-selection").event}.bind(this);
+
 var _ = require("underscore");
 var mark = require("./Mark");
 var utils = require("./utils");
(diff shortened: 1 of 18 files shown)


js/src/Bars.js
```diff
@@ -603,7 +566,7 @@ var Bars = mark.Mark.extend({
         elements = elements.filter(function(data, index) {
             return indices.indexOf(index) !== -1;
         });
-        elements.selectAll(".bar").style(style);
+        elements.selectAll(".bar").styles(style);
```
agoose77: One of many places where our change from V3 to V5 needs to be handled. 
I believe this change occurred in v4: https://stackoverflow.com/questions/41474452/style-method-does-not-take-object

js/src/Bars.js
```diff
@@ -677,10 +640,7 @@ var Bars = mark.Mark.extend({
                        ((selected.length === 0) ? null : selected),
                        {updated_view: this});
         this.touch();
-        if(!d3.event) {
```
agoose77: This shouldn't be necessary now, as `getEvent` should be valid within an event?

js/src/Bars.js
```diff
@@ -677,10 +640,7 @@ var Bars = mark.Mark.extend({
...
-            d3.event = window.event;
-        }
-        var e = d3.event;
+        var e = d3.getEvent();
```
agoose77: We can remove this in a separate cleaning pass, I imagine.

js/src/BarsModel.js
```diff
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 
-var d3 = Object.assign({}, require("d3-array"), require("d3-scale"));
+var d3 = Object.assign({}, require("d3-array"), require("d3-scale"), require("d3-scale-chromatic"));
```
agoose77: I can't pin down when it happened, but since v3.x I think d3.schemeCategory10 was split into d3-scale-chromatic.

js/src/Graph.js
```diff
@@ -366,36 +366,17 @@ var Graph = mark.Mark.extend({
     },
 
     process_interactions: function() {
-        var interactions = this.model.get("interactions");
```
agoose77: Most of this is now common behaviour in the superclass

js/src/Graph.js
```diff
@@ -366,36 +366,17 @@ var Graph = mark.Mark.extend({
...
-            } else {
-                this.reset_hover();
+        Graph.__super__.process_interactions.apply(this);
+        const interactions = this.model.get("interactions");
+
+        if(interactions.click !== undefined &&
+           interactions.click !== null) {
+            if (interactions.click == "select") {
+                this.event_listeners.parent_clicked = this.reset_selection;
+                this.event_listeners.element_clicked = this.click_handler;
```
agoose77: All of the marks which overload this interaction only use it to set a custom `click_handler`. We could generalise this just as we do for `reset_selection` to remove the need to overload at all in many cases.

js/src/Pie.js
```diff
@@ -303,7 +265,8 @@ var Pie = mark.Mark.extend({
             })
             .each(function(d) {
                 this._current = d;
-            });
+            }).on("click", function(d, i) {
```
agoose77: I've moved this here as a bugfix, and also for correctness. From my understanding we would only want to add the event handler to new elements.

JohanMabille:
> I'm  under the impression that webpack is causing this bug, but I'm not  proficient enough with webpack to really know how to solve it.

Indeed the problem of `d3.event`  resolution is due to webpack, and the `getEvent` workaround is the recommended way to solve it. I'm not aware of any other fix.

I'll have a look in detail at your PR on Monday

rmenegaux:
@agoose77 thanks for tackling this, we've been meaning to do it for a long time (cf issue #144).

`process_interactions` could also call separate functions in `Mark.js` (`process_click`, `process_hover`, `process_legend_hover`, `process_legend_click` ...), that way marks would only have to override the individual functions (and most, as you pointed out, would only touch `process_click`).

Otherwise I'm good to merge this, it really cleans the code up!

agoose77:
Hi @rmenegaux,
I've taken care of the changes as requested. Let me know what you think, and thanks for the feedback!

agoose77:
Awesome, thanks!

PR #795, opened 2019-03-22
Use dict to dispatch event handlers on Python side

agoose77:
At present, the Python widget event dispatching is implemented with a giant `if ..: elif ...:` block, which is both unperformant and more work to maintain. This moves everything into a dict defined in `__init__`.


martinRenou:
Thanks!

agoose77:
Haha, can't believe this is merged after so long! Thanks :)

martinRenou:
Ahah yeah sorry for the wait!

PR #796, opened 2019-03-22
Move `display_el_classes` from some `MarkModel` objects to `Mark` cou…

agoose77:
…nterparts.

This isn't a bug, but I can't see any reason for this being here from grepping the codebase; it seems this should really have been defined on the `Mark` objects.

### DIFF
--- a/js/src/Mark.ts
+++ b/js/src/Mark.ts
@@ -35,7 +35,7 @@ export abstract class Mark extends widgets.WidgetView {
         this.setElement(document.createElementNS(d3.namespaces.svg, "g"));
         this.d3el = d3.select(this.el);
         super.initialize.apply(this, arguments);
-    },
+    }
 
     render() {
         this.x_padding = 0;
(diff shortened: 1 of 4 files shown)


martinRenou:
@agoose77 Would you like to rebase your PR on current master?

agoose77:
@martinRenou sure thing, let me know how it looks!

martinRenou:
We moved the code to typescript so you also need to remove the line https://github.com/bloomberg/bqplot/pull/796/files#diff-a3c52227b747ea6d85fd36653e29b4f2R128 on `MarkModel` :)

martinRenou:
Something I don't get is: what is this `display_el_classes` for? If it was not on the right place on the first time why was it working?

agoose77:
It's used by the tooltip system to identify elements which trigger a tooltip. It shouldn't be looked up on the model at all (see https://github.com/bloomberg/bqplot/blob/bcc71ad052e19cd408e645a6a3db3dde2c5f8a1a/js/src/Mark.ts#L488)

Issue #788, opened 2019-02-20
Changing color doesn't seem to trigger redraw/update


wolfv:
When changing the color trait, it doesn't seem to immediately trigger a redraw.

If changing e.g. x/y coordinates after color, it redraws right away with the correct color. 

Otherwise, need to hover or wait.

wolfv:
When running the binder / scatter example, adding the following after the basic example:

```
scatt.color = [5] * 30
```

shows the bug.

maartenbreddels:
I guess you mean:
```
scatt.colors = ['red'] * len(scatt.x)
```
note that color and colors exist ;)

martinRenou:
This seems to be fixed by #798. Would you mind closing it @wolfv ?

PR #798, opened 2019-03-26
Fix scatter color updating

rmenegaux:
Removes extra call to color updating when changing `color` data in `Scatter`.
Fixes issue #788


wolfv:
Nice

ChakriCherukuri:
@rmenegaux I tested and it works fine. Thanks for fixing this as I needed it for my demos!

Issue #444, opened 2017-02-28
Bars stroke cant be removed


dmadeka:
The stroke is set in the CSS, where it should be None and set to default in Python

PR #802, opened 2019-04-09
add ability to disable bar chart stroke outline

mbektasbbg:
When bar stroke is set to None as `Bars(stroke=None)`, a black outline is still shown. This PR fixes the issue by making sure stroke style set to `"none"`


andymcarter:
Just came across this issue myself, thanks for the fix. Looks like this should also close https://github.com/bloomberg/bqplot/issues/444.

PR #804, opened 2019-04-14
Add `nodejs` as dependency

wkerzendorf:
I tried to install bqplot and @SylvainCorlay suggested that it needs to have `nodejs` installed. Hope that's a useful PR.


SylvainCorlay:
Thanks @tylere . Sorry for the late reply.

It may be more appropriate to mention that installing the *Jupyterlab* extension also requires having nodejs.

martinRenou:
As of JupyterLab 3, this is not needed anymore

PR #805, opened 2019-04-15
Migrate bqplot to Typescript

jasongrout:
This migrates to Typescript 3.4.


jasongrout:
@JohanMabille has checked most of the examples. The Square market map is apparently not working in master, and the market map interactions are also not working in master, so we ignored those bugs for now. Other things appear to work, so we're merging this.

PR #810, opened 2019-04-17
Remove SquareMarkerMap

martinRenou:
`SquareMarkerMap` does not work, and looking at the file history it seems like it did not work for a long time. So I simply remove it.


martinRenou:
If this one get merged, I think it should be after #805. In order to prevent conflicts

jasongrout:
I rebased this on master. Thanks!

PR #814, opened 2019-04-17
Move more files to use es2015 classes

jasongrout:
CC @JohanMabille @martinRenou


JohanMabille:
@jasongrout awesome!
It would be nice if you could merge this PR, I've started migrating the Lines hierarchy, but this requires updating MarkModel and Mark and ends up with errors in BarsModel (which should be fixed with your changes).

jasongrout:
If you've approved this, I'll merge it.

JohanMabille:
Looks good to me!
Refactoring the brush selector is something we wanted to do after upgrading d3 so any intermediate temporary solution that allows to move one is fine for me.

PR #815, opened 2019-04-17
Toolbar.js var->let

martinRenou:


jasongrout:
Many of these can probably be `const` instead, right?

PR #816, opened 2019-04-17
Fix TS build on master

martinRenou:


js/src/FastIntervalSelector.ts
```diff
@@ -146,11 +146,6 @@ export class FastIntervalSelector extends BaseXSelector {
         }, this);
     }
 
-    scale_changed() {
```
martinRenou: This one is never called

PR #831, opened 2019-04-18
Finish converting to es6 classes

jasongrout:
I think this finishes out the conversion to es6 classes.

PR #833, opened 2019-04-19
Es6

JohanMabille:
This needs a rebase when #831 is merged.


jasongrout:
Thanks!

I think there were some places where I changed `.styles` back to `.style`, including one or two places where I have an awkward iteration. Do you want to change those back now?

JohanMabille:
Indeed, that would be nice. I'm trying to fix the Selector for now, I'll check that after.

PR #834, opened 2019-04-19
Fixed drag in ScatterBase.ts

JohanMabille:


js/src/ScatterBase.ts
```diff
@@ -26,13 +26,17 @@ export abstract class ScatterBase extends mark.Mark {
     render() {
         const base_creation_promise = super.render();
 
+        var that = this;
+        // Warning: arrow functions actually breaks the drag; that should be fixed
+        // with generics from the latest definitions of d3-selectoin and d3-drag
```
jasongrout: "d3-selection" (transposed o and i)

js/src/ScatterBase.ts
```diff
@@ -26,13 +26,17 @@ export abstract class ScatterBase extends mark.Mark {
...
         this.drag_listener = d3.drag()
-          .subject((d: any) => {
-              return {x: this.x_scale.scale(d.x), y: this.y_scale.scale(d.y)};
+          .subject(function(d: any) {
+              return {x: that.x_scale.scale(d.x), y: that.y_scale.scale(d.y)};
           })
-          .on("start", (d, i) => { return this.drag_start(d, i, this); })
-          .on("drag", (d, i) => { return this.on_drag(d, i, this); })
-          .on("end", (d, i) => { return this.drag_ended(d, i, this); });
+          .on("start", function(d, i) { return that.drag_start(d, i, this); })
```
jasongrout: do you need to change the `'this` third argument back to `that`? Or is it really needing to pass in the `this` that d3 passes in?

JohanMabille: As far as I understand, d3 changes `this` to be the current selection and the functions `drag_start`, `on_drag` and `drag_ended` accept a `dragged_node` third argument (which should be the current selection in the context of drag), so I think we must pass `this` as the third argument.

jasongrout: ah, and that explains why arrow functions break things.

jasongrout: in that case, can you remove the comment that: 

> that should be fixed with generics from the latest definitions of d3-selection and d3-drag

since arrow functions will break this always?

js/src/ScatterBase.ts
```diff
@@ -26,13 +26,17 @@ export abstract class ScatterBase extends mark.Mark {
...
         const base_creation_promise = super.render();
 
+        var that = this;
+        // Warning: arrow functions actually breaks the drag; that should be fixed
+        // with generics from the latest definitions of d3-selection and d3-drag
+        // from @types
         this.drag_listener = d3.drag()
-          .subject((d: any) => {
-              return {x: this.x_scale.scale(d.x), y: this.y_scale.scale(d.y)};
+          .subject(function(d: any) {
```
jasongrout: an arrow function is actually okay here, right? However, +1 for keeping it `function` for consistency with the below calls.

JohanMabille: Indeed this one could be replaced with an arrow function.

jasongrout:
Thanks!

js/src/ScatterBase.ts
jasongrout: > since arrow functions will break *this* always?

Sorry for the unintended pun :)

PR #843, opened 2019-04-25
Convert the last js file to typescript

jasongrout:
Convert the last js file to typescript


jasongrout:
I took the celebration as a positive review, @martinRenou - let me know if you want me to back this out :).

PR #844, opened 2019-04-25
Typescript target es2015->es2017

jasongrout:
We figured in JupyterLab that any supported browser in the last 2 years or so now supports the es2017 emitted javascript (last one to support native async/await was Safari 10.1 in middle 2017).

### DIFF
--- a/js/tsconfig.json
+++ b/js/tsconfig.json
@@ -4,7 +4,6 @@
       // "declaration": true,
       "esModuleInterop": true,
       "incremental": true,
-      "lib": ["es2015", "dom"],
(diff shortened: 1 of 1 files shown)


maartenbreddels:
Does that include mobile?

maartenbreddels:
I mean, jupyterlab on mobile may not make a lot of sense, but widgets are more likely to end at mobile/tablet

jasongrout:
Mobile looks good too (about 2 years):

https://www.chromestatus.com/feature/5643236399906816
https://caniuse.com/#feat=async-functions

PR #845, opened 2019-04-25
Fix Pie chart

### LINKED ISSUE #841: Pie plot doesn't show up
Pie plot doesn't show up when using master, no JavaScript error is displayed

martinRenou:
This should fix #841


jasongrout:
Thanks!

PR #854, opened 2019-05-24
Fix testing and enable many tests

maartenbreddels:
In 8eb8b163abe9ee6306f6918067e2f36c1caef2ef the typescript target was moved to es2017. Somehow with my local chrome (74.0.3729.157 on osx) this caused `figure.mark_views` not to be undefined in the unittest. The only explaination I can think of is the rules for when promises are resolved have changed with es2017.

The changes from 8eb8b163abe9ee6306f6918067e2f36c1caef2ef did not enter into #661 (mistake with rebasing?) and never triggered failing karma tests. This was fixed in c3a075f33a3fac39e3347da4cd1ba7881e7902d1, however PR #853 did not fail on Travis, this may be explained by chrome headless being older and maybe having different rules for when Promises are resolved.

The first commit solves this by explicitly creating a promise that will be resolved when the mark_views is created, which is much more robust. After this fix, all the test could be enabled again.

### DIFF
--- a/js/src/Figure.ts
+++ b/js/src/Figure.ts
@@ -58,6 +58,12 @@ export class Figure extends widgets.DOMWidgetView {
         this.el.appendChild(this.renderer.domElement);
         this.el.appendChild(svg_interaction)
 
+        // For testing we need to know when the mark_views is created, the tests
+        // can wait for this promise.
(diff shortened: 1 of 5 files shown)


martinRenou:
Thanks a lot for enabling the tests :)

PR #857, opened 2019-05-29
Fix error message to mention bqplot

astrofrog:
This error confused me when I was building several packages in one go


maartenbreddels:
👍 
The failure is due to a timing issue that is solved in https://github.com/bloomberg/bqplot/pull/854 
This change is unrelated and can be merged directly.

Issue #858, opened 2019-05-30
Issue with bqplot 0.12.0a1 in Jupyter Lab


astrofrog:
I'm trying to make use of the WebGL scatter in the latest alpha release of bqplot. I set up a clean environment with:

```
$ python3.7 -m venv ~/python/test-bqplot-alpha
$ source ~/python/test-bqplot-alpha/bin/activate
$ pip install jupyterlab bqplot --pre
$ jupyter labextension install @jupyter-widgets/jupyterlab-manager bqplot@0.5.0-alpha.0
```

Then started up Jupyter Lab and ran:

```python
import numpy as np
from bqplot import Figure, LinearScale, Axis, ColorScale, ScatterGL
    
scale_x = LinearScale(min=0, max=1)
scale_y = LinearScale(min=0, max=1)
scales = {'x': scale_x,
          'y': scale_y}
axis_x = Axis(scale=scale_x, label='x')
axis_y = Axis(scale=scale_y, label='y', orientation='vertical')

figure = Figure(scales=scales, axes=[axis_x, axis_y])

scales_image = {'x': scale_x,
                'y': scale_y,
                'image': ColorScale(min=0, max=1)}

scatter = ScatterGL(x=np.random.random((1000)), y=np.random.random((1000)), scales=scales_image)

figure.marks = (scatter,)

figure
```

In Jupyter Lab, the axes appear but the plot is empty, with the following error:

```
three.module.js:16269 Uncaught (in promise) TypeError: string.replace is not a function
    at parseIncludes (three.module.js:16269)
    at new WebGLProgram (three.module.js:16596)
    at WebGLPrograms.acquireProgram (three.module.js:17041)
    at initMaterial (three.module.js:22621)
    at setProgram (three.module.js:22761)
    at WebGLRenderer.renderBufferDirect (three.module.js:21783)
    at renderObject (three.module.js:22544)
    at renderObjects (three.module.js:22514)
    at WebGLRenderer.render (three.module.js:22280)
    at ScatterGL.render_gl (ScatterGL.js:417)
parseIncludes @ three.module.js:16269
WebGLProgram @ three.module.js:16596
WebGLPrograms.acquireProgram @ three.module.js:17041
initMaterial @ three.module.js:22621
setProgram @ three.module.js:22761
WebGLRenderer.renderBufferDirect @ three.module.js:21783
renderObject @ three.module.js:22544
renderObjects @ three.module.js:22514
WebGLRenderer.render @ three.module.js:22280
[...]

PR #859, opened 2019-05-30
Fix WebGL plotting in Jupyter Lab

astrofrog:
This declares ``raw-loader`` as a runtime dependency since it is needed to load some of the shader code (specifically for the scales).

If we don't specify this, we are dependent on the default versions of raw-loader in notebook and lab which might be different - in particular, recent versions of raw-loader used by Jupyter Lab don't return a string when used with require() but instead return an object with a 'default' property, which causes issues in three.js.

Fixes https://github.com/bloomberg/bqplot/issues/858

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -63,7 +63,8 @@
     "popper.js": "^1.0.0",
     "three": "^0.91.0",
     "topojson": "^1.6.24",
-    "underscore": "^1.8.3"
+    "underscore": "^1.8.3",
+    "raw-loader": "^0.5.1"
   },
   "files": [
     "dist/",
--- a/js/package.json
+++ b/js/package.json
@@ -63,8 +63,7 @@
     "popper.js": "^1.0.0",
     "three": "^0.91.0",
     "topojson": "^1.6.24",
-    "underscore": "^1.8.3",
-    "raw-loader": "^0.5.1"
+    "underscore": "^1.8.3"
   },
   "files": [
     "dist/",
--- a/js/src/Figure.ts
+++ b/js/src/Figure.ts
@@ -22,7 +22,15 @@ import * as popperreference from './PopperReference';
 import popper from 'popper.js';
 import * as THREE from 'three';
 
-THREE.ShaderChunk['scales'] = require('raw-loader!../shaders/scales.glsl')
+const scales = require('raw-loader!../shaders/scales.glsl');
+
+// Recent versions of raw-loader return an object where the shader string can
+// be accessed via the 'default' property
+if (typeof scales === "string") {
+  THREE.ShaderChunk['scales'] = scales;
+} else {
+  THREE.ShaderChunk['scales'] = scales.default;
+}
 
 export class Figure extends widgets.DOMWidgetView {
 


jasongrout:
What issues are caused? Can we check if the import has a default attribute?

astrofrog:
@jasongrout in this statement:

```
THREE.ShaderChunk['scales'] = require('raw-loader!../shaders/scales.glsl')
```

old versions of raw-loader return a string, new versions return an object with a ``default`` property. I could just check here which is the case and adjust automatically, which would mean the raw-loader version doesn't matter. Would you prefer that fix? In any case, ``raw-loader`` should be listed as a runtime dependency though.

maartenbreddels:
> In any case, `raw-loader` should be listed as a runtime dependency though.

I don't think so, the loader is only needed when building the bundles right, or did something change with raw-loader 2? I don't think the raw-loader should end up in the bundle, or am I getting the concepts of runtime/dev dependency wrong?

astrofrog:
I may have used the wrong term, but I mean that raw-loader definitely needs to be mentioned in the ``dependencies`` section (not just ``devDependencies``) otherwise we are at the mercy of whatever version Jupyter Lab wants to use (which is what was causing the issue here).

The shader code is loaded using raw-loader at the time when Jupyter Lab is opened, not when it is built, hence why I mentioned run-time dependencies.

astrofrog:
I've updated the approach here, but I'm seeing the following error today when trying to install the Jupyter Lab extension:

```
> bqplot@0.5.0-alpha.0 build /home/tom/Code/bqplot/js
> tsc && webpack && lessc less/bqplot.less css/bqplot.css

src/Mark.ts:103:58 - error TS2345: Argument of type 'unknown' is not assignable to parameter of type 'WidgetModel'.

103             scale_promises[key] = that.create_child_view(model);
                                                             ~~~~~

src/MarketMap.ts:333:58 - error TS2345: Argument of type 'unknown' is not assignable to parameter of type 'WidgetModel'.
  Type '{}' is missing the following properties from type 'WidgetModel': defaults, isNew, initialize, comm_live, and 78 more.

333             scale_promises[key] = that.create_child_view(model);
                                                             ~~~~~


Found 2 errors.

```

This seems to be unrelated to the changes I made as it happens even on master. Any idea what that could be due to?

jasongrout:
The code changes in this PR look good to me (caveat: I haven't tested).

FYI, you could do `scales.default` instead of `scales['default']`, which is visually complex.

jasongrout:
> This seems to be unrelated to the changes I made as it happens even on master. Any idea what that could be due to?

Possibly due to the recent release of the jlab widget manager and associated packages. I'll check.

jasongrout:
> I'm seeing the following error today when trying to install the Jupyter Lab extension:

What commands are you using to do this? And is it on bqplot master? What is your jlab version?

astrofrog:
I have Jupyter Lab 1.0.0a5 and I am running ``jupyter labextension install .`` inside the js directory. I see this on bqplot master (as well as this branch).

astrofrog:
Weirdly it worked yesterday, so something happened in the last 24h...

jasongrout:
I published a new version of the jlab manager (0.41) yesterday that supports jlab 1a5. If you're using jlab 1a3, you'll need to use jlab manager 0.40.

astrofrog:
I just installed jupyterlab-manager 0.41 since I have Jupyter Lab 1.0.0a5 but it's still crashing during the build:

```
$ jupyter labextension list
JupyterLab v1.0.0a5
Known labextensions:
   app dir: /home/tom/python/debug-bqplot/share/jupyter/lab
        @jupyter-widgets/jupyterlab-manager v0.41.0  enabled  OK
```

jasongrout:
This works great for me:
```
conda create -y -n tmp-bqplot859 notebook 'tornado<6' ipywidgets
conda activate tmp-bqplot859
pip install --pre jupyterlab
pip install -ve /path/to/bqplot/repo
jupyter labextension install @jupyter-widgets/jupyterlab-manager
jupyter labextension install /path/to/bqplot/repo/js
```

Given the alpha was just released, and also a new version of the jlab widget manager to support the alpha, perhaps there are some old dependencies somewhere in your chain? I'd try cleaning things out and reinstalling.

astrofrog:
I ran into the same issue on another computer but indeed the failure is gone after setting up a new environment, so false alert.

astrofrog:
Just confirming that this PR works and fixes bqplot in Jupyter Lab

stonebig:
it would be nice to have a bqplot update with this for Jupterlab-1.0.0 (release today ?)

agoose77:
Any updates on this? Can I do anything to help move it along?

astrofrog:
I've rebased this and it looks like the CI is passing now. This is ready for a final review/merging.

maartenbreddels:
The breaking change is described here: https://github.com/webpack-contrib/raw-loader/pull/69

Basically, the advice is to use (for raw-loader >=2):
```
import str from '!!raw-loader!./string.txt'
```

If bqplot moves to raw-loader 2, we'll use version >= 2 during the normal bundle building and Jupyter lab's bundle building. I'll test that out now and see if that works.

maartenbreddels:
I think #881 is a cleaner solution, I'll also use that for ipyvolume.

martinRenou:
This PR can be closed, it has been fixed by #881

PR #860, opened 2019-06-04
Fixup interaction layer

SylvainCorlay:

PR #861, opened 2019-06-05
WIP: fix: events from the background were never captured because they were hidden behind the webgl canvas

maartenbreddels:


maartenbreddels:
Replaces #860 
To both catch events on the background and to have the background color/style visible behind the axes, I had to add a similar rectangle in the foreground figure that is fully transparent.

ChakriCherukuri:
I tested example notebooks and they all work fine. I noticed that the missing data cells in market map are still black. I think this was fixed by @martinRenou. I'll file a separate issue for this

PR #867, opened 2019-06-14
Fix MarketMap color when data is NaN

martinRenou:
This fixes your issue concerning the color in MarketMap @ChakriCherukuri.

I'm not sure it's the proper fix though, do you mind looking at it @maartenbreddels ?


maartenbreddels:
Looks good to me, I'm not sure we can test this with a unit test, otherwise, I give it green light!.

ChakriCherukuri:
tested and it works fine. Thanks @martinRenou

PR #869, opened 2019-06-19
Fix date serialization

### LINKED ISSUE #866: Interval Selectors with date scales are not returning datetimes on the python side
If I have an interval selector using a date scale, the 'selected' trait on the python side is returning two floats (which represent the int timestamp) instead of two date objects. 

Look at the code in the 'Fast Interval Selector' section of the Interaction Layer example notebook to highlight this issue

martinRenou:
This should fix #866.

The bug has been introduced in 815a2de887c8d4e54e963f422149b4c88ea48de9

### DIFF
--- a/js/src/utils.ts
+++ b/js/src/utils.ts
@@ -56,7 +56,8 @@ export function convert_dates(value) {
     if(isTypedArray(value))
         return value;
     const convert_to_date = function(x) {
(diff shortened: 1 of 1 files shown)


ChakriCherukuri:
tested this and it looks good. Thanks @martinRenou for fixing this

PR #870, opened 2019-06-19
Remove unused attributes in Selectors

martinRenou:
This removes an unused `_view_name` attribute on the `Selector` class and an `is_date` attribute on the `MultiSelector` class.

It also fixes some pep8 issues.


stonebig:
Isn't there an update neede to get bqplot working with  Jupyterlab-1.0.0-rc.1 ? I fail to make it work at the moment

PR #881, opened 2019-07-18
fix: Upgrade to raw-loader 2 (like jupyter lab)

maartenbreddels:
This wayt we can load the shaders the same way.  This will allow bqplot to use WebGL within Jupyter lab.

raw-loader <2 was using common-js, so you could simply do:
```
require('raw-loader!../shaders/scales.glsl')
```
To get the shader string.

Jupyter lab uses raw-loader `~2.0.0`, which uses [ES module style](https://github.com/webpack-contrib/raw-loader/pull/69), requiring you to do:
```
require('raw-loader!../shaders/scales.glsl').default
```

@astrofrog solved this by supporting both cases in #859. However, if we upgrade the devDependencies in bqplot to raw-loader 2, we can use the latter case for bundle building and jupyter lab building.

Replaces #859

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -46,7 +46,7 @@
     "less": "^3.8.1",
     "less-loader": "^4.1.0",
     "mocha": "^5.2.0",
-    "raw-loader": "^0.5.1",
+    "raw-loader": "~2.0.0",
     "rimraf": "^2.6.1",
     "sinon": "^7.3.2",
     "sinon-chai": "^3.3.0",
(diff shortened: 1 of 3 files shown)


martinRenou:
Is this ready to be merged?

martinRenou:
Doesn't this fix break compatibility with the classic Notebook? Shaders don't seem to compile anymore in the classic Notebook with this fix

maartenbreddels:
did you run npm install again to really install the new webpack?

martinRenou:
I guess, because I could get it to work with JupyterLab, but I will check again

martinRenou:
Works perfectly for me! I must have had a broken environment back then.

@SylvainCorlay would it be ok to merge this fix?

PR #890, opened 2019-08-13
Revert #786

jasongrout:
In JupyterLab, it appears that #786 is causing a GridHeatMap to show a blank plot. It seems that possibly what is happening is that it can't calculate an accurate width, so the plot ends up being something like 5 pixels wide (it has a min width of 125px, and 60px*2 is subtracted off for margins, so the plot mark gets the remaining 5px, it seems).

For now, reverting this until we can figure out the root of the issue. Perhaps there is a race condition in measuring an accurate width?

```python
from __future__ import print_function
import numpy as np
from bqplot import *

np.random.seed(0)
data = np.random.randn(10, 10)

col_sc = ColorScale()
grid_map = GridHeatMap(color=data, scales={'color': col_sc})
Figure(marks=[grid_map], padding_y=0.0, min_aspect_ratio=0.01)
```
gives
Screen Shot 2019-08-13 at 11 30 04 AM

### DIFF
--- a/bqplot/marks.py
+++ b/bqplot/marks.py
@@ -1451,10 +1451,6 @@ class GridHeatMap(Mark):
     anchor_style: dict (default: {'fill': 'white', 'stroke': 'blue'})
         Controls the style for the element which serves as the anchor during
         selection.
-    display_format: string (default: None)
-        format for displaying values. If None, then values are not displayed
-    font_style: dict
-        CSS style for the text of each cell
 
     Data Attributes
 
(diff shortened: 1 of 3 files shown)


jasongrout:
@ChakriCherukuri - I'm seeing this especially in jlab, only sporadically in classic notebook.

For JLab, install from source with:

```sh
jupyter labextension install @jupyter-widgets/jupyterlab-manager
jupyter labextension install ./js
```
then start jlab with
```
jupyter lab --watch
```
which should rebuild lab anytime bqplot changes.

martinRenou:
Closing as outdated

PR #892, opened 2019-08-14
Fix grid heatmaps

jasongrout:
This is an alternate hacky "fix" (workaround) for #890 that preserves backwards compatibility.

### DIFF
--- a/js/src/GridHeatMap.js
+++ b/js/src/GridHeatMap.js
@@ -395,6 +395,12 @@ var GridHeatMap = mark.Mark.extend({
     },
 
     draw: function() {
+        // This is a total kludge to try to get around an apparent race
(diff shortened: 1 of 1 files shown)


jasongrout:
In debugging, it seems that this is appropriately updating DOM elements when the scale gets sized for the final figure size:

https://github.com/bloomberg/bqplot/blob/0a4ec84752152b6c671a53eeac9f1231b7638ddf/js/src/GridHeatMap.js#L441-L444

However, this does *not* seem to be updating the DOM, and we're not sure why:

https://github.com/bloomberg/bqplot/blob/0a4ec84752152b6c671a53eeac9f1231b7638ddf/js/src/GridHeatMap.js#L455-L463

and that seems to be the root of the issue here. Someone should fix this :).

PR #903, opened 2019-09-05
Adds a precomputed statistics option to the boxplot

cdknorow:
**Describe your changes**
Adds a precomputed_statistics option to the bqplot graph which allows a user to pass 
the quantiles and outliers as the y variable [q0, q25, q50 q75, q100, outlier1, outlier2, ... outlierN] 
and set the precomputed_statistics flag.

**Testing performed**
Still working on testing this, need some feedback on the implementation.

**Additional context**

### DIFF
--- a/bqplot/marks.py
+++ b/bqplot/marks.py
@@ -904,6 +904,8 @@ class Boxplot(Mark):
         If set to None, box_with is auto calculated
     auto_detect_outliers: bool (default: True)
         Flag to toggle outlier auto-detection
+    precomputed_statistics: bool (default: True)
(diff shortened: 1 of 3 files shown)


cdknorow:
Looking for some feedback on the api change. Does this make sense to you guys or should it be implemented in a different way?

maartenbreddels:
I think this is a very good idea and should be a general pattern adopted in bqplot. In cases where the data is too large, you want this to be computed at the kernel.

cdknorow:
@martinRenou I put this up a while ago. Not sure who is the correct person to review.

PR #914, opened 2019-09-20
Fix GridHeatMap color update

### LINKED ISSUE #913: GridHeatMap does not add/remove rows or columns when new data is assigned
When assigning a new matrix to the `color` parameter, the grid is not reshaped to conform to the new data:

```
import numpy as np
from bqplot import *
grid_map = GridHeatMap(color=np.random.randn(5, 5),
                       scales={'color': ColorScale()})
Figure(marks=[grid_map])
```

Run the below in a new cell, and you can observe that no new rows or columns are created, though the values are updated for the visible ones:

`grid_map.color = np.random.randn(10, 10)`

martinRenou:
Should fix #913

### DIFF
--- a/bqplot/marks.py
+++ b/bqplot/marks.py
@@ -49,6 +49,7 @@ from traitlets import (Int, Unicode, List, Enum, Dict, Bool, Float,
 from traittypes import Array
 
 from numpy import histogram
+import numpy as np
 
 from .scales import Scale, OrdinalScale, LinearScale
 from .traits import (Date, array_serialization,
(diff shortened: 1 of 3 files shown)


SylvainCorlay:
Can you comment a bit about the logic here?

martinRenou:
Sure. Now `row` and `column` attributes are None by default. And computed using a JavaScript range on the fly when the data changes. 

Also the PR contains a fix on the redraw using the d3 `join` feature for updating already existing rects and adding new ones automatically.

SylvainCorlay:
@ChakriCherukuri do you know if there is a case where we would want to set `row` or `columns` to a different value than the natural one from the data? If not, we may as well remove the trait altogether for 0.12.

Other than this question, this looks good to me.

bqplot/marks.py
```diff
@@ -1519,12 +1519,8 @@ class GridHeatMap(Mark):
     font_style = Dict().tag(sync=True)
 
     def __init__(self, **kwargs):
-        data = kwargs['color']
-        kwargs.setdefault('row', range(data.shape[0]))
-        kwargs.setdefault('column', range(data.shape[1]))
```
SylvainCorlay: should this have been `range(data.shape[0] + 1)` and `range(data.shape[1] + 1)` respectively?

js/src/GridHeatMapModel.ts
```diff
@@ -57,22 +58,24 @@ export class GridHeatMapModel extends MarkModel {
...
-        const that = this;
         this.colors = this.get("color");
-        this.rows = this.get("row");
-        this.columns = this.get("column");
 
+        const num_rows = this.colors.length;
         const num_cols = this.colors[0].length;
+
+        this.rows = this.get("row") || _.range(num_rows);
+        this.columns = this.get("column") || _.range(num_cols);
```
SylvainCorlay: same here, should we use `num_rows + 1` and `num_cols + 1`?

martinRenou: As per in-person discussion. We found that `_.range(num_rows)` is a valid default value because it uses an Ordinal scale by default and rows are centered on ticks (0, 1, 2, 3...), and in this case values in `row` are the values on which rows are centered.

Issue #856, opened 2019-05-24
Issue when showing multiple plots using WebGL


astrofrog:
I've run into an issue which manifests itself by the fact that when showing multiple figures that use WebGL (e.g. ScatterGL), one or more of the figures ends up empty. The following notebook reproduces this issue for me consistently in Firefox (but not Chrome):

https://gist.github.com/astrofrog/5434c4759bed440f98e2b84cbc6ab490

The issue is best demonstrated by selecting 'Restart Kernel and Run all cells'. If you do this repeatedly, every time (or almost) the first plot is empty:

![Screenshot from 2019-05-24 16-32-49](https://user-images.githubusercontent.com/314716/58339486-9c31e380-7e41-11e9-98d9-927f0c1a3c0d.png)

The console shows:

```
Error: WebGL warning: <SetDimensions>: Exceeded 16 live WebGL contexts for this principal, losing the least recently used one. three.module.js:21296
Error: WebGL warning: loseContext: Context is already lost. three.module.js:21296
Error: WebGL warning: <SetDimensions>: Exceeded 16 live WebGL contexts for this principal, losing the least recently used one. three.module.js:21296
Error: WebGL warning: loseContext: Context is already lost. three.module.js:21296
THREE.WebGLRenderer 91 three.module.js:21158
Error: WebGL warning: <SetDimensions>: Exceeded 16 live WebGL contexts for this principal, losing the least recently used one. three.module.js:21296
Error: WebGL warning: loseContext: Context is already lost.
```

martinRenou:
I remember we discussed this with @maartenbreddels, we should consider not creating a WebGL context when it is not needed. That would delay this kind of issues.

Anyway, you would still have this issue if you create more than 16 ScatterGL plots, and I don't think we'll be able to go around it.

maartenbreddels:
Yes, I was waiting for some PR's to be merged before working on this, but there are two issues here:
 * we shouldn't create the canvas when there are no gl marks
 * when we lose the context, we should make a screenshot of the canvas, and render that instead, similar to what pythreejs does (actually, it should be a library!)

martinRenou:
> when we lose the context, we should make a screenshot of the canvas, and render that instead, similar to what pythreejs does

Is it really worth it? The plot won't respond to any interaction anymore

maartenbreddels:
But it will go back to a WebGL canvas when that is needed, just turns static momentarily.

PR #927, opened 2019-09-26
Only create WebGL context when needed

### LINKED ISSUE #924: Rare issue: mark_views.views undefined when using WebGL marks
This sometimes happens when a webgl mark is used:
```
Figure.js:839 Uncaught TypeError: Cannot read property 'views' of undefined
    at e.Figure.render_gl (Figure.js:839)
    at e.Figure._update_gl (Figure.js:833)
    at Figure.js:829
```

Which means `this.mark_views.views` is undefined. This may be a timing issue with promises.

martinRenou:
Only create WebGL context when needed

The WebGL context is now lazily created. This should postpone this issue: https://github.com/bloomberg/bqplot/issues/856. Because it will then happen only if there are more than ~18 plots that actually need a WebGL context.

The logic is the following:
- if there is no mark that needs a WebGL context, the context is not created
- if a mark needing a WebGL context is added, we create the context
- we never remove the created WebGL context

This PR should also fix #924

### DIFF
--- a/js/src/Figure.ts
+++ b/js/src/Figure.ts
@@ -97,7 +97,7 @@ export class Figure extends widgets.DOMWidgetView {
         return return_value;
     }
 
-    render () {
+    render () : Promise<any> {
         let min_width = this.model.get("layout").get("min_width");
         let min_height = this.model.get("layout").get("min_height");
         if(typeof min_width === "string" && min_width.endsWith('px')) {
@@ -227,9 +227,9 @@ export class Figure extends widgets.DOMWidgetView {
 
         const figure_scale_promise = this.create_figure_scales();
 
-        figure_scale_promise.then(() => {
+        return figure_scale_promise.then(() => {
             this.mark_views = new widgets.ViewList(this.add_mark, this.remove_mark, this);
-            this.mark_views.update(this.model.get("marks")).then((views) => {
+            const mark_views_updated = this.mark_views.update(this.model.get("marks")).then((views) => {
                 this.replace_dummy_nodes(views);
                 this.update_marks(views);
                 this.update_legend();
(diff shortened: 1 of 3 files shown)


maartenbreddels:
should we add unit tests for this?

martinRenou:
Yes, will do :)

martinRenou:
@maartenbreddels tests added. Would you mind reviewing the PR?

js/src/Figure.ts
```diff
@@ -922,13 +929,23 @@ export class Figure extends widgets.DOMWidgetView {
...
+    render_gl() : Promise<void> {
+        // Nothing to render using a WebGL context
+        if (!this.renderer) {
+            return Promise.resolve();
+        }
+
+        // Marks are not created yet, maybe next frame?
+        if(this.mark_views === undefined) {
+            return Promise.resolve().then(() => {
+                this.update_gl();
```
maartenbreddels: I think this will not resolve after the update_gl really did its work

js/src/Figure.ts
```diff
@@ -922,13 +929,23 @@ export class Figure extends widgets.DOMWidgetView {
     }
 
     _update_gl() {
-        this.render_gl()
         this._update_requested = false;
+        this.render_gl();
```
maartenbreddels: should this maybe change to?
```suggestion
        await this.render_gl();
        this._update_requested = false;
```

js/src/Figure.ts
martinRenou: I actually think it does resolve the issue... Because at this point in the code `_update_requested` is false. So we actually call `requestAnimationFrame(() => this._update_gl());` inside `update_gl`, which was not the case before.

js/src/Figure.ts
martinRenou: `this._update_requested = false;` must be before the `render_gl` call, if we want to fix the issue

js/src/Figure.ts
martinRenou: > I think this will not resolve

Oh I thought you were talking about resolving the issue. You mean resolve the Promise?

martinRenou:
@maartenbreddels what if we simply do:

```javascript
render_gl() : Promise<void> {
    // Nothing to render using a WebGL context
    if (!this.renderer || this.mark_views === undefined) {
        return Promise.resolve();
    }

     [....]
}
```
in the `render_gl` method.

If there are no mark_views there is nothing to render anyway...

martinRenou:
@SylvainCorlay Would it be ok to merge this PR? It fixes a race condition that sometimes prevents the figure to be displayed.

PR #929, opened 2019-09-30
0.11.x Scales models fix

martinRenou:
cc. @kaiayoung


martinRenou:
Not sure the CI failure has to do with my PR, maybe we should remove the Python 3.3 test from the CI?

PR #941, opened 2019-10-01
Fixing unwanted scrollbar when BQPlot figures are wrapped in a VBox

ibdafna:
**Describe your changes**
Added a margin parameter to the bqplot class.

**Testing performed**
Verified rendered plots wrapped in VBox do not have a horizontal scrollbar.

**Additional context**
Nothing to add.


js/less/bqplot.less
```diff
@@ -109,6 +109,7 @@
     display: flex;
     width: 640px;
     height: 480px;
+    margin: 0%
```
SylvainCorlay: `margin: 0;`  should be sufficient.

ibdafna: Thanks!

SylvainCorlay:
Closing this one since we now have the `width: auto` in the 0.12.x by default.

However, we should recommend not using 100%.

PR #952, opened 2019-10-16
Auto width for bqplot figures

SylvainCorlay:
This PR removes the natural width of `bqplot`, which was `640px` and replaces it with `auto`.

This will make bqplot figures stretch horizontally in flexbox layouts, such as output areas and nested vboxes. People who were trying to obtain that behavior were generally setting `width` to `100%` causing the figure to overflow its container because of margins, and scrollbars to be displayed. With this change, this should not be required anymore.

Before:

bqplot-normal-width

After:

bqplot-full-width

cc @kaiayoung @cquah @maartenbreddels @martinRenou @jasongrout

### DIFF
--- a/js/less/bqplot.less
+++ b/js/less/bqplot.less
@@ -107,7 +107,7 @@
 .bqplot {
     box-sizing: border-box;
     display: flex;
-    width: 640px;
+    width: auto;
     height: 480px;
 }
 .bqplot > .svg-background {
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
This is the main thing to merge (or not) before 0.12.0 is tagged.

maartenbreddels:
I am 👍 on this, for 0.12, also in voila it looks good if it uses the full width of the page. 

Also, I think that now voila becomes more common this responsive behavior is better default behavior.

SylvainCorlay:
Merging following @kaiayoung's approval and the green light on gitter.

PR #964, opened 2019-11-05
Fix Axis size issues

### LINKED ISSUE #962: Graph color and scale has changed since release of v0.12
**Describe the bug**
Since the release of v0.12, I'm experiencing problems in the way graph of plots appear:
1. The color of grid which used to be kinda greyish-blue with white grid lines, is now plane white with grey grid lines.
2. The x-axis scale has also increased making graphs look horizontally stretched.

**To Reproduce**
By just plotting the basic line chart using pyplot API with default configurations, it can be observed on the 2 versions:
1. `bqplot v0.11.8`
image

2. `bqplot v0.12.0`
Plot for exact same code
image

Also, in v0.12.0, some of the grid lines go unexpectedly missing for line charts like this:
image

**Environment**
 - Operating System and Version: Ubuntu 16.04
 - I am using a conda environment - in which everything is same just upgraded bqplot from v0.11.8 to v0.12.0

**Additional context**
I am not aware if this has been explicitly changed in this new version on bqplot (v0.12.0) since I couldn't find any changelog.

martinRenou:
Will hopefully fix #962

### DIFF
--- a/js/src/Axis.ts
+++ b/js/src/Axis.ts
@@ -39,9 +39,6 @@ export class Axis extends WidgetView {
     render() {
         this.d3el.style("display", this.model.get("visible") ? "inline" : "none");
         this.parent = this.options.parent;
-        this.margin = this.parent.margin;
-        this.height = this.parent.height - (this.margin.top + this.margin.bottom);
-        this.width = this.parent.width - (this.margin.left + this.margin.right);
 
         const scale_promise = this.set_scale_promise(this.model.get("scale"));
         const offset_promise = this.get_offset_promise();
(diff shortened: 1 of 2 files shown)


martinRenou:
The CI failure is not related

js/src/Figure.ts
```diff
@@ -69,7 +69,7 @@ export class Figure extends widgets.DOMWidgetView {
         if (width_undefined && height_undefined) {
             // Same as the defaults in bqplot.less
             suggested_height = 480;
-            suggested_width = 640;
+            suggested_width = this.el.clientWidth;
```
maartenbreddels: This does not work in render, since the DOM is not attached (it will give 0). I think we really should have some predefined width, in case the DOM is not attached (offscreen rendering).

martinRenou: Yes I realised that later... That is why I've put the PR in WIP. I will revert this change.

martinRenou:
@maartenbreddels @SylvainCorlay I suppose this is kind of critical. We might want to release `0.12.1` quickly with it. What do you think?

SylvainCorlay:
Thanks. Will make a 0.12.1 including this change as soon as possible.

Issue #967, opened 2019-11-11
DateScale is inaccurate on scale of hours


joders:
**Describe the bug**
Feeding time data into plots that is on the scale of less than a day this bug becomes visible. The plot shows the data displaced by the amount of the timezone offset used by the browser of the user. 

**To Reproduce**
```
import importlib
import bqplot
importlib.reload(bqplot)
from bqplot import *
xsc=DateScale()
ysc=LinearScale()
mark=Scatter(scales={'x':xsc,'y':ysc})
xax=Axis(scale=xsc)
yax=Axis(scale=ysc,orientation='vertical')
with mark.hold_sync():
    mark.x=np.array([
        "2019-01-01T00:00:00",
        "2019-01-02T00:00:00",
        "2019-01-03T00:00:00",
    ]).astype("datetime64[s]")
    mark.y=[0,1,0]
fig = Figure(marks=[mark],axes=(xax,yax))
fig
```

**Expected behavior**
I would expect the data points in the plot to be located at the dates that I'm feeding into the system.

**Screenshots**
![image](https://user-images.githubusercontent.com/12979604/68560175-d38d5e80-047a-11ea-94e8-3803aca4cd16.png)

**Environment:**
Reproduced this in Firefox and Chrome
`bqplot.__version__ == '0.12.0'`
Ubuntu 19.10

**Comment**
I already wrote a fix for the issue which I'm going to post in a few minutes/hours.

SylvainCorlay:
@joders closing as I merged your PR - which should be in 0.12.8.

joders:
nice. thanx

PR #968, opened 2019-11-11
fix: DateScale doesn't apply browser timezone for ticks anymore

joders:
*Issue number of the reported bug or feature request: #967 *

On the frontend dates are represented by timestamps (js native Date internally uses a timestamp). But, conversion of a mere date into a timestamp isn't really possible when it's not known in which timezone the date was recorded (2019-01-01 00:00 doesn't give enough information for conversion into a timestamp. 2019-01-01T00:00+02:00 does). As convention utc is used to generate a timestamp even though that timestamp will most likely not be the actual point in time when the date's event occurred. This is ok if the timestamp is converted back into a timezone-less date using utc.

In the frontend this conversion back was performed using the local timezone of the user's browser. The frontend uses javascript native conversion functions for dates (new Date(...)) and also d3's conversion functions (timeFormat, timeSecond, timeMinute, timeHour). Luckily, d3 caters for the case that timestamps are encoded using utc. The corresponding functions are utcFormat, utcSecond, utcMinute, utcHour.

**Testing performed**
I ran the tests in the "tests" directory:
=== 9 passed, 12 warnings in 0.10 seconds ===
I executed the example given in the bug and confirmed that the ticks now show the correct dates.

### DIFF
--- a/js/src/Axis.ts
+++ b/js/src/Axis.ts
@@ -243,7 +243,7 @@ export class Axis extends WidgetView {
         if(this.axis_scale.model.type === "date" ||
            this.axis_scale.model.type === "date_color_linear") {
             if(this.model.get("tick_format")) {
-                return d3.timeFormat(this.model.get("tick_format"));
+                return d3.utcFormat(this.model.get("tick_format"));
             } else {
                 return this.guess_tick_format();
             }
@@ -254,7 +254,7 @@ export class Axis extends WidgetView {
                 //check the instance of the elements in the domain and
                 //apply the format depending on that.
                 if(utils.is_valid_time_format(tick_format)) {
-                    return d3.timeFormat(tick_format);
+                    return d3.utcFormat(tick_format);
                 } else {
                     return d3.format(tick_format);
                 }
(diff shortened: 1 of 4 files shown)


martinRenou:
@SylvainCorlay I suppose this is ready to be merged. @joders could you just redo your commit with the `--signoff` option? `git commit --amend --signoff && git push -f origin datescale-timezone-fix` should do it :)

joders:
I signed it off

joders:
Do I still have to do something for this to pass?

SylvainCorlay:
@joders many many thanks for this. I merged the PR which fixes a number of issues we were having with date scales. This will be in 0.12.8.

joders:
Thank you a lot for creating bqplot :) 
It's been by a margin the most satisfying python package I used thus far (python power user) and I can't understand why so few people seem to know about it.

SylvainCorlay:
> It's been by a margin the most satisfying python package I used thus far (python power user) and I can't understand why so few people seem to know about it.

That's great to hear. We are definitely going to push harder on the project and communicate more as it is moving to be a community project!

acelere:
Yes! Thank you both! These date issues were my biggest problem with bqplot because I deal only with time data... I had to make all sorts of work arounds but now... Uhu! 0.12.8 can't wait!!!

acelere:
And yes, bqplot is AWESOME!

acelere:
I tested this with bqplot v0.12.9 and it is now working.

Issue #653, opened 2018-05-20
Capturing mouse click events


sergeyfridrikh:
After spending hours reading documentation and looking into examples of bqplot use cases I still can not answer a simple question: how to capture mouse position from a mouse click. Please help.

Also, if the bqplot team believes that the quote from documentation presented below (and there are plenty of those) helps users in any way they are missing something very important here.

Quote:

class bqplot.scales.OrdinalScale(**kwargs)
An ordinal scale.
A mapping from a discrete set of values to a numerical range.

DougRzz:
Hi,
If any user believes that the documentation is not clear, constructive suggestions/contributions will be welcomed.  :-)
I'm not sure about the mouse position. I don't think its been implimented.  You are free to submit an issue.

DougRzz:
I'm a mechanical engineer and not a software developer. So i understand what you are saying. I learnt bqplot last year using the examples in https://github.com/bloomberg/bqplot/tree/master/examples. The examples are really good. I believe there is a section where it describes how to change the size of the markers.  If you want to get creative with interactive plots (and you don't want to write JavaScript), bqplot is really good and it is worth learning. Hopefully the documentation will improve over time. Thanks for the feedback.

andymcarter:
My 2c from a engineer turned recent software developer:

A combination of the tutorial notebooks and the docs (http://bqplot.readthedocs.io/) was how I learnt, first copy and pasting the examples from notebooks and gradually requiring the greater detail in the docs, understanding which kwargs belong to Axes, Scales, Marks etc. Learning the library from the docs alone would be challenging for me.

How did you end up setting the marker size? The parameter name 'default_size' might be the cause of some confusion, as it is 'default' for the plot you are making, not a bqplot-wide default that requires setting.

Searching the docs for 'scatter marker size' leads straight to the list of kwargs for the Scatter mark, and 'default_size' is listed. Explaining how to implement this in an API call feels a little out of scope. I originally thought the docs to be a bit parsimonious, but in fact they explain the API concisely, and leave any implementation to the notebooks.

sergeyfridrikh:
@andymcarter  Exactly, I also use 'default_size' to set the marker size. Should it be this way? The name default_size becomes a misnomer if you can change the variable behind it and breeds confusion. Besides, why not give an example of how such a basic thing as setting marker size can be accomplished? 

Thanks for pointing me to the examples. They have been helpful, but definitely they are far from covering majority of typical use cases.

[...]

Issue #972, opened 2019-11-19
Coordinates for Background Clicks


joders:
**Problem**
Using a scatter plot it is impossible to get the coordinates of a click on the graph without adding a point. `on_background_click()` only notifies that a click happened but not where. `on_click()` handlers don't seem to be called ever. None of the other events nor `.observe()` provide the desired information.

**Desired Functionality**
It would be nice to have `.on_background_click_with_coordinates()`. Which should of course pass the coordinates to the handler.

**Alternatives**
One somewhat fishy workaround is to use `bqplot.Image.on_element_click()`. Image can be added to the same figure as the mark you want to capture coordinates for. Then clicks on that figure are captured, the handler registered with `.on_element_click()` is called and the desired information is passed to that handler.

joders:
This is related, if not a duplicate of #653

PR #973, opened 2019-11-19
Capturing background clicks with coordinates

joders:
*Issue number of the reported bug or feature request: #972

**Changes**
I added a default click event to scatter plots which is triggered whenever the click interactions on the mark is set to neither 'add', 'delete' nor 'select'. The mark now has a further handler registration function called `.on_background_click_with_coordinates()`.

This implementation might not be ideal as it doesn't tie in with the present background click handling and also it doesn't have its own click interaction on the mark.

**Testing performed**
I only manually observed the events emitted by the mark to confirm that the desired information is present.

### DIFF
--- a/bqplot/marks.py
+++ b/bqplot/marks.py
@@ -246,6 +246,7 @@ class Mark(Widget):
         self._legend_hover_handlers = CallbackDispatcher()
         self._element_click_handlers = CallbackDispatcher()
         self._bg_click_handlers = CallbackDispatcher()
+        self._bg_click_with_coordinates_handlers = CallbackDispatcher()
         self.on_msg(self._handle_custom_msgs)
 
     def on_hover(self, callback, remove=False):
@@ -266,6 +267,9 @@ class Mark(Widget):
     def on_background_click(self, callback, remove=False):
         self._bg_click_handlers.register_callback(callback, remove=remove)
 
+    def on_background_click_with_coordinates(self, callback, remove=False):
+        self._bg_click_with_coordinates_handlers.register_callback(callback, remove=remove)
+
     def _handle_custom_msgs(self, _, content, buffers=None):
         if content.get('event', '') == 'hover':
             self._hover_handlers(self, content)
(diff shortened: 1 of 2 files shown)


joders:
This also fixes #653

joders:
to test:

```
from bqplot import *

xs=LinearScale(min=0,max=1)
ys=LinearScale(min=0,max=1)
xa=Axis(scale=xs)
ya=Axis(scale=ys,orientation='vertical')
mark=Scatter(scales={'x':xs,'y':ys},interactions={'click':''})   #  <-- IMPORTANT
fig=Figure(marks=[mark],axes=[xa,ya])

def background_click_handler(mark, event):
    print(event['point'])
mark.on_background_click_with_coordinates(background_click_handler)

fig
```

martinRenou:
Thanks for the PR anyway! This is super useful :)

joders:
> Should we not enhance the current `on_background_click` with the coordinates instead of adding a new Python method?

I did have a look at that but as far as I can remember there is no easy way to do that (or it's impossible). On the other hand the `background_click_with_coordinates` could also serve the ordinary background click function though from what I remember this might have performance implications.

> Thanks for the PR anyway! This is super useful :)

Thanks :)

jfraymond:
This is a very useful addition.
I spent hours looking the the documentation for this feature.
It would be perfect it it was available for all types of plots, not only scatter plots.

maartenbreddels:
Hi joders,

thanks, I think this could be very useful, even apart from #1066 which listens to all mouse events / clicks. I think there are many cases where we'd like to capture clicks that are not on a mark. However, I don't think the place should be in a Mark since it does not make sense in the case of multiple marks, do you agree? I think the best place would be in Figure.ts, which is actually the originator of the events  (ctrl-f for bg_clicked in that file).

Regards,

aarten

joders:
Hello,
I currently don't understand the architecture well enough to answer this and forgot quite a few things in the last months. I thought you can have multiple axes in one direction (in which case outputting a single coordinate for that direction isn't enough). I would definitely agree if axes in one direction are always aligned. Maybe I'm confusing scales and axes. Also, I guess it would fulfill @jfraymond 's needs.
Cheers,
Jonathan

Mathemmagician:
Hello from the future,

Was anything like this ever merged? Especially interested in getting mouse coordinates for the bq.Map

maartenbreddels:
Hi, you could maybe take a look at https://github.com/glue-viz/bqplot-image-gl/pull/57 instead

Mathemmagician:
Not sure that would work since Map allows for various zoom ins. I think the approach above works for stationary widgets. Cool library though

PR #976, opened 2019-11-27
fix: screenshot CSS was missing CSS variables

maartenbreddels:
**Describe your changes**
From the CSS, I now extract the variables used, get the computed values, and define those css variables in the generated css.

**Testing performed**
Manually call fig.save_svg and test if the output matches the classical notebook. Maybe a good idea to also test in Lab, with dark mode.


maartenbreddels:
I reverted the background color to its old behaviour, now it is jupyter lab theme compatible.

However, document.body's background is not the same color as shown in the notebook, but this is unrelated to this PR.

I think this can be merged as is.

martinRenou:
@maartenbreddels Do you know why tests are failing on this one?

maartenbreddels:
Ah I did not notice. Yes, the screenshot is also being used during testing, so now it takes into account the CSS, some colors have changed, will fix it.

Issue #981, opened 2019-12-04
ColorAxis random position in Firefox


davidbrochart:
**Describe the bug**
The `ColorAxis` sometimes appears in the middle of the figure in Firefox.

**To Reproduce**
1. `jupyter notebook bqplot/examples/Scales/Color\ Scales.ipynb`
2. Execute cells 1 and 2.
3. Execute cell 2 again until you see the `ColorAxis` appear in the middle of the figure.

**Expected behavior**
In Chromium, the `ColorAxis` always appears on the right side, as expected.

**Screenshots**
![img](https://user-images.githubusercontent.com/4711805/70133687-e4e71680-1686-11ea-856c-3abad1a03cb6.png)

**Environment (please complete the following information):**
 - Ubuntu Linux 19.10
 - Browser (if applicable): Firefox

martinRenou:
This is super weird, it should have been fixed by https://github.com/bloomberg/bqplot/pull/964. I am wondering if you are executing the right JavaScript on the page..

martinRenou:
https://github.com/bloomberg/bqplot/pull/982 Should definitely fix this, as we get rid of this hardcoded size.

davidbrochart:
Indeed, #982 seems to fix it. Thanks!

PR #982, opened 2019-12-04
Refactor Figure sizing logic

martinRenou:
**Describe your changes**
Before this change, the Figure was rendered once with a default size of `640x480` before `this.el` was even attached to the DOM. And there was a resize when the element is actually attached to the page.

After the change, the Figure waits for `this.el` to be attached before doing any rendering, and it renders in the available space specified by the CSS rules (width: auto, height: 480 by default). This prevents doing an extra resize, and it prevents the figure to take a first hardcoded size.

### DIFF
--- a/js/src/Figure.ts
+++ b/js/src/Figure.ts
@@ -25,7 +25,15 @@ import { WidgetView } from '@jupyter-widgets/base';
 
 THREE.ShaderChunk['scales'] = require('raw-loader!../shaders/scales.glsl').default;
 
-export class Figure extends widgets.DOMWidgetView {
+interface IFigureSize
+{
+    width: number,
+    height: number
+};
+
+
+export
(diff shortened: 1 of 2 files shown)


maartenbreddels:
This does ignore the min/max aspect ratio btw.

martinRenou:
Yes, we need to discuss this :)

That is the exact reason for the WIP

maartenbreddels:
I think this PR should be explored, it will cause bqplot to be more performant. However, #985 fixes the real issue mentioned in #981.

martinRenou:
@maartenbreddels @SylvainCorlay I think this is good to go! It now respects the min/max aspect ratio and it is signed off :)

maartenbreddels:
LGTM, and works.

SylvainCorlay:
Reviewed together. I approve.

PR #985, opened 2019-12-06
fix: ColorScale can miss events because render does not return a Promise

maartenbreddels:
*fixes: #981*

**Describe your changes**
Render should return a promise, otherwise, the Figure will not wait before sending events, causing the ColorAxis to miss them, which in turn causes a layout issue.

**Testing performed**
This is a timing issue, which only happens when there are many browser tabs open, or other forms of delay. This cannot be unit tested as it. It is however not observed to occur with this change.


SylvainCorlay:
@maartenbreddels this appears to require a rebase.

Also, there is the new sign-off thing that we need to do in the last commit of a PR.

PR #993, opened 2019-12-18
performance: avoid calling relayout too often

maartenbreddels:
**Describe your changes**
Figure.relayout is being called very often (5x at initial display), and unconditionally will do the work (call requestAnimation). These calls can stack up when resizing the window as well, leading to bad performance. We now keep track if the relayout was already requested, and will only schedule a new one when the last one is finished (similar to how we do it with webgl and in ipyvolume).

**Testing performed**
Tests pass, and tested in notebook. Putting a log statement in relayoutImpl shows 5 prints before, 2 after (so I think we can still improve even more if we want).

### DIFF
--- a/js/src/Figure.ts
+++ b/js/src/Figure.ts
@@ -89,7 +89,7 @@ class Figure extends widgets.DOMWidgetView {
         });
     }
 
-    private renderImpl (args: any) {
+    private renderImpl () {
         const figureSize = this.getFigureSize();
         this.width = figureSize.width;
         this.height = figureSize.height;
@@ -253,11 +253,7 @@ class Figure extends widgets.DOMWidgetView {
 
             document.body.appendChild(this.tooltip_div.node());
             this.create_listeners();
-            if(args === undefined || args.add_to_dom_only !== true) {
-                //do not relayout if it is only being added to the DOM
-                //and not displayed.
-                this.relayout();
-            }
+
             // In the classic notebook, we should relayout the figure on
             // resize of the main window.
             window.addEventListener('resize', () => {
--- a/js/src/Figure.ts
+++ b/js/src/Figure.ts
@@ -551,7 +551,10 @@ class Figure extends widgets.DOMWidgetView {
         super.processPhosphorMessage.apply(this, arguments);
         switch (msg.type) {
         case 'resize':
-            this.relayout();
+            const figureSize = this.getFigureSize();
+            if ((this.width !== figureSize.width) || (this.height !== figureSize.height)) {
+                this.relayout();
+            }
             break;
         }
     }
--- a/js/src/Figure.ts
+++ b/js/src/Figure.ts
@@ -82,7 +82,11 @@ class Figure extends widgets.DOMWidgetView {
     }
 
     render () {
-        this.displayed.then(this.renderImpl.bind(this));
+        // we cannot use Promise.all here, since this.layoutPromise is resolved, and will be overwritten later on
+        this.displayed.then(() => {
+            // make sure we render after all layouts styles are set, since they can affect the size
+            this.layoutPromise.then(this.renderImpl.bind(this));
+        });
     }
 
     private renderImpl (args: any) {
--- a/js/src/Figure.ts
+++ b/js/src/Figure.ts
@@ -256,8 +256,10 @@ class Figure extends widgets.DOMWidgetView {
 
             // In the classic notebook, we should relayout the figure on
             // resize of the main window.
-            window.addEventListener('resize', () => {
-                this.relayout();
+            const relayoutMethod = this.relayout.bind(this);
+            window.addEventListener('resize', relayoutMethod);
+            this.once('remove', () => {
+                window.removeEventListener('resize', relayoutMethod);
             });
 
             return Promise.all([mark_views_updated, axis_views_updated]);
--- a/js/src/Figure.ts
+++ b/js/src/Figure.ts
@@ -554,11 +554,11 @@ class Figure extends widgets.DOMWidgetView {
     }
 
     relayout() {
-        const figureSize = this.getFigureSize();
-        this.width = figureSize.width;
-        this.height = figureSize.height;
-
-        window.requestAnimationFrame(() => {
+        const relayoutImpl = () => {
+            this.relayoutRequested = false; // reset relayout request
+            const figureSize = this.getFigureSize();
+            this.width = figureSize.width;
+            this.height = figureSize.height;
             // update ranges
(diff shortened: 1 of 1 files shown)


js/src/Figure.ts
```diff
@@ -975,6 +980,7 @@ class Figure extends widgets.DOMWidgetView {
     y_padding_arr: any;
 
     private _update_requested: boolean;
+    private _relayout_requested: boolean;
```
martinRenou: Maybe you need a default value of `false` here

martinRenou: And I guess we don't need any underscore prefix in TypeScript. It is already defined as private.

martinRenou: I would also be in favor of using camelCase from now on.

maartenbreddels:
> It is not really fixing the issue of calling `relayout` too much, but just working around it... Would it be easy to reduce the number of `relayout` calls?

Basically any call to requestAnimationFrame should be 'protected' by such a mechanism, since you can have 1 JS event loop cycle schedule something like 10 of these. It does not make sense to do 10 redraws/relayout calls in the next cycle.

martinRenou:
I understand that. But again, `relayout` is still called a lot and that is the root of the issue. This is only a workaround.

martinRenou:
We might want to track where those `relayout` calls come from and try to reduce them as much as possible. 10 `relayout` calls at figure creation seems to be **a lot**. There might be a logic issue here.

Anyway, I am fine with the workaround for now. Just saying we might want to fix the root cause of this issue later.

maartenbreddels:
Fair point, if would still be required when resizing the window, since you can get two resize events, which is out our control.

So we have two problems, 1) relayout is being called too often by use (we should avoid that) 2) relayout may be triggered externally, because of browser events that we should solve it with this PR.

So this solved 2, but not 1?

maartenbreddels:
https://github.com/jupyter-widgets/ipywidgets/issues/2605 is one of the issue we cannot avoid in bqplot. I'll see if I can identify where the 5 come from.

martinRenou:
I feel like we can avoid this extra relayout since this fix: https://github.com/bloomberg/bqplot/pull/982. I am opening a PR that removes this extra relayout call #994

maartenbreddels:
I think I understand the issue for 80%.

If figure is a child of a Box, it will get two resize events. Those two resize events collapse with the 1st commit since all these events seem to be handled in 1 JS event loop cycle. In another JS event loop cycle we responded to the 'after-attach' event.
However, since our parent class already responds to this by firing the displayed event (which resolved the displayed promise, and our renderImpl wait for this promise, we can remove this case.
https://github.com/jupyter-widgets/ipywidgets/blob/92d7d42c00a1b0d9ce921533acb08beefdea3eb2/packages/base/src/widget.ts#L955

We are now down to 1 relayout call on initial display 🎉 

I've also tested this in Jupyter Lab, and we get 9 (!!) resize events when we do this:
```python
widgets.VBox([widgets.VBox([widgets.VBox([widgets.VBox([fig])])])])
```
It seems every VBox causes 2 resize events in fig, and fig itself gets one.

I'll mention that in https://github.com/jupyter-widgets/ipywidgets/issues/2605 as well.

I think this PR is ready for review @martinRenou, let me know what you think of this.

js/src/Figure.ts
```diff
@@ -558,7 +557,8 @@ class Figure extends widgets.DOMWidgetView {
         this.width = figureSize.width;
         this.height = figureSize.height;
```
martinRenou: Actually those three lines could be in the `relayoutImpl` function

js/src/Figure.ts
```diff
@@ -975,6 +979,7 @@ class Figure extends widgets.DOMWidgetView {
     y_padding_arr: any;
 
     private _update_requested: boolean;
+    private relayoutRequested: boolean;
```
martinRenou:
```typescript
private relayoutRequested: boolean = false;
```
Otherwise I believe it defaults to `undefined` (which does not really change the behavior here but I guess we'd better be explicit).

martinRenou:
Kudos on removing the `after-attach` one!

martinRenou:
> We are now down to 1 relayout call on initial display tada

Do you know where this remaining `relayout` call comes from? It would be awesome if we could turn this 1 into a 0!

maartenbreddels:
I actually didn't think we had to reduce it to 0, but it's from renderImpl, we call it ourself, and you fix that in https://github.com/bloomberg/bqplot/pull/994 (my last comment was wrong).

But even if we get rid of that call, we would need to find a way to ignore those resize events. What we could do, is keep a boolean after the first real render call, since it does not make sense to 'relayout' when you didn't even render yet right?

martinRenou:
> But even if we get rid of that call, we would need to find a way to ignore those resize events. What we could do, is keep a boolean after the first real render call, since it does not make sense to 'relayout' when you didn't even render yet right?

I guess yes :) Let's do that. Maybe you want to apply the changes I made in #994 in your PR so that your PR clears all the `relayout` calls? (I am fine not having credits for that)

maartenbreddels:
Got it to 0, but we should test this carefully, I propose we merge this after the next release.

maartenbreddels:
Because we did not remove the resize handler, we got a lot of views that were detached, causing these error messages:
```
Error: <rect> attribute width: A negative value is not valid. ("-120")
```
These are gone now.

js/src/Figure.ts
```diff
@@ -256,8 +256,10 @@ class Figure extends widgets.DOMWidgetView {
 
             // In the classic notebook, we should relayout the figure on
             // resize of the main window.
-            window.addEventListener('resize', () => {
-                this.relayout();
+            const relayoutMethod = this.relayout.bind(this);
+            window.addEventListener('resize', relayoutMethod);
+            this.once('remove', () => {
```
martinRenou: You can also re-implement the `remove` method. `window.addEventListener('resize', this.relayout, this);` and `window.removeEventListener('resize', this.relayout, this);` when removing. We might want to remove other event listeners in this method.

maartenbreddels:
I think the last step into making resizing feel more fluent is debouncing. I think you should never react to 'resize' events by doing anything that is expensive (a simple plot takes >30ms to redo, leading to max 30 fps), it just makes the browser feel really laggy.

I propose we add a debouncer before the relayout, so that layout can only be called after 300ms of 'rest'.

js/src/Figure.ts
maartenbreddels: I prefer it this way, now the code is more local to each other.

PR #994, opened 2019-12-18
Remove unneeded extra relayout call

martinRenou:
I put this in WIP, as we might want to discuss this before merging too quickly.


maartenbreddels:
I have never seen this being called, so I'm pretty sure we can remove it.

martinRenou:
Replaced by #993

Issue #995, opened 2019-12-18
Hist issues


martinRenou:
There seem to be multiple issues in the `Hist` implementation:

- [ ] `Hist` does not respect the `selected_style` and `unselected_style` attributes (see #918, kindof related: #476) 
- [ ] `Hist` does not respect the `colors` attribute (see #965). It basically takes the first color in this list for the unselected bars, and the second color for the selected ones, which is really not like other marks behave
- [ ] `Hist` might benefit from having a `sample_color` attribute (counterpart of the `sample` attribute from which it computes the histogram, but for the color) and supporting a color scale
- [ ] `Hist` could inherit from `Bar` (or have the same `BarBase` base class), I understand that Hist have the exact same visual representation as `Bar`, only the way it retrieves the data is different
- [ ] `Hist` does not support 2D data, it could

martinRenou:
I'll see if I can tackle the 2 first issues today.

PR #998, opened 2019-12-18
Hist respect colors

martinRenou:
This will fix the two first tasks on #995


martinRenou:
This needs some tests

martinRenou:
Added tests!

PR #1003, opened 2019-12-30
DOC: Use nbsphinx

mgeier:
This has been attempted before:  #679.

**Describe your changes**

Switch from `jupyter_sphinx` to `nbsphinx`.

Change `usage.rst` to `usage.ipynb`.

**Testing performed**

Built the docs, looks good.


mgeier:
I've signed off, please review.

SylvainCorlay:
That looks good to me. I am fine with using nbsphinx instead of jupyter_sphinx - although both don't quite do the same thing generally.

SylvainCorlay:
@mgeier does nbsphinx run the notebook to generate the widget metadata or should we do this?

mgeier:
By default, and if there are no outputs stored in the notebook (as is the case in this PR), `nbsphinx` executes the notebook (with the help of `nbconvert`), which generates the widget data.

But you should probably not take my word for it, you should test it for yourself.

You might even want to add this to your CI (here's an example setup: https://github.com/spatialaudio/nbsphinx/blob/master/.circleci/config.yml).

mgeier:
Any news on this?

I've just re-based this PR for your merging convenience.

JohanMabille:
LGTM. ping @SylvainCorlay

maartenbreddels:
LGTM as wel!

PR #1007, opened 2020-01-13
Axis: Fix axis side validation

### LINKED ISSUE #956: Assigning the side attribute for the axis is not reliable
**To Reproduce**
```python
import bqplot as bq
axis = bq.Axis(scale = bq.LinearScale() ) 
axis.orientation = 'vertical'
axis.side = 'right'
```
But ```axis.side != 'right'``` unless ```axis.side ='right'``` is executed in another cell.

**Expected behavior**
axis.side == 'right'

**Screenshots**
image

**Environment (please complete the following information):**
 - Operating System and Version: Windows 10
 - Browser [e.g. Firefox, Safari] (if applicable): Chrome

**Suggestion:**
Remove the orientation attribute and only have the side attribute.  Orientation is superfluous if side is specified as either top, bottom, left or right.

martinRenou:
This fixes #956

### DIFF
--- a/bqplot/axes.py
+++ b/bqplot/axes.py
@@ -117,8 +117,7 @@ class Axis(BaseAxis):
     icon = 'fa-arrows'
     orientation = Enum(['horizontal', 'vertical'], default_value='horizontal')\
         .tag(sync=True)
-    side = Enum(['bottom', 'top', 'left', 'right'],
-                allow_none=True, default_value=None).tag(sync=True)
+    side = Enum(['bottom', 'top', 'left', 'right'], default_value="left").tag(sync=True)
     label = Unicode().tag(sync=True)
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
This looks ok to me but it might still be possible to engineer a case where subsequent settings of `side` and `orientation` result in an inconsistent state with a race condition, because the attributes can be used to modify each other and the transformations don't commute.

Merging as an initial improvement but we should move to another model.

martinRenou:
> This looks ok to me but it might still be possible to engineer a case where subsequent settings of side and orientation result in an inconsistent state with a race condition, because the attributes can be used to modify each other and the transformations don't commute.
Merging as an initial improvement but we should move to another model.

You are absolutely right, and I think this PR broke the axis actually (see https://github.com/bloomberg/bqplot/issues/1024).

Maybe we should revert this change for now, what do you think?

PR #1011, opened 2020-01-14
Scatter: default_opacities -> opacities

martinRenou:
Deprecate Scatter's `default_opacities` in favor of `opacities` (`default_colors` was deprecated in favor of `colors` in bqplot `0.8.4`).

Other marks had `opacities` instead of `default_opacities`, so this will make bqplot's API more homogeneous.

This is the first of several PRs for making `Mark`'s implementation more general, by adding common marks logic inside of it. For this purpose we need to make marks APIs more homogeneous.

### DIFF
--- a/bqplot/marks.py
+++ b/bqplot/marks.py
@@ -522,7 +522,7 @@ class _ScatterBase(Mark):
         'opacity': {'dimension': 'opacity'},
         'rotation': {'dimension': 'rotation'}
     }).tag(sync=True)
-    default_opacities = Array([1.0])\
+    opacities = Array([1.0])\
         .tag(sync=True, display_name='Opacities', **array_serialization)\
         .valid(array_squeeze, array_dimension_bounds(1, 1))
     hovered_style = Dict().tag(sync=True)
@@ -592,7 +592,7 @@ class Scatter(_ScatterBase):
         Stroke color of the marker
     stroke_width: Float (default: 1.5)
         Stroke width of the marker
-    default_opacities: list of floats (default: [1.0])
+    opacities: list of floats (default: [1.0])
         Default opacities of the markers. If the list is shorter than
         the number
         of points, the opacities are reused.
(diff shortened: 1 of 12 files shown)


martinRenou:
Small tip for the reviewer, use `?w=1` in the PR diff url in order to remove whitespaces changes: https://github.com/bloomberg/bqplot/pull/1011/files?w=1

js/src/Mark.ts
```diff
@@ -208,6 +208,28 @@ export abstract class Mark extends widgets.WidgetView {
         return this.colors[index % len];
     }
 
+    get_mark_color(data, index) {
```
martinRenou: Those methods were in `ScatterBase.ts`, they will be very useful for other marks as well, not only for `Scatter`, so let's move them to the base class

martinRenou:
@SylvainCorlay If you have time reviewing :)

SylvainCorlay:
This is a longstanding inconsistency in bqplot.

I would be happy to see this go through although it is a backward incompatible change.

martinRenou:
This is not backward incompatible as far as I understand. I added a `default_opacities` property that prints a deprecated warning for now.

SylvainCorlay:
Oh great!

martinRenou:
Any change we can merge this PR?

martinRenou:
Something went wrong when I wanted to make the rebase. I'll reopen this.

maartenbreddels:
Since it is a breaking change, we should wait for 0.13 I'm afraid 😢

martinRenou:
Why do you say it's a breaking change?

I think we agreed with Sylvain it was not:
> This is not backward incompatible as far as I understand. I added a default_opacities property that prints a deprecated warning for now.

martinRenou:
Rebased again, and I updated the tests and the Notebook examples.

@SylvainCorlay do you think it should be merged or should we actually wait for 0.13?

PR #1014, opened 2020-01-15
Remove BaseModel

martinRenou:


SylvainCorlay:
This is formally backward incompatible on the js side for people extending bqplot with new marks.

cc @maartenbreddels

maartenbreddels:
Nothing in https://github.com/glue-viz/bqplot-image-gl/ uses this.

PR #1018, opened 2020-01-16
GridHeatmap: Fix selection

### LINKED ISSUE #1013: GridHeatMap selection broken
**Describe the bug**
The selected attribute of gridheatmap is not correctly updated

**To Reproduce**
```python
import numpy as np
from bqplot import *

data = [[0, 1, 2], [3, 4, 5], [6, 7, 8]]

col_sc = ColorScale(scheme='RdYlGn')
grid_map = GridHeatMap(color=data, scales={'color': col_sc})

fig = Figure(marks=[grid_map], layout={'width': '400px', 'height': '500px'},
            padding_y=0, fig_margin={'bottom': 0, 'left': 0, 'right': 0, 'top': 0})

grid_map.unselected_style = {'opacity': 0.2}
grid_map.interactions = {'click': 'select'}
```
Now select some squares:
select

This will set the `selected` attr to something like: `array([[0, 0]])`, it will contain only one selected square (the first one that was selected), not all of them.

martinRenou:
This will fix #1013

### DIFF
--- a/js/src/GridHeatMap.ts
+++ b/js/src/GridHeatMap.ts
@@ -151,8 +151,8 @@ export class GridHeatMap extends Mark {
         const row = args.row_num;
         const column = args.column_num;
         const that = this;
-        let idx = this.model.get("selected") || [];
-        let selected = this._cell_nums_from_indices(Array.from(idx));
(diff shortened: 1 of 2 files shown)


martinRenou:
This needs a test

Issue #1023, opened 2020-01-20
Problem displaying bqplot within an ipywidget tabbed interface


mstanley103:
**Problem statement:**
I recently reinstalled Anaconda and hit a problem on code that had previously worked OK.  Specifically, I have a tabbed interface created using ipywidgets and am attempting to display bqplot line plots within individual tabs.  The problem I'm seeing is that only the currently selected tab displays the plot correctly.  If I select another tab (and plot), that plot is not displayed correctly until I re-display the entire form.  Then the first tab has the problem.  As I said, it's a fresh (from scratch) re-installation of anaconda, and I am using the Python 3.7.6 kernel.

**To reproduce the problem:**

In running the simplified notebook contained in the attached .zip file (Run All cells), you'll see the first plot displayed normally on Tab1:
![image](https://user-images.githubusercontent.com/985116/72690576-d344bb80-3ada-11ea-9554-091446ec1628.png)

Looking at Tab2, you SHOULD see the same plot, but you don't:
![image](https://user-images.githubusercontent.com/985116/72690587-f53e3e00-3ada-11ea-9502-c0ae6c553621.png)

Now if you re-execute the display(gui) command:
![image](https://user-images.githubusercontent.com/985116/72690596-07b87780-3adb-11ea-9f4c-0df3ee6b0c9c.png)

and now Tab1 has the problem:

![image](https://user-images.githubusercontent.com/985116/72690670-df7d4880-3adb-11ea-93de-d2394cf05c2a.png)

I've added a Tab3 to the sample testcase to show that a Matplotlib plot displayed within an Output shows OK.

**Expected behavior**
The plot should display correctly regardless of which tab is selected.

**Environment:**
 - Windows 10 Pro 64-bit
 - Google Chrome Version 79.0.3945.117
 - the attached conda.txt file shows the environment asdumped using the "conda env export" command.

**Attachments:**
[conda.txt](https://github.com/bloomberg/bqplot/files/4083926/conda.txt)
[debug_script1.zip](https://github.com/bloomberg/bqplot/files/4083927/debug_script1.zip)

martinRenou:
Thanks a lot for filling this issue. This has been fixed on master very recently https://github.com/bloomberg/bqplot/pull/982. It will be part of the next release.

mstanley103:
Fantastic. Thank you for the quick response.  I'll keep an eye out for that release.
Thanks for making such a great library available!

martinRenou:
[...]

PR #1029, opened 2020-01-20
Fix: Render/resize the plot only if it is visible

martinRenou:
This delays the rendering and the resize of the plot to when the plot is actually visible on the page.

It is using the [IntersectionObserver API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) for knowing if the plot is visible on the page or not

This fixes an issue when having bqplots in Tab widgets.

Fixes https://github.com/bloomberg/bqplot/issues/1023

This is related to https://github.com/jupyter-widgets/ipywidgets/issues/2605

### DIFF
--- a/js/src/Figure.ts
+++ b/js/src/Figure.ts
@@ -36,6 +36,8 @@ export
 class Figure extends widgets.DOMWidgetView {
 
     initialize() {
+        super.initialize.apply(this, arguments);
+
         // Internet Explorer does not support classList for svg elements
         this.el.classList.add("bqplot");
         this.el.classList.add("figure");
@@ -59,7 +61,18 @@ class Figure extends widgets.DOMWidgetView {
             this._initial_marks_created_resolve = resolve;
         });
 
-        super.initialize.apply(this, arguments);
+        this.debouncedRelayout = _.debounce(() => {
+            this.hasBeenShown.then(() => {
+                this.relayout();
+            });
+        }, 300);
+
+        this.hasBeenAttached = this.displayed;
+        this.hasBeenShown = new Promise((resolve, reject) => {
+            this.once('shown', () => {
+                resolve();
+            });
+        });
     }
 
     protected getFigureSize (): IFigureSize {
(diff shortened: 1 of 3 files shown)


martinRenou:
Opening this PR as draft. We really need to discuss this. 

All of this might need a change in ipywidgets: wait for the first `after-show` event to be triggered before resolving the `displayed` promise.

cc. @maartenbreddels

martinRenou:
This will fix #1023 by lazily rendering the plot.

lazy_load

martinRenou:
Unfortunately, it does not work well when doing a resize of the page

resize

martinRenou:
@maartenbreddels I have a working version using the intersection observer. Although it breaks tests for now.

martinRenou:
Tests are passing again by enforcing the visible status of the plot in tests.

martinRenou:
@maartenbreddels I'd be interested to know if that works with your tablet

jasongrout:
I've also opened a new issue about fixing the underlying issue in ipywidgets: https://github.com/jupyter-widgets/ipywidgets/issues/2762

jasongrout:
You could just listen for the `after-attach` and `after-show` Lumino events, right?

martinRenou:
It was actually the first version of this PR (see 0c56d3ba16b2c7d285f4dcf890954fa44c616253), but it was not working properly (see [this comment](https://github.com/bloomberg/bqplot/pull/1029#issuecomment-576661533)).

Also, the intersection Observer is a bit different from the `after-show`. It will trigger the `show` event only if the plot is actually visible on the screen (some pixels are visible). While the Lumino's `after-show` event is triggered even if the page is completely scrolled down and the plot is not visible.
So it also brings speed improvements on the resize: only resizing plots that are visible.

jasongrout:
>  So it also brings speed improvements on the resize: only resizing plots that are visible.

How does this handle the case where you scroll a visible bqplot plot off the screen, then resize the browser window, then scroll back to see the plot? Will it automatically resize as soon as it becomes visible?

jasongrout:
https://github.com/bloomberg/bqplot/pull/1047 is an alternative implementation that uses the phosphor after-show event. It's not as fancy, in that it doesn't use the browser visibility capabilities, but it is quite a bit simpler. Thoughts?

martinRenou:
> How does this handle the case where you scroll a visible bqplot plot off the screen, then resize the browser window, then scroll back to see the plot? Will it automatically resize as soon as it becomes visible?

Yes :)

martinRenou:
> #1047 is an alternative implementation that uses the phosphor after-show event. It's not as fancy, in that it doesn't use the browser visibility capabilities, but it is quite a bit simpler. Thoughts?

I agree. I tried this approach at first but could not make it work. I guess I forgot the `this.pWidget.isVisible` part that I did not know of.

Even though this PR brings more optimizations on the resize (resize plot only if it is visible), I would say simpler is better, we can close my PR.

martinRenou:
Closed in favor of #1047

PR #1030, opened 2020-01-21
Fix: Mark.selected should be a TypedArray / selection echo

maartenbreddels:
**Describe your changes**
The `Mark.selected` attribute should be a typed array. Although bqplot supports JSON/lists, they will be converted to a numpy array on the kernel side. This will then be seen as a change on the kernel and will be sent back to the frontend. This causes a lot of echoing, causing old selections to be visible in between.

**Testing performed**
ScatterBase and Hist are tested, the rest follows a similar pattern.

**Unrelated issue**
I also had to change the return value of OHCL, since TypeScript complained. I am not sure this is correct, since I don't understand this method. A review would be wise.

### DIFF
--- a/js/src/Bars.ts
+++ b/js/src/Bars.ts
@@ -220,7 +220,7 @@ export class Bars extends Mark {
         }
 
         const abs_diff = this.x_pixels.map(function (elem) { return Math.abs(elem - pixel); });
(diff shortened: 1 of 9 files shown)


martinRenou:
@SylvainCorlay Do you mind reviewing/merging this?

PR #1031, opened 2020-01-22
Remove unused function

martinRenou:
I suppose this is a leftover. It is not used anywhere.


SylvainCorlay:
Will merge after #1035 is in.

jdemeyer:
See #1054

PR #1033, opened 2020-01-23
Improve plot API for marker_str argument

### LINKED ISSUE #1027: marker_str does not appear to work on plot
In the documentation of pyplot.plot, it includes the marker_str param, which a user would expect to work like in matplotlib.  I couldn't find any change in behaviour by using this parameter.

**To Reproduce**
`

plt.figure(1)
plt.clear()
n = 100
plt.plot(np.linspace(0.0, 10.0, n), np.cumsum(np.random.randn(n)), marker_str ='g-o')
plt.show()

`

**Expected behavior**
the line to turn green

**Additional context**
I was able to change the colour using the colors argument, although the palette is particularly unattractive!

`

plt.figure(1)
plt.clear()
n = 100
plt.plot(np.linspace(0.0, 10.0, n), np.cumsum(np.random.randn(n)), colors=['green'])
plt.show()
`

martinRenou:
This will fix #1027

It allows passing `marker_str` as a keyword argument:
```python
# plot(y)
plt.plot(np.cumsum(np.random.randn(n)))

# plot(y, marker_str)
plt.plot(np.cumsum(np.random.randn(n)), marker_str='b--')
plt.plot(np.cumsum(np.random.randn(n)), 'r--')

# plot(x, y, marker_str)
plt.plot(np.linspace(0.0, 1000., n), np.cumsum(np.random.randn(n)), marker_str='g-o')
plt.plot(np.linspace(0.0, 10.0, n), np.cumsum(np.random.randn(n)), 'g-')
```

### DIFF
--- a/bqplot/pyplot.py
+++ b/bqplot/pyplot.py
@@ -671,12 +671,11 @@ def _infer_x_for_line(y):
 
 
 @_process_data('color')
-def plot(*args, **kwargs):
+def plot(x, y=None, marker_str=None, **kwargs):
     """Draw lines in the current context figure.
 
-    Signature: `plot(x, y, **kwargs)` or `plot(y, **kwargs)`, depending of the
-    length of the list of positional arguments. In the case where the `x` array
-    is not provided.
+    Signature: `plot(x, y, marker_str=None, **kwargs)` or `plot(y, marker_str=None, **kwargs)`, depending on the number of
+    positional arguments.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


ChakriCherukuri:
marker_str is not implemented as a keyword arg explicitly, to keep the pyplot api similar to matlab/matplotlib (For e.g. see [here](https://matplotlib.org/api/_as_gen/matplotlib.pyplot.plot.html))

I think we should retain the old behavior IMHO.

martinRenou:
Thanks for your comment @ChakriCherukuri. Closing this then.

Issue #780, opened 2019-02-01
IndexError on BrushSelector with OrdinalScale


kaiayoung:
Attempting to make a selection on a `Scatter` mark with an `OrdinalScale` results in an `IndexError`, see below for an example:

```
from bqplot import *
from bqplot.interacts import BrushSelector
sc_x = LinearScale()
sc_y = OrdinalScale()

scatt = Scatter(x=[1, 2, 3], y=[1, 2, 3], scales={'x': sc_x, 'y': sc_y})

Figure(marks=[scatt],
       interaction=BrushSelector(x_scale=sc_x, y_scale=sc_y, marks=[scatt]))
```

`IndexError: index 1 is out of bounds for axis 0 with size 1`

rmenegaux:
This is indeed a bug. For `OrdinalScale`, the `brush.selected_y` is a list of all the categories selected, which is not necessarily of length 2. The conversion from `selected_x/y` to `selected` expects them to be of length 2.
We could either:
- Deprecate `BrushSelector.selected`, because `selected_x` and `selected_y` can be of different lengths. Update the documentation to say that `BrushSelector.selected_x/y` and `BrushIntervalSelector.selected` are [min, max] arrays in the case of numerical scales, and [all elements] arrays in the case of categorical scales. 
- Force them to be length 2 arrays, by taking the first and last elements.

martinRenou:
This specific issue seems to have been fixed (using the second option @rmenegaux mentionned).

But there was still an error using `OrdinalScale` on master, which should be fixed by #1040

martinRenou:
![brush](https://user-images.githubusercontent.com/21197331/74520663-514c8480-4f18-11ea-956f-de616656f904.gif)

PR #1040, opened 2020-02-14
Fix error with `invert` not being a d3's Ordinal scale method

martinRenou:
Related to #780

### DIFF
--- a/js/src/ScatterBase.ts
+++ b/js/src/ScatterBase.ts
@@ -561,12 +561,12 @@ export abstract class ScatterBase extends Mark {
 
         if (!this.model.get("restrict_y")){
             const x = this.model.get('x').slice(); // copy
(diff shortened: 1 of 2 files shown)


js/src/ScatterBase.ts
```diff
@@ -663,8 +663,8 @@ export abstract class ScatterBase extends Mark {
         const yn = new y.constructor(y.length+1)
         xn.set(x)
         yn.set(y)
-        xn[x.length] = x_scale.scale.invert(curr_pos[0]);
-        yn[y.length] = y_scale.scale.invert(curr_pos[1]);
```
martinRenou: I actually think this `scale` property should be private on scales. That would prevent this at least.

martinRenou:
@supriyakhandekar The BrushSelector with the OrdinalScale should work on master now :)

PR #1045, opened 2020-03-03
Update to work in JupyterLab 2.0

jasongrout:
I think this is all that is needed to get bqplot to work in JupyterLab 2.0 while still having it work in JupyterLab 1.x.

Notice that I added some phosphor dependencies, which are needed (we are importing from phosphor, but didn't list phosphor as a dependency).

Is there a reason for not having `@jupyter-widgets/base@^1` as a dependency, just curious? In other words, a version of `^1 || ^2 || ^3`?

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -56,7 +56,9 @@
     "webpack-cli": "^3.0.8"
   },
   "dependencies": {
-    "@jupyter-widgets/base": "^2.0.0",
+    "@jupyter-widgets/base": "^2 || ^3",
+    "@phosphor/messaging": "^1",
+    "@phosphor/widgets": "^1",
     "d3": "^5.7.0",
     "d3-selection-multi": "^1.0.1",
     "is-typedarray": "^1.0.0",


js/package.json
```diff
@@ -56,7 +56,9 @@
     "webpack-cli": "^3.0.8"
   },
   "dependencies": {
-    "@jupyter-widgets/base": "^2.0.0",
+    "@jupyter-widgets/base": "^2 || ^3",
+    "@phosphor/messaging": "^1",
+    "@phosphor/widgets": "^1",
```
martinRenou: Isn't it needed to use `lumino` now? I guess it will still compile and work just fine though.

martinRenou: Isn't it needed to use `lumino` now? I guess it will still compile and work just fine though.

jasongrout: If you want this codebase to still work with jlab 1, we can't move to lumino. Jlab 2 will automatically force using lumino as needed.

martinRenou: Makes sense, thanks!

SylvainCorlay:
> Is there a reason for not having @jupyter-widgets/base@^1 as a dependency, just curious? In other words, a version of ^1 || ^2 || ^3?

Yes. I am not sure I remember the exact reasoning, but it was related to the fact that bqplot makes use of JupyterLab CSS variables, which are not exposed by `base ^1`. in the classic notebook context.

jasongrout:
> Yes. I am not sure I remember the exact reasoning, but it was related to the fact that bqplot makes use of JupyterLab CSS variables, which are not exposed by `base ^1`. in the classic notebook context.

Ah. No problem, it just means that the current bqplot js doesn't work in jlab 0.35, but I think that's fine.

PR #1047, opened 2020-03-04
Relayout figures after show events.

### LINKED ISSUE #1023: Problem displaying bqplot within an ipywidget tabbed interface

jasongrout:
Fixes #1023

**Describe your changes**
Relayout a figure on the after-show event. Also, optimize a bit by not laying out figures that are known to be invisible (for example, in a background tab or accordion slot.

**Testing performed**

```python
from ipywidgets import Tab
f=plt.figure()
plt.plot([1,2,3], [1,2,3])
g=plt.figure()
plt.plot([3,2,1], [1,2,3])
Tab(children=[f, g])
```

Currently the plot in the non-active tab is rendered in a minimal size and not updated. This updates that plot when it is shown by activating a tab. This also optimizes a bit by not redrawing the figures in non-active tabs as the frame is resized.

**Additional context**

This is an alternative fix to https://github.com/bloomberg/bqplot/pull/1029. It is not as fancy (i.e., it doesn't use the browser visibility primitives), but it is much simpler.

### DIFF
--- a/js/src/Figure.ts
+++ b/js/src/Figure.ts
@@ -551,8 +551,9 @@ class Figure extends widgets.DOMWidgetView {
         super.processPhosphorMessage.apply(this, arguments);
         switch (msg.type) {
         case 'resize':
+        case 'after-show':
             const figureSize = this.getFigureSize();
-            if ((this.width !== figureSize.width) || (this.height !== figureSize.height)) {
+            if (this.pWidget.isVisible && ((this.width !== figureSize.width) || (this.height !== figureSize.height))) {
                 this.debouncedRelayout();
             }
             break;
--- a/js/src/Figure.ts
+++ b/js/src/Figure.ts
@@ -552,9 +552,12 @@ class Figure extends widgets.DOMWidgetView {
         switch (msg.type) {
         case 'resize':
         case 'after-show':
-            const figureSize = this.getFigureSize();
-            if (this.pWidget.isVisible && ((this.width !== figureSize.width) || (this.height !== figureSize.height))) {
-                this.debouncedRelayout();
+            if (this.pWidget.isVisible) {
+                const figureSize = this.getFigureSize();
+                if ((this.width !== figureSize.width) || (this.height !== figureSize.height)) {
+                    this.debouncedRelayout();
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
Thanks, we discussed it with Martin, and we think that this is a preferable fix.

Using the visibility primitives of the browser actually may be a good optimization to include upstream (in Lumino) -  although they may be costly themselves.

I think that we can close #1029.

jasongrout:
> Using the visibility primitives of the browser actually may be a good optimization to include upstream (in Lumino)

I think it makes a lot of sense to explore how to use browser visibility primitives in general, though I agree they may be much more costly. Also, I've been following work at https://github.com/WICG/display-locking, which may be relevant.

Issue #1048, opened 2020-03-04
Feature request: BrushCircleSelector and/or BrushEllipseSelector


astrofrog:
Currently bqplot implements a number of selection interactions including ``BrushIntervalSelector`` and ``BrushSelector``. It would be nice to have the ability to do circular/elliptical selections, so this is a feature request for a ``BrushCircleSelector`` (or ``BrushEllipseSelector``). Perhaps an ellipse selection makes more sense to start with, later adding the option to constrain the aspect ratio in screen or data coordinates.

PR #1050, opened 2020-03-06
feat: ellipse brushing

maartenbreddels:
*Issue number of the reported bug or feature request: #1048*

**Describe your changes**
This adds an EllipseBrushSelector, which can be used as circular selector. To enforce a circular selector, put pixel_aspect=1. Note that this means the aspect is 1 in pixel coordinates, another requirement that could be added is having an aspect ratio in data coordinates (domain). This PR does not address that, and I think both can be orthogonal in use cases.

**Testing performed**
All operations are testing in the test framework.

**Issue**
Reshaping may change the aspect ratio of the ellipse (as visible in the demo), due to the implementation of how reshaping/dragging is implemented. We use the same code as the brushing, which assuming if you move X pixels in the x-direction, and Y pixels in the y-direction, you want an ellipse of aspect ratio X/Y. While actually when dragging the border, you probably want to calculate the 'angle' (more formally the eccentric anomaly https://en.wikipedia.org/wiki/Ellipse#Parametric_representation) and use that as a 'grab' point. I am not sure yet how this should be implemented, or what 'feels' natural.

Ideas are welcome: @astrofrog  @SylvainCorlay @martinRenou 

**Demo**
ellipse


maartenbreddels:
The reshaping is more natural now.
ellipse-2

astrofrog:
Works great! I ran into two small issues:

* The clip rectangle of the axes isn't recognized as it is for other interacts:

clip
 
* Setting ``color`` on the interact doesn't work as it does for other interacts - I need to set the style/border_style directly.

maartenbreddels:
TODO: 'backport' https://github.com/glue-viz/bqplot-image-gl/pull/23

PR #1059, opened 2020-03-17
Relayout on after-attach event

### LINKED ISSUE #1055: Figure size drawn incorrectly when used as a tooltip
Looks like the same issue that was fixed by #1029, but is still reproducible.
@SylvainCorlay @martinRenou 

```python
from bqplot import *
# Figure 1
sc1 = OrdinalScale()
sc2 = LinearScale()
bar2 = Bars(x=[1,2,3], y=[3,2,1], scales={'x': sc1, 'y': sc2})
ax_x2 = Axis(scale=sc2, orientation='vertical')
ax_y2 = Axis(scale=sc1)
tooltip_fig = Figure(marks=[bar2], axes=[ax_x2, ax_y2])

# Figure 2
sc3 = OrdinalScale()
sc4 = LinearScale()
bar = Bars(x=[1,2,3], y=[1,2,3], scales={'x': sc3, 'y': sc4}, tooltip=tooltip_fig)
ax_x = Axis(scale=sc3)
ax_y = Axis(scale=sc4, orientation='vertical')

Figure(marks=[bar], axes=[ax_x, ax_y], layout={'height':'300px'})
```
Screen Shot 2020-03-12 at 10 47 01 PM

martinRenou:
relayout

Should fix #1055

### DIFF
--- a/js/src/Figure.ts
+++ b/js/src/Figure.ts
@@ -115,6 +115,12 @@ class Figure extends widgets.DOMWidgetView {
         this.margin = this.model.get("fig_margin");
 
         this.update_plotarea_dimensions();
+        // we hide it when the plot area is too small
+        if ((this.plotarea_width < 1) || (this.plotarea_width < 1) ) {
+            this.el.style.visibility = "hidden";
+        } else {
+            this.el.style.visibility = "";
+        }
         // this.fig is the top <g> element to be impacted by a rescaling / change of margins
 
         this.fig = this.svg.append("g")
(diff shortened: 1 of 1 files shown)


martinRenou:
This is not ideal. Because, as you can see in the GIF, the figure is displayed a first time with a size of `0`, and due to the margins it actually gets a size of `-20 x -20`.

@maartenbreddels I know you did not agree with this last time we talked about it, but I really think that when the Figure has no available space, it should not be displayed at all. This would fix this glitch.

maartenbreddels:
This is now also fixed, by making the whole element invisible (so it takes up DOM space if needed).

PR #1060, opened 2020-03-17
Add default sizing for a figure in tooltip

martinRenou:
This will give a default size to the figure when added in a tooltip. The size used is 2/3 of the old natural size, which was 640/480

tooltip

### DIFF
--- a/js/less/bqplot.less
+++ b/js/less/bqplot.less
@@ -284,6 +284,10 @@
     -moz-user-select: var(--bq-mark-tooltip--moz-user-select);
     background-color: var(--bq-mark-tooltip-background-color);
 }
+.mark_tooltip > .bqplot.figure {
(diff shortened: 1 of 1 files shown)


maartenbreddels:
I agree with this. Just to make this clear, this is only the size it will get when it's used as a top-level popup. A few questions:
 * Do we want this as a default for any popup? Also a vbox/hbox/button etc?

martinRenou:
It might make sense to use it for other widgets like ipyleaflet maps. But it might be difficult to find a rule that always work. Because we don't want this for sliders/buttons.

martinRenou:
In the meantime, maybe we can merge this?

maartenbreddels:
I'm in favour of libraries opting in themselves. So I'd say we should merge this.

PR #1066, opened 2020-03-24
feat: MouseInteraction to capture mouse events

maartenbreddels:
To implement https://github.com/glue-viz/glue-jupyter/issues/168

**Describe your changes**
A new Interaction widget to allow mouse events to be captured in the kernel.

(Screenc apture will come, licecap does not seem to work)


maartenbreddels:
@astrofrog noted that the x coordinate seems a bit off. It is difficult to see from this screen capture though.

ibdafna:
LGTM!

astrofrog:
This works great for me :+1:

martinRenou:
Rebased

maartenbreddels:
Thanks, I converted this to draft, because I'm not yet sure this is the way to go, since this does not compose with other interacts, for instance you cannot have zoom/pan *and* have the mouse events at the same time. We did take this approach in bqplot-image-gl, but lets see first what lessons we can learn from there.
I can imagine we want this feature on the Figure instead.

As a sidenode, ipyevents can do some of this, but I will not be possible to get the domain coordinates using that, so we will always want a bqplot specific solution.

Issue #1067, opened 2020-03-25
BrushSelector does not update when selection changed programmatically


astrofrog:
**Describe the bug**

If ``selected_x`` and/or ``selected_y`` are changed on a brush selector, the visible selection rectangle does not update immediately. Instead, the browser window needs to be resized or refreshed.

**To Reproduce**

Create a figure with a brush selector:

```python
from bqplot import pyplot as plt
from bqplot.interacts import BrushSelector
fig = plt.figure()
plt.plot([1, 2, 3], [2, 4, 3])
plt.show()
interact = BrushSelector(x_scale=fig.axes[0].scale, y_scale=fig.axes[1].scale)
fig.interaction = interact
```

Then click and drag to draw a shape. Once this is done, change ``selected_x`` and ``selected_y``:

```python
interact.selected_x = [1.4, 2.4]
interact.selected_y = [2.4, 4]
```

The rectangle does not update unless the window is resized or refreshed.

**Expected behavior**

The brush selector should update immediately

**Screenshots**

![Peek 2020-03-25 14-06](https://user-images.githubusercontent.com/314716/77551829-b21d9400-6eaa-11ea-9998-f76a7a9cd58c.gif)

**Environment (please complete the following information):**
 - Operating System and Version: Ubuntu 20.04 (pre-release)
 - Browser: Firefox

**Additional context**
It looks like this is still an issue with the latest developer version of bqplot

maartenbreddels:
Also, I expect:
```
fig.interaction = None
fig.interaction = interact
```
To draw the BrushSelector.
The workaround, to force it to draw is to modify fig_margin:
```
fig.fig_margin = {'top': 60, 'bottom': 60, 'left': 60, 'right': 61}
```

SylvainCorlay:
Fixed in #1069 !

PR #1069, opened 2020-03-31
fix: brush DOM state was not always in sync with model

maartenbreddels:
*Issue number of the reported bug or feature request: #1067*

**Describe your changes**
BrushSelector.relayout was doing too much, a part is split off into the method `syncModelToBrush`, which is now called in render and in selected_changed (which is called when the selected_x/y properties change).

Also, render is made async, and modernized, so that it's only attached to the DOM after the scales are created. This also makes the unittests much cleaner.

**Testing performed**
Full unittestst included, which tests if the rectangle (green) is drawn (pixel color is tested).

### DIFF
--- a/js/src/BrushSelector.ts
+++ b/js/src/BrushSelector.ts
@@ -339,32 +339,31 @@ export class BrushSelector extends BrushMixinXYSelector {
 
 export class BrushIntervalSelector extends BrushMixinXSelector {
 
-    render() {
-        super.render();
+    async render() {
+        await super.render();
         this.brush_render();
 
-        const that = this;
-        const scale_creation_promise = this.create_scales();
-        Promise.all([this.mark_views_promise, scale_creation_promise]).then(function() {
-            that.brush = (that.model.get("orientation") == "vertical" ? d3.brushY() : d3.brushX())
-              .on("start", _.bind(that.brush_start, that))
-              .on("brush", _.bind(that.brush_move, that))
-              .on("end", _.bind(that.brush_end, that));
-            that.brush.extent([[0, 0], [that.width, that.height]]);
+        await this.mark_views_promise;
+        await this.create_scales();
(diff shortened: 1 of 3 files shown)


maartenbreddels:
The same fixes are done for the interval selector, we can change them programmatically, and they DOM state will always be in sync with the model.

astrofrog:
This looks great and solves the issues I was seeing with the rectangular selection tool.

js/src/BrushSelector.ts
```diff
@@ -424,15 +426,17 @@ export class BrushIntervalSelector extends BrushMixinXSelector {
...
         this.set_range([this.scale]);
         this.brush.extent([[0, 0], [this.width, this.height]]);
+        this.syncModelToBrush();
+    }
 
+    private syncModelToBrush() {
         if(this.model.get("selected")) {
             const range = this.model.get("selected").map(this.scale.scale).sort(
                 function(a, b) { return a - b; });
+            console.log(range)
```
SylvainCorlay, line 436: Outstanding debug statement.

SylvainCorlay:
@maartenbreddels this change appears to cause the brush state to bounce between front-end and backend - or between views.

I think we need to rethink this...

js/src/BrushSelector.ts
```diff
@@ -412,6 +413,7 @@ export class BrushIntervalSelector extends BrushMixinXSelector {
                 (a: number, b: number) => a - b);
             this.update_mark_selected(pixel_extent);
         }
+        this.syncModelToBrush();
```
SylvainCorlay, line 416: This is the reason why we had the bounce. There is a first empty_selection called when we start brushing which was causing the bouncing. In #1098, I set the `brushing` attribute to `true` earlier so that it does not cause bouncing.

maartenbreddels:
Just to make sure, we don't need to revert any of this right?

ibdafna:
This PR seems to be causing #1180. We might need to revisit the logic

PR #1076, opened 2020-04-16
Make wealth of nations ticks a bit nicer

jasongrout:
This makes it clearer the x scale is a log scale.

bqplot-fs8


SylvainCorlay:
I love it! Merging.

PR #1084, opened 2020-04-22
Add image pixelation control

agoose77:
I'd opened an old PR for this, which I've now closed.

This PR adds the ability to enable pixellation for an image. This is useful when using an Image display as a heatmap, as it is more performant than the builtin heatmap.


maartenbreddels:
See also bqplot-image-gl, but I like it, I think I had a similar PR opened once.

agoose77:
> See also bqplot-image-gl, but I like it, I think I had a similar PR opened once.

You absolutely did, I must have lost the attribution somewhere. This is based upon your own work elsewhere.

I did see bqplot-image-gl, but at the time (last year) I could not get it to work with my env.

SylvainCorlay:
Thanks! Merging.

PR #1086, opened 2020-04-24
Make debounvedRelayout a method

### LINKED ISSUE #1087: debouncedRelayout used before assigned
debouncedRelayout is called before assigned.

```
TypeError: this.debouncedRelayout is not a function
```

The culprit is this call.

```js
        switch (msg.type) {
            case 'resize':
            case 'after-show':
            case 'after-attach':
                if (this.pWidget.isVisible) {
                    const figureSize = this.getFigureSize();
                    if ((this.width !== figureSize.width) || (this.height !== figureSize.height)) {
                        this.debouncedRelayout();
                    }
                }
                break;
        }
```

SylvainCorlay:
Fixes #1087.

cc @maartenbreddels

### DIFF
--- a/js/src/Figure.ts
+++ b/js/src/Figure.ts
@@ -262,9 +262,6 @@ class Figure extends widgets.DOMWidgetView {
 
             // In the classic notebook, we should relayout the figure on
             // resize of the main window.
-            this.debouncedRelayout = _.debounce(() => {
-                this.relayout();
-            }, 300);
             window.addEventListener('resize', this.debouncedRelayout);
             this.once('remove', () => {
                 window.removeEventListener('resize', this.debouncedRelayout);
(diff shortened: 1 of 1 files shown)


js/src/Figure.ts
```diff
@@ -274,6 +271,12 @@ class Figure extends widgets.DOMWidgetView {
         });
     }
 
+    debouncedRelayout() {
+        _.debounce(() => {
```
martinRenou, line 275: Not sure that works. `_.debounce` returns the debounced function. So `debouncedRelayout` must be the output of `_.debounce`, it should not call it.

martinRenou: I saw this issue multiple times, and I am not sure how to properly fix it in TypeScript. But the proper JavaScript way of doing it might be to change the prototype manually.

maartenbreddels:
see https://github.com/bqplot/bqplot/pull/1072 it needs to be an object, it cannot be a method, since it needs to be per Figure object.

js/src/Figure.ts
SylvainCorlay: oh right

martinRenou:
Oh I see. That is a bit annoying in terms of design don't you think? But then your PR is the only way we can get it "right" I guess.

maartenbreddels:
Yes, it needs to be a per object thing. In Python, it can be done with a decorator like I do in vaex, but I don't know of a better way for Javascript.

PR #1097, opened 2020-04-28
fix: redraw the d3-brush when the domain/screen changes

### LINKED ISSUE #1085: Selections only working in initially visible tab/accordion containers
It appears that selections are only working on `Figure`s in the initially visible tab of ipywidgets containers. In the example below, note that the selection only works on the first tab that is visible. I had thought that this may be related to #1047 , but it seems that some DOM elements for the selector are missing?

```python
from bqplot import *
from ipywidgets import *
from bqplot.interacts import *

scale_x1 = LinearScale()
scale_y1 = LinearScale()
scatter1 = Scatter(x=[1,2,3], y=[3,2,1], scales={'x':scale_x1, 'y':scale_y1})
fig1 = Figure(marks=[scatter1], interaction=BrushSelector(x_scale=scale_x1,
                     y_scale=scale_y1, marks=[scatter1]), background_style={'fill':'#eee'})

scale_x2 = LinearScale()
scale_y2 = LinearScale()
scatter2 = Scatter(x=[1,2,3], y=[3,2,1], scales={'x':scale_x1, 'y':scale_y1})
fig2 = Figure(marks=[scatter2], interaction=BrushSelector(x_scale=scale_x2,
                     y_scale=scale_y2, marks=[scatter2]), background_style={'fill':'#eee'})

tabs = widgets.Tab([fig1,fig2])
tabs.selected_index=0
tabs.set_title(0,'Selector works')
tabs.set_title(1,"Selector doesn't work")

tabs
```

maartenbreddels:
Fixes #1085

Caused by #1069 ffaecc2589f9ba18f06ed7596a0be072f79a0f1e

I removed the call to this.d3el.call(this.brush); because it gave strange results.
The 'strange' results were actually caused by calling extend, instead of move.

Look a bit of time because we did not have #1096

### DIFF
--- a/js/src/BrushSelector.ts
+++ b/js/src/BrushSelector.ts
@@ -296,7 +296,7 @@ export class BrushSelector extends BrushMixinXYSelector {
         } else {
             const extent = [[selected_x[0], selected_y[0]],
                           [selected_x[1], selected_y[1]]];
-            this.brush.extent(extent);
+            this.brush.move(this.d3el, extent);
             const pixel_extent_x = selected_x.map(this.x_scale.scale).sort(
                 function(a, b) { return a - b; });
             const pixel_extent_y = selected_y.map(this.y_scale.scale).sort(
@@ -317,7 +317,9 @@ export class BrushSelector extends BrushMixinXYSelector {
 
         this.brush.extent([[0, 0], [this.width, this.height]]);
         this.syncModelToBrush();
+        this.brushsel = this.d3el.call(this.brush);
     }
+
 
     private syncModelToBrush() {
         if(this.model.get("selected_x") && this.model.get("selected_y")) {
@@ -327,6 +329,7 @@ export class BrushSelector extends BrushMixinXYSelector {
                 function(a, b) { return a - b; });
             this.brush.move(this.d3el, [[range_x[0], range_y[0]], [range_x[1], range_y[1]]]);
         }
+        this.brushsel = this.d3el.call(this.brush);
     }
 
     // TODO: check that we've properly overridden the mixin.


SylvainCorlay:
@kaiayoung feel free to merge!

SylvainCorlay:
@maartenbreddels I may have to revert this actually. This appears to be causing a regression.

js/src/BrushSelector.ts
```diff
@@ -317,8 +317,10 @@ export class BrushSelector extends BrushMixinXYSelector {
 
         this.brush.extent([[0, 0], [this.width, this.height]]);
         this.syncModelToBrush();
+        this.brushsel = this.d3el.call(this.brush);
```
martinRenou, line 320: This seems to be executed twice, once in `syncModelToBrush` and once here, should we remove one of those lines @maartenbreddels ?

maartenbreddels: Indeed, this seems not needed.

js/src/BrushSelector.ts
```diff
@@ -296,7 +296,7 @@ export class BrushSelector extends BrushMixinXYSelector {
         } else {
             const extent = [[selected_x[0], selected_y[0]],
                           [selected_x[1], selected_y[1]]];
-            this.brush.extent(extent);
+            this.brush.move(this.d3el, extent);
```
maartenbreddels, line 299: @martinRenou this one might also be needed in #1202 I think when you update selected_x and selected_y from the kernel, this misbehaved. I should have better explained this in this PR.

martinRenou: Ok I will apply it to my PR.

PR #1100, opened 2020-04-29
Fix case of empty fast interval selector

SylvainCorlay:


SylvainCorlay:
This makes the fast interval selector behave exactly the same way as the brush selector in that regard.

PR #1111, opened 2020-05-06
preserve array attributes when duplicating them

### LINKED ISSUE #1108: Dragging incompatible with DateScale
When ending a dragging operation the values on an axis of type `DateScale` are corrupted. They change their type from `numpy.datetime64` to `numpy.float64`. In general this happens everytime a message with `"method":"update"` is sent. That also means that if `update_on_move` is set for the mark it happens already immediately after the dragging action is started.

### to reproduce
```
import ipywidgets
import bqplot
import numpy as np
sx = bqplot.DateScale()
sy = bqplot.LinearScale()
ax = bqplot.Axis(scale=sx)
ay = bqplot.Axis(scale=sy, orientation='vertical')
x = np.vectorize(np.datetime64)(["2020-01-01", "2020-01-02", "2020-01-03"])
y = [1., 2., 3.]
mark = bqplot.Scatter(x=x, y=y, scales={'x': sx, 'y': sy})
mark.enable_move = True
#mark.update_on_move = True
fig = bqplot.Figure(marks=[mark], axes=[ax, ay])
ta = ipywidgets.Textarea()
def dr(mark, event):
    ta.value = str(mark.x)
mark.on_drag(dr)
def de(mark, event):
    ta.value = str(mark.x)
mark.on_drag_end(de)
ipywidgets.VBox([fig, ta])
```

### solution
The behavior was investigated already. It was introduced by c62e8a146 where a `slice` operation is supposed to make a copy of an array. In the case of this bug the `slice` should copy a `Float64Array` which has `"type": "date"` attached as attribute. The output of slice does not have the attribute attached. 

The following javascript example demonstrates that `slice` doesn't copy the attributes on an array:

```
a = new Float64Array()
a.myAttribute = "someValue"
b = a.slice()
console.log(a.myAttribute)
console.log(b.myAttribute)    // error: attribute not present
```

There are two options how this can be fixed:

-cleaner fix: not putting foreign attributes on an array. People don't expect attributes to be added to an array and doing so can have unpredictable effects (e.g. if array decides to use such an attribute in the future). So, the type information would have to go somewhere else. It could for instance be part of a class that wraps the array.
-quicker fix: assign and slice: `b = Object.assign(a.slice(), a)` (arrays' attributes stay "polluted")

### recurrences

There are also the following places in the code where attributes would be lost due to the same reasons described above:

```
MarketMap.ts:            const selected = this.model.get("selected").slice();
ScatterBase.ts:            const x = this.model.get('x').slice(); // copy
ScatterBase.ts:            const y = this.model.get('y').slice()
ScatterBase.ts:        let x = this.model.get("x").slice();
ScatterBase.ts:        let y = this.model.get("y").slice();
utils.ts:        return obj.slice()
```

### versions
Google Chrome Version 81.0.4044.129 (Official Build) (64-bit)
bqplot.__version__ == '0.12.9'

joders:
This fixes #1108 (it's what is described as the "quicker fix" in the issue)

### DIFF
--- a/js/src/MarketMap.ts
+++ b/js/src/MarketMap.ts
@@ -514,7 +514,7 @@ export class MarketMap extends Figure {
 
     cell_click_handler(data, id, cell) {
         if(this.model.get("enable_select")) {
-            const selected = this.model.get("selected").slice();
+            const selected = Object.assign(this.model.get("selected").slice(), this.model.get("selected"));
             const index = selected.indexOf(data.name);
             if(index == -1) {
                 // not already selected, so add to selected
--- a/js/src/ScatterBase.ts
+++ b/js/src/ScatterBase.ts
@@ -560,12 +560,12 @@ export abstract class ScatterBase extends Mark {
             y_scale = this.scales.y;
 
         if (!this.model.get("restrict_y")){
-            const x = this.model.get('x').slice(); // copy
+            const x = Object.assign(this.model.get('x').slice(), this.model.get('x')); // copy
             x[i] = x_scale.invert(d[0]);
             this.model.set("x", x);
         }
         if (!this.model.get("restrict_x")){
-            const y = this.model.get('y').slice()
+            const y = Object.assign(this.model.get('y').slice(), this.model.get('y'))
             y[i] = y_scale.invert(d[1]);
             this.model.set("y", y);
         }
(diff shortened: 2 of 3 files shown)


SylvainCorlay:
This looks good to me - and I find the solution with `Object.assign` elegant. Thanks!

PR #1113, opened 2020-05-07
Fix bars coloring computation

martinRenou:
Should fix https://github.com/bqplot/bqplot/issues/1110

### DIFF
--- a/js/src/Bars.ts
+++ b/js/src/Bars.ts
@@ -689,35 +689,21 @@ export class Bars extends Mark {
         //if y is 1-d, each bar should be of 1 color.
         //if y is multi-dimensional, the corresponding values should be of
         //the same color.
-        const that = this;
-        const color_scale = this.scales.color;
         if (this.model.mark_data.length > 0) {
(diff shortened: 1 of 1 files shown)


js/src/Bars.ts
```diff
@@ -831,19 +810,23 @@ export class Bars extends Mark {
...
+            .selectAll('.bar')
+            .style("fill", fill ? this.get_mark_color.bind(this) : "none")
+            .style("stroke", stroke ? stroke : this.get_mark_color.bind(this))
+            .style("opacity", this.get_mark_opacity.bind(this))
+            .style("stroke-width", stroke_width);
+    }
+
+    get_mark_color(data, index) {
+        // Workaround for the bargroup, the color index is not the same as the bar index
+        return super.get_mark_color(data, data.color_index);
```
martinRenou, line 829: This is the fix. The `index` passed by d3 was wrong in the case of having bars under bargroups, so we need to pass the `color_index` to `get_mark_color` manually.

SylvainCorlay:
Many thanks!

PR #1114, opened 2020-05-07
fix: HandDraw resulted in NaN/blanks because of TypeArray

maartenbreddels:


maartenbreddels:
During drawing, you'd get no line, that's fixed now.

PR #1117, opened 2020-05-07
fix(Lines): set id properly on initial draw

maartenbreddels:
Follow up of #1114

On initial draw, a handdraw would still not redraw the line / set the y data.

The problem was that the curve id wasn't set. `curves_sel` is initially an empty d3 selection, and after we attached data and added the nodes, we should reassign to `curves_sel` it seems. I thought d3 would keep the DOM and its selection in sync, but apparently it does not, or something else is going on.

Some more background why it doesn't work: https://github.com/bqplot/bqplot/blob/f2f01a3df3b7238ea126363bc5f93c70d89d1a92/js/src/HandDraw.ts#L118 use the id to set the data.

### DIFF
--- a/js/src/Lines.ts
+++ b/js/src/Lines.ts
@@ -545,7 +545,7 @@ export class Lines extends Mark {
 
     draw(animate) {
         this.set_ranges();
-        const curves_sel = this.d3el.selectAll(".curve")
+        let curves_sel = this.d3el.selectAll(".curve")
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
> I thought d3 would keep the DOM and its selection in sync, but apparently it does not, or something else is going on.

Me too, that it weird. Although it fixes the issue so moving on with this. To be investigated.

PR #1123, opened 2020-05-12
Remove redundant code from Lines/FlexLines

martinRenou:
- `Lines.get_element_color` was exactly the same as its superclass's method `Mark.get_mark_color`
- `FlexLines.get_element_color` is slightly different, as it does not cycle through colors. But I feel like it would be good that all marks respects the same logic by cycling through the colors if there are multiple ones.


SylvainCorlay:
> FlexLines.get_element_color is slightly different, as it does not cycle through colors. But I feel like it would be good that all marks respects the same logic by cycling through the colors if there are multiple ones.

I agree.

PR #1126, opened 2020-05-12
Fix lines removing

martinRenou:
This should fix line removing:

removing

### DIFF
--- a/js/src/Lines.ts
+++ b/js/src/Lines.ts
@@ -536,7 +536,7 @@ export class Lines extends Mark {
 
     draw(animate) {
         this.set_ranges();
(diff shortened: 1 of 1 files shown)


js/src/Lines.ts
```diff
@@ -558,7 +558,7 @@ export class Lines extends Mark {
 
         const fill = this.model.get("fill"),
             area = (fill === "top" || fill === "bottom" || fill === "between");
-        curves_sel = this.d3el.selectAll(".curve");
```
martinRenou, line 561: We were overwriting the selection here, resulting in the `.exit()` to be empty, so we could not remove the lines. @JohanMabille would you mind reviewing this?

PR #1130, opened 2020-05-13
Add support for serializing list of variant-sized lists

### LINKED ISSUE #1129: Boxplot not working with arrays of data with different lengths
It appears that `BoxPlot()`s can no longer be instantiated with arrays in the `y` trait that have different lengths. See the code below:

```python
from bqplot import *

y_data = [
   [0,1,2,3,4,5,6],
   [0,1,2,3,4,5],
   [0,1,2,3,4,5,6],
]

boxes = Boxplot(x=[1, 2, 3], y=y_data, scales={'x': OrdinalScale(), 'y': LinearScale()})
Figure(marks=[boxes])
```
err

martinRenou:
Will fix #1129

### DIFF
--- a/bqplot/traits.py
+++ b/bqplot/traits.py
@@ -157,16 +157,16 @@ def array_to_json(ar, obj=None, force_contiguous=True):
 
     if ar.dtype.kind == 'O':
         # Try to serialize the array of objects
-        isstring = np.vectorize(lambda x: isinstance(x, six.string_types))
-        islist = np.vectorize(lambda x: isinstance(x, list))
-        istimestamp = np.vectorize(lambda x: isinstance(x, pd.Timestamp))
+        is_string = np.vectorize(lambda x: isinstance(x, six.string_types))
+        is_timestamp = np.vectorize(lambda x: isinstance(x, pd.Timestamp))
+        is_array_like = np.vectorize(lambda x: isinstance(x, (list, np.ndarray)))
 
-        if np.all(istimestamp(ar)):
+        if np.all(is_timestamp(ar)):
             ar = ar.astype('datetime64[ms]').astype(np.float64)
             array_type = 'date'
-        elif np.all(isstring(ar)):
+        elif np.all(is_string(ar)):
             ar = ar.astype('U')
-        elif np.all(islist(ar)):
+        elif np.all(is_array_like(ar)):
             return [array_to_json(np.array(row), obj, force_contiguous) for row in ar]
         else:
             raise ValueError("Unsupported dtype object")
--- a/bqplot/traits.py
+++ b/bqplot/traits.py
@@ -152,49 +152,48 @@ def array_from_json(value, obj=None):
 def array_to_json(ar, obj=None, force_contiguous=True):
     if ar is None:
         return None
-    if ar.dtype.kind in ['O']:
-        has_strings = False
-        all_strings = True  # empty array we can interpret as an empty list
-        for el in ar:
-            if isinstance(el, six.string_types):
-                has_strings = True
-            else:
-                all_strings = False
-        if all_strings:
+
+    array_type = None
+
+    if ar.dtype.kind == 'O':
+        # Try to serialize the array of objects
+        isstring = np.vectorize(lambda x: isinstance(x, six.string_types))
+        islist = np.vectorize(lambda x: isinstance(x, list))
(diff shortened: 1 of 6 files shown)


martinRenou:
Will fix #1129

martinRenou:
@maartenbreddels @SylvainCorlay

SylvainCorlay:
So we discussed this approach in person with @martinRenou and I am happy with it.

I think we should probably discuss it with @maartenbreddels before merging.

martinRenou:
Yes, I also want to add more JavaScript tests. All Notebooks seem to still work fine though.

js/src/test/binary_serialization.ts
```diff
@@ -1,46 +1,46 @@
...
-    it("date arrays", async function() {
+    it("deserialize/serialize date arrays", async function() {
         // 2005-02-25 and 2 days after:
-        let x_wire = {dtype: 'float64', value: new DataView((new Float64Array([1109289600000, 1109376000000, 1109462400000])).buffer), type: 'date'}
-        let x_ar = array_or_json.deserialize(x_wire, null);
+        const x_wire = {dtype: 'float64', value: new DataView((new Float64Array([1109289600000, 1109376000000, 1109462400000])).buffer), type: 'date'}
+        const x_ar = array_or_json.deserialize(x_wire, null);
         expect(x_ar.type).to.equal('date')
-        let x =  Array.prototype.slice.call(x_ar);
+        const x =  Array.prototype.slice.call(x_ar);
```
maartenbreddels, line 27: FYI: I used slice because I didn't know any better, but I now use a cleaner solution like `[...x_ar]`. You don't have to change it now.

bqplot/traits.py
```diff
@@ -152,49 +152,48 @@ def array_from_json(value, obj=None):
...
 
-    if ar.dtype.kind in ['S', 'U']:  # strings to as plain json
-        return ar.tolist()
-    type = None
+    array_type = None
 
     if ar.dtype.kind == 'O':
-        # If it's a Timestamp object
+        # Try to serialize the array of objects
+        isstring = np.vectorize(lambda x: isinstance(x, six.string_types))
```
maartenbreddels, line 160: should we assume py3? If so, we can remove six?
```suggestion
        isstring = np.vectorize(lambda x: isinstance(x, str))
```

maartenbreddels:
We may also want to support arrays of arrays, I could push this:

```diff
diff --git a/bqplot/traits.py b/bqplot/traits.py
index 0181d2e0..a35571a3 100644
--- a/bqplot/traits.py
+++ b/bqplot/traits.py
@@ -158,7 +158,7 @@ def array_to_json(ar, obj=None, force_contiguous=True):
     if ar.dtype.kind == 'O':
         # Try to serialize the array of objects
         isstring = np.vectorize(lambda x: isinstance(x, six.string_types))
-        islist = np.vectorize(lambda x: isinstance(x, list))
+        islist_or_array = np.vectorize(lambda x: isinstance(x, (list, np.ndarray)))
         istimestamp = np.vectorize(lambda x: isinstance(x, pd.Timestamp))
 
         if np.all(istimestamp(ar)):
@@ -166,7 +166,7 @@ def array_to_json(ar, obj=None, force_contiguous=True):
             array_type = 'date'
         elif np.all(isstring(ar)):
             ar = ar.astype('U')
-        elif np.all(islist(ar)):
+        elif np.all(islist_or_array(ar)):
             return [array_to_json(np.array(row), obj, force_contiguous) for row in ar]
         else:
             raise ValueError("Unsupported dtype object")
diff --git a/tests/binary_serialization_test.py b/tests/binary_serialization_test.py
index c78e12b1..35af1273 100644
--- a/tests/binary_serialization_test.py
+++ b/tests/binary_serialization_test.py
@@ -105,7 +105,7 @@ def test_nested_list():
 
     data = np.array([
         [0, 1, 2, 3, 4, 5, 6],
-        [0, 1, 2, 2, 3],
+        np.array([0, 1, 2, 2, 3]),
         ['foo', 'foobar', 'bar'],
     ])
```

bqplot/traits.py
martinRenou: Do you mind if we do that in another PR? Because it is kind-of unrelated to the change I am making.

martinRenou:
Thanks :smile:

PR #1132, opened 2020-05-15
Reduce tooltip flickering

martinRenou:


maartenbreddels:
Can you explain what the problem was?

martinRenou:
I recall the problem was that a tooltip to a bar chart was flickering when the mouse was moving along the bar.

PR #1140, opened 2020-05-18
Remove Notebooks that were unintentionaly commited

martinRenou:


maartenbreddels:
Should this include an entry in gitignore: `Untitled*.ipynb` ? :)

martinRenou:
Yes good idea

PR #1142, opened 2020-05-18
Improve tooltip styling

martinRenou:
Old visual aspect:

old_tooltip

New visual aspect:

new_tooltip_font

### DIFF
--- a/js/less/bqplot.less
+++ b/js/less/bqplot.less
@@ -50,7 +50,7 @@
     --bq-selector-fill: var(--jp-border-color1);
 
     --bq-mark-tooltip-border: 2px solid var(--jp-border-color0);
-    --bq-mark-tooltip-box-shadow: 1px 1px 2px rgba(245, 245, 245, 0.8);
+    --bq-mark-tooltip-box-shadow: var(--jp-elevation-z6);
     --bq-mark-tooltip-background-color: var(--jp-layout-color0);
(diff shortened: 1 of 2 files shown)


js/less/bqplot.less
```diff
@@ -50,7 +50,7 @@
     --bq-selector-fill: var(--jp-border-color1);
 
     --bq-mark-tooltip-border: 2px solid var(--jp-border-color0);
-    --bq-mark-tooltip-box-shadow: 1px 1px 2px rgba(245, 245, 245, 0.8);
+    --bq-mark-tooltip-box-shadow: var(--jp-elevation-z6);
```
martinRenou, line 53: This seems to be the value used in JupyterLab for box shadows

martinRenou: Actually this variable is not there in Jupyter Notebook, so I now use the actual variable value.

maartenbreddels:
Really nice, about 10x nicer in my opinion :)

martinRenou:
It would also be nice if the tooltip font style had the same default values as the axis font one.

SylvainCorlay:
Could you please add a screenshot with the new font?

martinRenou:
I updated the PR comment with a new screenshot

SylvainCorlay:
This looks *soo* much better.

PR #1149, opened 2020-05-20
Fixup jupyterlab1 theme support

SylvainCorlay:
Ping @jasongrout

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,7 @@ language: python
 python:
   # We don't actually use the Travis Python, but this keeps it organized.
     - 2.7
-    - 3.3
-    - 3.4
-    - 3.5
+    - 3.7
 addons:
   firefox: latest
 install:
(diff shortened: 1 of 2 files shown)


bqplot/_version.py
```diff
@@ -1,8 +1,8 @@
-version_info = (0, 11, 8, 'final', 0)
+version_info = (0, 11, 9, 'final', 0)
```
martinRenou, line 1: Maybe this commit was pushed by accident?

SylvainCorlay: yes, there is an issue with the 0.11.x branch. It is missing the commit for 0.11.9, while the tag is there.

Thankfully, there has been only two commits since. We can probably fix history.

martinRenou: Oh sorry I did not realize it was 0.11

SylvainCorlay:
@jasongrout I am thinking of tagging a new 0.11.x version - also maybe supporting JupyterLab 2.0.

PR #1164, opened 2020-06-19
Add opacity_mode attribute to the Bars class

martinRenou:
Introduce `opacity_mode` attribute to the `Bar` mark, which behaves like the `color_mode` attribute.

As discussed in https://github.com/bqplot/bqplot/issues/1163, we might need another mode that would allow retrieving the behavior that was in `0.11`.

### DIFF
--- a/bqplot/marks.py
+++ b/bqplot/marks.py
@@ -973,6 +973,8 @@ class Bars(Mark):
         'element' means for every dimension of y, all bars have same color.
         'auto' picks 'group' and 'element' for 1-d and 2-d values of
         Y respectively.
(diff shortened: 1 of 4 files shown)


martinRenou:
I added a "real_element" possible value for the color/opacity mode. I really don't like the name but could not come up with something better. Any idea welcome!

martinRenou:
Staying in draft mode, I just want to be sure the `"auto"` mode is correctly respected

ibdafna:
I haven't tested, but the code looks good to me!

martinRenou:
@ibdafna I think it's ready

ibdafna:
LGTM, let's merge!

PR #1167, opened 2020-07-01
Fix stroke color being set fill color, despite default being "none".

ibdafna:
The stroke color of the Bars mark is set to be that of the fill color, despite a default of "none". This creates an issue when the fill attribute of the Bars mark changes after being constructed, as the stroke does not change with it, leading to unintended behavior. 

Before:
image

After:
image

### DIFF
--- a/js/src/Bars.ts
+++ b/js/src/Bars.ts
@@ -814,7 +814,7 @@ export class Bars extends Mark {
             })
             .selectAll('.bar')
             .style("fill", fill ? this.get_mark_color.bind(this) : "none")
-            .style("stroke", stroke ? stroke : this.get_mark_color.bind(this))
+            .style("stroke", stroke ? stroke : "none")
             .style("opacity", this.get_mark_opacity.bind(this))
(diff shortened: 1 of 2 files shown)


js/src/Bars.ts
```diff
@@ -814,7 +814,7 @@ export class Bars extends Mark {
             })
             .selectAll('.bar')
             .style("fill", fill ? this.get_mark_color.bind(this) : "none")
-            .style("stroke", stroke ? stroke : this.get_mark_color.bind(this))
+            .style("stroke", stroke ? stroke : "none")
```
martinRenou, line 817: It's another heterogeneity here. The `fill` attribute being a boolean, I would have expected the `stroke` attribute to be a boolean as well. And I would expect both the fill and the stroke to behave the same when choosing their colors.

You PR is definitely an improvement (even a bug fix) in the current bqplot implementation, but we might be able to make some more improvements on the API.

SylvainCorlay:
Understood. This looks good to me.

Issue #1162, opened 2020-06-15
Text location


npschafer:
https://github.com/bqplot/bqplot/blob/ac85a703f404be4e5209787ca576fefd76da25c4/js/src/Scatter.ts#L192

Thanks for taking the time to make such an awesome package and for releasing it for free and as open source.

I would rather not have my point labels ('names') offset by (text_loc, -text_loc) (see the above line reference). Is there any way to accomplish this without directly editing the Scatter.ts file?

Cheers

ibdafna:
I'll take a look to see if we can add something on the python side to control this behaviour.

ibdafna:
Implemented in #1168

PR #1168, opened 2020-07-07
Add Scatter mark label positioning offset

ibdafna:
Requested in #1162, this PR adds horizontal and vertical offsets for the 'names' attribute in the Scatter mark, which overlays labels above each data point. It adds two float traitlets to the python side:

1. ~name_offset_x~ label_display_horizontal_offset (float, value denotes pixels, default value=None)
2. ~name_offset_y~ label_display_vertical_offset (float, value denotes pixels, default value=None)

Both of these are optional, and if no values are passed to the constructor, positioning will fall back to the default behaviour currently exhibited. 

scatter_labels

### DIFF
--- a/bqplot/marks.py
+++ b/bqplot/marks.py
@@ -610,6 +610,12 @@ class Scatter(_ScatterBase):
         Labels for the points of the chart
     display_names: bool (default: True)
         Controls whether names are displayed for points in the scatter
+    name_offset_x: float (default: None)
+        Adds an offset, in pixels, to the horizontal positioning of the 'names'
+        label above each data point
+    name_offset_y: float (default: None)
+        Adds an offset, in pixels, to the vertical positioning of the 'names'
+        label above each data point
     enable_move: bool (default: False)
         Controls whether points can be moved by dragging. Refer to restrict_x,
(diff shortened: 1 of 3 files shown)


ibdafna:
I thought about doing that, but the reason I went with name_offset is that
oddly we call labels 'names' in the Scatter mark. Ideally we can change
both so it's labels and label_display_offset as you suggest, but that would
be a breaking change. If you don't think it makes a difference whether it's
'names' or 'labels' then I'm happy to make the change anyway! Not strongly
opinionated on this :)

martinRenou:
I also don't have a strong opinion on the naming, I'm fine with whatever that makes sense to you :)

martinRenou:
Though it would be nice to have more homogeneity in the bqplot API, but we'll never get there without breaking changes, so it's more of a long term discussion I guess.

ibdafna:
Changed it. We might as well start making things consistent now instead of waiting in perpetuity for the perfect time to rewrite the APIs.

ibdafna:
@martinRenou okay to merge this?

npschafer:
Thanks a bunch @ibdafna

PR #1170, opened 2020-07-08
Setup.py: Do not install tests

### LINKED ISSUE #1079: Don't include tests/ package in wheel
Currently, `pip install bqplot` installs a top-level package 'tests'. This is bad, for what I hope are obvious reasons. The fix is simple: change `find_packages()` call in 'setup.py' to `find_packages(exclude=["tests"])`. (Be sure to nuke the 'build/' directory before rebuilding; otherwise it may still pick up 'tests'.)

martinRenou:
Fix #1079

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -178,7 +178,7 @@ setup_args = dict(
             'bqplot/static/index.js',
             'bqplot/static/index.js.map',
         ]),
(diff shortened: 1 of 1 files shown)


maartenbreddels:
👍

PR #1171, opened 2020-07-09
Rename "js" directory to "ts"

martinRenou:


SylvainCorlay:
Hum, this may make things difficult to track wrt history, and other open PRs.

martinRenou:
I agree, do you think we should keep it that way then?

SylvainCorlay:
Not sure, let's keep this open and discuss it later.

maartenbreddels:
I suggest we do this after 0.13 when we don't have to backport so often anymore.

ibdafna:
What's the benefit or renaming? I only see net negatives here - we lose git history for the files for a name change which doesn't change much of the meaning.

PR #1177, opened 2020-07-21
Fix Bars mark rendering oversize rectangle / NaN issues

ibdafna:
This PR fixes an issue where updating the data (with partial NaNs) and Y-domain max traits simultaneously, results in an incorrect, oversize rectangle (see gif). It also adds a quick patch for NaNs which are passed to D3's scale, causing console log errors.

The issue is introduced in:
https://github.com/bqplot/bqplot/commit/e11e193339037ffbeec590eaaae33f50666b6641#diff-7fa0ead9ffe41bff2e2270d08efeb58aR365

It seems like we introduced two different transition names, but without any apparent benefit (which also causes the issue described above), so I have reverted that change.

Before:
bqplot_bar_bug

After:
bqplot_bar_fix

### DIFF
--- a/js/src/Bars.ts
+++ b/js/src/Bars.ts
@@ -45,7 +45,7 @@ export class Bars extends Mark {
             this.process_interactions();
             this.create_listeners();
             this.compute_view_padding();
-            this.draw();
+            this.draw(false);
         });
     }
 
@@ -80,12 +80,12 @@ export class Bars extends Mark {
         const x_scale = this.scales.x, y_scale = this.scales.y;
         this.listenTo(x_scale, "domain_changed", function () {
             if (!this.model.dirty) {
-                this.draw();
+                this.draw(false);
             }
         });
         this.listenTo(y_scale, "domain_changed", function () {
             if (!this.model.dirty) {
-                this.draw();
+                this.draw(false);
             }
         });
     }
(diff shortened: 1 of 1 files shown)


js/src/Bars.ts
```diff
@@ -80,12 +80,12 @@ export class Bars extends Mark {
         const x_scale = this.scales.x, y_scale = this.scales.y;
         this.listenTo(x_scale, "domain_changed", function () {
             if (!this.model.dirty) {
-                this.draw();
+                this.draw(false);
```
martinRenou, line 83: Is there a reason for removing the animation when the domain changes or is it a leftover from a debug session?

ibdafna:
I added the boolean to be more explicit, it has the same result if you
don't pass an argument for 'animate'. The draw bars function only applies
animation if animate=true.

js/src/Bars.ts
ibdafna: I changed this flag to true now so we have an animation consistent with the initial screenshot!

ibdafna:
Added transition animation to domain changes, so we're consistent with the transition animation in the initial post!

bqplot_bar_fix_2

martinRenou:
It looks much better with the animation indeed :)

js/src/BarsModel.ts
```diff
@@ -103,6 +103,10 @@ export class BarsModel extends markmodel.MarkModel {
                     if (isNaN(value)) {
                         value = 0;
                     }
+
+                    // so we don't pass NaNs to d3 scale
+                    y_elem[index] = isNaN(y_elem[index]) ? 0 : y_elem[index]
```
kaiayoung, line 108: Are there any side effects if the base of the bars isn't 0?

ibdafna: Setting it to negative baseline did create an issue. Good catch! The weird thing is, I don't see the NaN issues anymore after removing this line. It looks like they were also caused by the transition naming difference.

js/src/BarsModel.ts
```diff
@@ -103,6 +103,7 @@ export class BarsModel extends markmodel.MarkModel {
                     if (isNaN(value)) {
                         value = 0;
                     }
+
```
kaiayoung, line 106: Can we remove this from the commit since there are no longer any changes in this file?

ibdafna:
The NaN problem isn't fixed. I will need to find a fix without using my line of code.

ibdafna:
This PR should be good to merge. The NaN issue is unrelated and I will fix it in a separate PR!

PR #1186, opened 2020-07-31
Invert dark theme detection logic

SylvainCorlay:


SylvainCorlay:
This fixes the gg theme plotarea background in the old classic notebook.

PR #1187, opened 2020-08-04
Added a few advanced applications and dashboards

ChakriCherukuri:
1. Dashboard for performing visual analytics of server logs
2. Equity Index performance time series analysis using fast interval selector
3. Added choropleth to Wealth Of Nations
4. Moved the data to the app folders
5. Added an Index nb which will serve as a starting point for accessing all the apps

I feel these advanced applications showcase the real power and interactivity of bqplot. More apps coming soon.


ibdafna:
Adding notes for a review now

ChakriCherukuri:
Thanks @ibdafna I fixed wild card imports. I'll look at other issues you mentioned

ChakriCherukuri:
@ibadana I fixed all the issues you mentioned except the ones below:

1. Logs Dashboard
  * Donut plots do not show any labels

2. Applications/Wealth of Nations/Choropleth
* Alignment of dropdown/play widgets doesn't fit in regular Jupyter notebook on maximum width

Big dashboards don't render in the narrow output cells of notebook. They render well on a full screen voila. 
Maybe we add a comment in the index nb that these apps render well in voila?

ibdafna:
That's cool, sure. Then yes, it would be sensible to add a disclaimer to let the users know these apps render better in voila compared to normal notebook. Although I must note that most people trying these out will probably not be aware of voila as the example notebooks are typically accessed using regular Jupyter notebook from the Binder link.

It might also be useful to submit these to the voila examples repo so they get more exposure in the right context!

ChakriCherukuri:
@ibdafna added disclaimer about using voila. 

IMHO, I think it's better to add these examples here. Not many people know how to build dashboards using bqplot and these examples provide code for doing the same.

ibdafna:
LGTM!

PR #1189, opened 2020-08-06
Fix incorrect Histogram binning 

ibdafna:
The migration from from D3 v3 to v4 introduced many changes, one of the more relevant ones for the Histogram mark is the d3.histogram() logic change. A change in logic introduced an issue where users passing a bins parameter to the Hist mark constructor were not getting the expected number of bins on the plot:

```python
from bqplot import Hist,Figure, Axis, LinearScale, OrdinalScale
import pandas as pd

df = pd.DataFrame([1,2,3,4],index=['A','B','C','D'])

x_sc = LinearScale()
y_sc = LinearScale()

hist = Hist(sample=df, scales={'sample': x_sc, 'count': y_sc}, bins=4)
ax_x = Axis(scale=x_sc, tick_format='0.2f')
ax_y = Axis(scale=y_sc, orientation='vertical')

Figure(marks=[hist], axes=[ax_x, ax_y], padding_y=0)
```

Before:
image

After:
image

Previously, we used to apply binning logic using the following d3 code:

```javascript
this.mark_data = d3.layout.histogram().bins(this.x_bins).value(function(d) {
                return d.value;
            })(x_data_ind);
```

The `.bins()` method accepted an upper bound for the number of `n` bins. But D3 v4 introduced some changes:
```javascript
this.mark_data = d3.histogram().thresholds(this.x_bins).value(function (d) {
                return d.value;
            })(x_data_ind);
```
The main, and most crucial change, is `thresholds()` returning `n+1` bins. As we were using a custom binning function (see below), the input to the `threshold()` method did not change from the input to the `bins()` method previously used and the value of `this.mark_data` was incorrect as a result. I am now using the built-in `d3.range()` function to generate the bins and passing those to the `thresholds()` method. 

https://github.com/bqplot/bqplot/blob/dc2cf2409b2b81684ee21a2202af973d261e1f04/js/src/HistModel.ts#L159

### DIFF
--- a/js/src/HistModel.ts
+++ b/js/src/HistModel.ts
@@ -18,6 +18,9 @@ import * as d3 from 'd3';
 import { MarkModel } from './MarkModel';
 import * as serialize from './serialize';
 
+//@ts-ignore
+window.d3 = d3;
+
 export class HistModel extends MarkModel {
 
     defaults() {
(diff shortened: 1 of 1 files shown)


ChakriCherukuri:
tested the PR. LGTM

PR #1192, opened 2020-08-12
Try lazy-loading widgets in Lab

### LINKED ISSUE #1190: Lazy-loading in lab
Hi folks!

Looking at a fairly heavy build, I'm seeing a pretty big column for bqplot and its three dependency...

Screenshot_2020-08-11 Webpack Bundle Analyzer  11 Aug 2020 at 16 20

> i'm creating that big opaque blue blob, third from the top in the right-most column :blush: 

Would a PR that moved that whole center column (bqplot + three) into some lazy loaders be welcome? We've done this on e.g. [pythreejs](https://github.com/jupyter-widgets/pythreejs/pull/297/files)?

Pro:
- faster initial load (e.g. to when the moons _start_ spinning)
- faster/less RAM during build

Con:
- delay the first time a bqplot widget is drawn

bollwyvl:
- fixes #1190

This uses the approach from [pythreejs](https://github.com/jupyter-widgets/pythreejs/pull/297/files) to defer loading the widgets until one is actually requested, which should shave nearly 30% off of the built lab. Because threejs is so large, it may also be worth deferring that, too, if possible, as it doesn't appear its usage is guaranteed if someone is using one of the simpler plots.

better still would be to arrive at a common, compatible packaging of threejs, which could be shared by all consumers (e.g. `pythreejs`, `bqplot`, `ipyvolume`) but I have no idea what the compatibility windows are like for all of them.

**2021-06**: sorry i let this slide for so long. I've rejiggered a few things to make working with `package.json` a bit more up-to-date with typescript conventions (weird as they are) with the `composite` build. Would love to see this land to get release along with #1379.

It's probably worth taking a hard look at the `sharedPackages` (e.g. `three.js`, `underscore`) as every extra copy of those heavies we don't have to load `n` times in e.g. jupyterlite makes me :grinning:


jasongrout:
Jlab automatically deduplicates threejs if the versions overlap. That's part of the big reason for the build step.

bollwyvl:
great, that's progress, but still built into `vendor~main` is still quite brutal. Working through it locally, will have some results in a tick...

bollwyvl:
Just because i looked:
```
ipyvolume    "three": "^0.97.0",
bqplot       "three": "^0.91.0",
pythreejs    "three": "^0.97.0",
```

So, hooray :tada:!

bollwyvl:
Here's the situation as of 8b1935e:

```
1.2M Aug 12 16:26 vendors~bqplot.a07da31a0075a039c91a.js
4.4M Aug 12 16:26 vendors~main.881efab41eb56f4e4f00.js
```

Screenshot_2020-08-12 Webpack Bundle Analyzer  12 Aug 2020 at 16 30

I don't actually know where threejs went, as I can't see it in stats, but it's working fine:

Screenshot from 2020-08-12 16-35-18

bollwyvl:
Of additional concern is the extra copy of `@lumino/widgets`... but I assume there's a good, known reason for it.

jasongrout:
> I don't actually know where threejs went

Perhaps it was split out to a different file now? Those choices are made automatically at build time, so adding an extension may lead to some different division.

The extension system rework in jlab 3 leads to more things being split out (and shared)





> Of additional concern is the extra copy of `@lumino/widgets`... but I assume there's a good, known reason for it.

There shouldn't be two copies of lumino widgets, as far as I know, so I think that is indeed a cause for concern.

bollwyvl:
I don't actually know where threejs went
>
> Perhaps it was split out to a different file now?
>

Yeah I'll do some more debugging... I also intend to pull in the other
threejs consumers and see what happens!

> The extension system rework in jlab 3 leads to more things being split out
> (and shared)
>
Oh I know, that work has been epic, and will be really good for users,
admins, and downstreams. As a dev, am awaiting the right time to start
putting some aggressive against master! But I'll be supporting stuff on lab
1 and 2 well into next year, no doubt. 😿

Of additional concern is the extra copy of @lumino/widgets... but I assume
> there's a good, known reason for it.
>
> There shouldn't be two copies of lumino widgets, as far as I know, so I
> think that is indeed a cause for concern.
>
Welp, it's using the phosphor alias thingy, and it's pulling the older
dist, rather than the es6... But would have to dig some more.

>

bollwyvl:
On the duplicate widgets: that was already appearing in the downstream where i [found the issue](https://github.com/bqplot/bqplot/issues/1190#issue-677177537), but the two copies appeared in `main` (of course)... having it split out has just isolated the symptom, I'm fairly sure I didn't _create_ it. 

For reference, a vanilla build only includes the index.es6.js

Screenshot_2020-08-12 Webpack Bundle Analyzer  12 Aug 2020 at 15 32

bollwyvl:
I tried a couple alternate import mechanisms to get at the "real" MessageLoop/Widget, and it's always still pulling its own copy in. While I _think_ it would be possible to get at the Widget class from some part of `@jupyter-widgets/base` or `controls` or something, I don't see a way to get at MessageLoop, as it's basically only ever used statically. Seems like some of these things _could_ be re-exported/normalized in an upstream... it hardly seems like "just an implementation detail" at this point...

bollwyvl:
Yeah, fiddled with some other stuff, no change. I don't think it can be solved from this side, short of changing the phosphumino dependency, which i assume is done to avoid having to maintain separate releases. I'm not entirely sure it's worth fixing this in Lab2 if it hasn't actually been creating problems...

bollwyvl:
As i haven't had to fire up the webpack visualizer in ages (thank GOODNESS), here's a simpler table of the load impact:

Screenshot from 2021-06-29 23-43-17

I've gotta think reducing initial load to 2 requests of ~30k on page setup vs all 11 of ~ 1.3mb is worth the hassle.

js/src/jupyterlab-plugin.ts
```diff
@@ -13,23 +13,28 @@
...
+  id: name,
+  requires: [IJupyterWidgetRegistry],
   activate: function (app, widgets) {
     widgets.registerWidget({
-      name: 'bqplot',
-      version: bqplot.version,
-      exports: bqplot,
+      name,
+      version,
+      exports: async () => import('./index'),
```
martinRenou, line 34: Is this the key feature of this PR? (thinking about doing the same in other widgets lib)

bollwyvl:
Yeah that's the feature, and _requires_ the "module: esnext" setting over
in tsconfig. To get at the package.json from a .ts, with "composite: true"
needed the extra tsconfig jiggery pokery.

bollwyvl:
Have repushed with various updates: it builds/runs locally on Ubuntu, and the tests pass on chromium 98.

guess i'd like to see what CI says before sinking any more time!

martinRenou:
Thanks @bollwyvl, just approved the CI

js/src/jupyterlab-plugin.ts
bollwyvl: yep, this is the key thing.

PR #1196, opened 2020-08-17
Fix screenshot image

maartenbreddels:
cc @rosteen

This fixes screenshots, which did not include images. Images referred to blob URLs, we transform them to dataUrl's so they are embedded in the svg (and thus also appear in the png)

### DIFF
--- a/js/src/Figure.ts
+++ b/js/src/Figure.ts
@@ -833,7 +833,7 @@ class Figure extends widgets.DOMWidgetView {
         // Even though the canvas may display the rendering already, it is not guaranteed it can be read of the canvas
         // or we have to set preserveDrawingBuffer to true, which may impact performance.
         // Instead, we render again, and directly afterwards we do get the pixel data using canvas.toDataURL
-        return this.render_gl().then(() => {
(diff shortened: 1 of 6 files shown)


rosteen:
After pip installing `bqplot` from this branch the `jdaviz` app that I'm testing fails to load. See screenshot for errors in the dev console (they continue, but appear to all be repetitions of the same error):

Screen Shot 2020-08-18 at 1 41 32 PM

maartenbreddels:
did you do a dev install?

maartenbreddels:
FYI: the reason why I think so is that your bqplot/extension.js does not seem to be loaded, which causes requirejs to fetch static/bqplot.js. 
For dev install you would need to add:
```
# jupyter nbextension install --py --symlink --sys-prefix bqplot
$ jupyter nbextension enable --py --sys-prefix bqplot
```

rosteen:
Ah, you're correct - I totally forgot to run the nbextension commands. Working after I ran those.

martinRenou (approved):
LGTM :+1:

PR #1198, opened 2020-08-31
Updates jupyterlab installation instructions.

arokem:
Includes the `jupyter lab build` step. See: https://github.com/bqplot/bqplot/issues/885#issuecomment-683528153

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,7 @@ For the experimental JupyterLab extension, install the Python package, make sure
 $ pip install bqplot
 $ jupyter labextension install @jupyter-widgets/jupyterlab-manager # install the Jupyter widgets extension
 $ jupyter labextension install bqplot
+$ jupyter lab build
 ```
 
 ### Loading `bqplot`


README.md
```diff
@@ -91,6 +91,7 @@ For the experimental JupyterLab extension, install the Python package, make sure
 $ pip install bqplot
 $ jupyter labextension install @jupyter-widgets/jupyterlab-manager # install the Jupyter widgets extension
 $ jupyter labextension install bqplot
+$ jupyter lab build
```
martinRenou, line 94: This should not be needed as `jupyter labextension install bqplot` did build Jupyter Lab. I wonder why it's needed in Binder, maybe they disabled building jupyterlab by default.

martinRenou (changes requested):

README.md
arokem: Gotcha. FYI: I am seeing this in another setting where repo2docker is used, so this might be (slightly) more general than Binder.

martinRenou: That's odd :confused:. @jtpio @ibdafna do you know why this is needed?

jtpio: > This should not be needed as `jupyter labextension install bqplot` did build Jupyter Lab.

Indeed it looks like it should not be needed. And the 2 calls to `jupyter labextension install` could be grouped together.

ibdafna:
I believe Ariel is on a repo with overall older packages including
JupyterLab. That might be the culprit. Ariel, can you try with a fresh
local installation and then use that configuration for a new Binder-enabled
repo?

arokem:
Thanks! This all makes sense to me. Unfortunately, I am working on a shared deployment with other users, and now is not a good time to change these things under the hood. But I understand that this will probably resolve itself when we are able to upgrade jupyterlab. So, I think that you can close the issue. Thanks again for all your help.

martinRenou:
With JupyterLab 3 and the federated extension system, this should not be needed anymore

PR #1200, opened 2020-09-01
fix: margin_update event listeners were never detached

maartenbreddels:
I noticed that in many cases we listen to margin_update (using .on), but never detach it.

In all cases, the context was this, so we could replace it using this.listenTo, which will automatically detach after calling .remove().


martinRenou (approved):
:+1:

Issue #1180, opened 2020-07-28
Selections only working in initially visible tab/accordion containers


mgrabmann:
I have a similar problem to the one reported in [#1085 ](https://github.com/bqplot/bqplot/issues/1085). 
The fix from there doesn't seem to work in all cases.

Here is an example:

```
import bqplot as bq
import numpy as np
from ipywidgets import *
from bqplot import pyplot as plt 

class HistTest():
    def __init__(self):
        self.scale_x = LinearScale()
        self.scale_y = LinearScale()
        self.hist_tmp = plt.hist(
            np.random.randn(100),
            scales={'sample': self.scale_x, 'count': self.scale_y},
        )
        self.sel_range = bq.interacts.BrushIntervalSelector(
            scale=self.scale_x,
        )
        self.fig_hist = bq.Figure(
            marks=[self.hist_tmp],
            interaction=self.sel_range
        )
            
    def show(self):
        return self.fig_hist
    
fig1 = HistTest()
fig2 = HistTest()
tabs = widgets.Tab(children=[fig1.show(), fig2.show()])
tabs.selected_index=0
tabs.set_title(0,'Selector works')
tabs.set_title(1,"Selector doesn't work")
tabs
```

ChakriCherukuri:
Like plotly and dashboards from other plotting liraries, bqplot should not be limited to the narrow output cells of notebooks. Using voila we can show the bqplot dashboards on full screen. Some of the examples I added are meant for that purpose.

samwelborn:
This is happening in my app using `Accordion`s.

I mentioned an issue [here](https://github.com/glue-viz/bqplot-image-gl/issues/79) about a figure not showing up immediately, and me having to resize my screen to see it.

I demonstrate this in the following 3 videos, along with the selectors not working when in accordions when they are not initially selected as mentioned above.

The first time I create the image viewer app that I made, it is already in an open accordion. While I have to resize my screen to see it, it still functions as desired:
![1](https://user-images.githubusercontent.com/55868530/151507796-4cb84b57-6420-48c3-82cd-7e03c30b6b50.gif)

This only happens with initially unopened accordions/tabs, as you can see here. I have referenced this image viewer in another section of my app and intentionally set `selected_index=0`:
![2](https://user-images.githubusercontent.com/55868530/151507939-776b3b47-ddb2-48f9-b85c-2d5c51a4692e.gif)

[...]

PR #1202, opened 2020-09-04
BrushIntervalSelector

martinRenou:
Fix https://github.com/bqplot/bqplot/issues/1180

See https://github.com/d3/d3-brush/issues/43

### DIFF
--- a/js/src/BrushSelector.ts
+++ b/js/src/BrushSelector.ts
@@ -419,7 +419,6 @@ export class BrushIntervalSelector extends BrushMixinXSelector {
             return;
         } else {
             const extent = [selected[0], selected[1]];
-            this.brush.move(this.d3el, extent);
             const pixel_extent = extent.map(this.scale.scale).sort(
                 (a: number, b: number) => a - b);
             this.update_mark_selected(pixel_extent);
@@ -445,14 +444,15 @@ export class BrushIntervalSelector extends BrushMixinXSelector {
             const range = this.model.get("selected").map(this.scale.scale).sort(
                 function(a, b) { return a - b; });
 
+            // Move and redraw the brush selector, preventing move events to be triggered
             this.ignoreBrushEvents = true;
             try {
                 this.brush.move(this.d3el, range);
+                this.brushsel = this.d3el.call(this.brush);
             } finally {
                 this.ignoreBrushEvents = false;
             }
         }
-        this.brushsel = this.d3el.call(this.brush);
     }
 
     reset() { }
--- a/js/src/BrushSelector.ts
+++ b/js/src/BrushSelector.ts
@@ -419,7 +419,7 @@ export class BrushIntervalSelector extends BrushMixinXSelector {
             return;
         } else {
             const extent = [selected[0], selected[1]];
-            const pixel_extent = extent.map(this.scale.scale).sort(
+            const pixel_extent = extent.map((v) => this.scale.offset + this.scale.scale(v)).sort(
                 (a: number, b: number) => a - b);
             this.update_mark_selected(pixel_extent);
         }
@@ -444,7 +444,7 @@ export class BrushIntervalSelector extends BrushMixinXSelector {
         this.ignoreBrushEvents = true;
         try {
             if (this.model.get("selected")) {
-                const range = this.model.get("selected").map(this.scale.scale).sort(
+                const range = this.model.get("selected").map((v) => this.scale.offset + this.scale.scale(v)).sort(
                     function(a, b) { return a - b; });
 
                 this.brush.move(this.d3el, range);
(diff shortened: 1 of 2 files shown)


maartenbreddels:
See also https://github.com/bqplot/bqplot/pull/1097

martinRenou:
@kaiayoung is it the issue you were mentioning?

maartenbreddels:
This looks similar to what I've done in #1097, although that also contains another change extend->move.

Could you dig into why that was needed?

martinRenou:
I am actually not sure this is needed, because the move is done later on in `syncModelToBrush`

js/src/BrushSelector.ts
```diff
@@ -317,7 +317,6 @@ export class BrushSelector extends BrushMixinXYSelector {
 
         this.brush.extent([[0, 0], [this.width, this.height]]);
         this.syncModelToBrush();
-        this.brushsel = this.d3el.call(this.brush);
```
martinRenou, line 320: As discussed in https://github.com/bqplot/bqplot/pull/1097

martinRenou:
@maartenbreddels do you think it's ready for merging?

ibdafna:
@martinRenou The plot renders in the second tab, but it looks like the bouncing is still there unfortunately. 

bar_chart_selector_bouncing

```python
from bqplot import (OrdinalScale, LinearScale, Bars, Figure, Axis, ColorScale, CATEGORY10)
from bqplot.interacts import BrushIntervalSelector
import ipywidgets as widgets
import bqplot as bqp
import pandas as pd
import numpy as np
np.random.seed(0)
plt_x = np.arange(10)
plt_y = np.random.randint(10, size=10)

def print_val(change):
    print(change['new'])
        
x_sc = OrdinalScale()
y_sc = LinearScale()

plot = Bars(x=plt_x, y=plt_y, scales={'x':x_sc, 'y':y_sc}, colors=['#ff5500'],
            labels=['Date', 'Reads'])

xax = Axis(scale=x_sc, grid_lines='none', num_ticks=10)
yax = Axis(scale=y_sc, tick_format=',d', grid_lines='none', orientation='vertical', min=0)

brushsel_date = BrushIntervalSelector(scale=x_sc, marks=[plot], color='skyblue')

plot.observe(print_val, names=['selected'])

fig = Figure(marks=[plot], axes=[xax, yax], 
                              padding_x=0.025, padding_y=0.025,
                             interaction=brushsel_date)


tab = widgets.Tab()
tab.children = [fig, fig]
# widgets.VBox([fig, fig])
tab
```

martinRenou:
This looks like another issue doesn't it? This PR fixes the resize of the BrushSelector https://github.com/bqplot/bqplot/issues/1180

I am looking at the bouncing between views issue, seems very related to https://github.com/bqplot/bqplot/pull/1098, there is so much code duplication in this library I would not be surprised we simply need to apply the same fix at another class.

ibdafna:
They're closely related, I think. If you look at #1180, the title says it is only visible initially but disappears when used in a tab or accordion, which arguably includes what we're seeing above. Originally the issue was with the selection only showing once as you say due to invalid figure size. But the current result is probably worse as you don't even end up with a single selection due to the bouncing. Happy to open a separate PR if needed!

martinRenou:
Ok, now I see why they are both related. Thanks.

I'm investigating a good way to fix this.

martinRenou:
I feel like the bouncing is due to the `this.brushsel = this.d3el.call(this.brush);` line which triggers `brush_start` and `brush_end`. This line is called when the selection changes for updating the brush position, and the fact that it triggers those events it might update the selection as well.

martinRenou:
There are multiple issues with the BrushIntervalSelector (some of them appear on the BrushSelector as well, but let's deal with them one at a time):

- [x] Updating the selector when a change appeared in another view/in the model needs to be done my calling `this.brush.move` and `this.d3el.call(this.brush)`. This results in triggering `brush_start` and `brush_move`, which itself will update the model and redraw the brush -> lots of bouncing. *This PR fixes this issue by discarding the events during the selector redraw*
- [x] The `selected` attribute was not using TypedArrays, resulting in bouncing between the Python model and the JavaScript one.
- [ ] The first `relayout` in a tab does not seem to work, which is very odd because `relayout` appears to work properly on window resizing.
- [ ] BrushIntervalSelector with an ordinal scale does not work. It's either coming from issues in the OrdinalScale itself, either coming from how the BrushIntervalSelector uses the scale, or it's coming from the `selected` attribute which does not have the right TypedArray

In this PR I use a `Float32Array` for the `selected` attribute, this seems like an odd choice to me, as we might not know the type of the buffer in advance, it depends on the scale or the data I guess.

martinRenou:
We'll need to apply similar fixes to the BrushSelector, and maybe the MultiSelector

ibdafna:
LGTM! Thank you

PR #1203, opened 2020-09-04
Add flake8 check and run Python tests on Github Actions

martinRenou:
See https://github.com/bqplot/bqplot/issues/1185


martinRenou:
@ibdafna should we merge this?

PR #1214, opened 2020-10-02
Refactor BrushSelector

martinRenou:
I looked for having a mixin class to reduce the code duplication in this file, unfortunately it seems impossible at the moment when base classes are abstract https://github.com/microsoft/TypeScript/issues/29653

### DIFF
--- a/js/less/bqplot.less
+++ b/js/less/bqplot.less
@@ -74,7 +74,7 @@
 /*
  * gg and classic bqplot themes differ in the styling of the svg-background.
  *
- * plot area background fill: 
+ * plot area background fill:
  *   - gg: contrasted with the jupyterlab background. var(--jp-layout-color2);
  *   - classic: less contrasted: var(--jp-layout-color1);
  *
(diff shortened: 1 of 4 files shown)


martinRenou:
Syncing the views is not implemented in MultiSelector. We should always construct the same number of brushes in each view. Let's do this in another PR.

martinRenou:
This PR fixes multiple issues on the MultiSelector, some similar to https://github.com/bqplot/bqplot/pull/1202.

- Fix filling of the selectors (the MultiSelector used to render the entire draw area in blue since 0.12.0)
- Fix labels rendering
- Fix labels positioning on relayout
- Fix relayout of the MultiSelector
- Code factorizing and cleaning

It also fixes events bouncing in the BrushSelector

martinRenou:
@ibdafna we discussed including this PR in the release during last call. It fixes quite a few problems on the selectors.

maartenbreddels:
Excellent work!

martinRenou:
Thanks!

PR #1222, opened 2020-10-09
Add support for Jupyterlab 3

martinRenou:


jtpio:
I think we also need this part in the `__init__.py` file:

https://github.com/jupyterlab/extension-cookiecutter-ts/blob/d8c08687f8661545a65d65ab8eb9433267a45b35/%7B%7Bcookiecutter.python_name%7D%7D/%7B%7Bcookiecutter.python_name%7D%7D/__init__.py#L9-L16

martinRenou:
@jtpio Oh ok, what is this for? Isn't JupyterLab able to automatically find the extensions installed under `/share/jupyter/labextensions/`?

martinRenou:
@jtpio let's get this one in first #1223

martinRenou:
This seems to work perfectly locally. Let's wait for a JupyterLab 3 before merging this and updating the CI.

maartenbreddels:
Hmm, all of the sudden the directory layout is different (js files are in a static subdir):
```
(jlab3f) mbp ~/src/bqplot (jlab3 ✗)$ ls share/jupyter/labextensions/bqplot/                                                                                                                                                 
package.json static

```

martinRenou:
Yes this has changed in one of the release candidates. The setup.py must be changed:

```python
('share/jupyter/labextensions/bqplot', 'share/jupyter/labextensions/bqplot/', '**'),
```

maartenbreddels:
ah, that's prettier! let me change that.

maartenbreddels:
`$ python setup.py bdist_wheel`

```
copying share/jupyter/labextensions/bqplot/static/536.9b1aaa73e0be13c207d2.js.LICENSE.txt -> build/bdist.macosx-10.9-x86_64/wheel/bqplot-0.12.19.data/data/share/jupyter/labextensions/bqplot/tatic
```

tatic???? 🤯

maartenbreddels:
I think this PR is orange, we get a:

```
ImportError: Error importing plugin "pytest_tornasync": No module named 'pytest_tornasync'
```
Which if I'm not mistaking is https://github.com/jupyter-server/jupyter_server/issues/337 
this should already be released (in 1.0.6), while we install 1.0.7 🤷

maartenbreddels:
Ok, GitHub UI shows https://github.com/jupyter-server/jupyter_server/commits/master in a weird order (by some date), that makes it look like https://github.com/jupyter-server/jupyter_server/issues/337 was merged before the 1.0.6 release, but it is not :( It is unreleased

js/package.json
```diff
@@ -76,13 +78,23 @@
...
+    "extension": "lib/jupyterlab-plugin",
+    "outputDir": "../share/jupyter/labextensions/bqplot",
+    "sharedPackages": {
+      "@jupyter-widgets/base": {
+        "bundled": false,
+        "singleton": true
+      },
+      "threejs": {
+        "bundled": true
+      }
```
jasongrout, line 91:
```suggestion
      }
```
The default is to bundle dependencies, so this is not needed, and the npm package name is `three`, so this isn't doing anything anyway.

vidartf: Maybe this should be `"three": { "singleton": true }` ?

jasongrout: If multiple extensions each provide a copy of three, setting the singleton to true just makes sure they all use the same version (which would be the highest version), even if is the wrong version.

I think it's fine to just not put anything, and let the system take care of version conflicts (i.e., things will be shared when possible, but if not, then multiple versions can coexist) - unless it is really important to share the same module even if it is the wrong version. Or if three itself cannot handle having multiple versions on a page.

vidartf: If multiple extensions need to pass references to three objects to each other, then it is important that they are shared, but I guess that should be managed via semver then, since other extensions might want to use a different version of threejs independently of what the other two are doing.

jasongrout: FYI, here is my analysis of how these options interplay: https://github.com/jupyterlab/jupyterlab/issues/9496#issuecomment-749381599

pyproject.toml
```diff
@@ -1,3 +1,3 @@
 [build-system]
-requires = ["jupyter_packaging~=0.7.7", "setuptools>=40.8.0", "wheel"]
+requires = ["jupyter_packaging~=0.7.7", "jupyterlab>=3.0.0rc4,==3.*", "setuptools>=40.8.0", "wheel"]
```
jtpio, line 2: It should now be possible to update to:

```suggestion
requires = ["jupyter_packaging~=0.7.9", "jupyterlab~=3.0", "setuptools>=40.8.0", "wheel"]
```

maartenbreddels:
Closes #1245

README.md
```diff
@@ -64,7 +64,7 @@ Using conda
 $ conda install -c conda-forge bqplot
 ```
 
-To enable bqplot with Jupyter lab:
+To enable bqplot with Jupyter lab (only needed for version < 3):
```
jasongrout, line 67:
```suggestion
To enable bqplot for JupyterLab version 1 or 2 (not needed for JupyterLab 3.0):
```

js/package.json
```diff
@@ -30,6 +31,7 @@
   },
   "devDependencies": {
     "@jupyter-widgets/controls": "^1.5.0 || ^2",
+    "@jupyterlab/builder": "^3.0.0-rc.4",
```
jasongrout, line 34:
```suggestion
    "@jupyterlab/builder": "^3.0.0",
```

js/package.json
```diff
@@ -30,6 +31,7 @@
   },
   "devDependencies": {
     "@jupyter-widgets/controls": "^1.5.0 || ^2",
```
jasongrout, line 33:
```suggestion
    "@jupyter-widgets/controls": "^1.5.0 || ^2 || ^3",
```

js/package.json
```diff
@@ -65,7 +67,7 @@
     "webpack-cli": "^3.3.12"
   },
   "dependencies": {
-    "@jupyter-widgets/base": "^2 || ^3",
+    "@jupyter-widgets/base": "^2 || ^3 || ^4.0.0-beta.1",
```
jasongrout, line 70:
```suggestion
    "@jupyter-widgets/base": "^2 || ^3 || ^4",
```

test-environment.yml
```diff
@@ -1,14 +1,14 @@
...
     - conda-forge
-    - defaults
 dependencies:
     - python
     - ipywidgets >=5.2.2
     - traitlets >=4.3.0
     - traittypes >=0.0.6
     - numpy >=1.10.4
     - pandas
+    - jupyterlab=3.0.0  # to build the lab federated bundle
```
jasongrout, line 11: Should we relax the pin?

```suggestion
    - jupyterlab=3  # to build the lab federated bundle
```

martinRenou:
@jasongrout How should we proceed concerning the `processPhophorMessage` method?

jasongrout:
> @jasongrout How should we proceed concerning the `processPhophorMessage` method?

Leave it. We left the phosphor->Lumino bridges in jlab 3. The hope is that we release ipywidgets 8 early this year, and that will be the migration to Lumino.

martinRenou:
For some reason, the CSS is not included anymore

jtpio:
Should the `jstargets` also include an entry pointing to the prebuilt assets for the lab extension? Similar to:

https://github.com/jupyterlab/extension-cookiecutter-ts/blob/4d0b82636eb72da1f7260536944823c3a3f075d0/%7B%7Bcookiecutter.python_name%7D%7D/setup.py#L25-L27

martinRenou:
This **should** be good. Trying one last time locally.

PR #1230, opened 2020-10-21
fix: install map_data

maartenbreddels:
Fixes #1228
Note that this only matters for the source distribution, the wheel seems to do fine. But I don't understand why the conda-forge distribution *does* include the json files.
@martinRenou maybe you can try to confirm the difference between this branch and master.
(Note that the files are include in the sdist, but a pip install will not put them in the site-packages/bqplot/data_data directory)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -96,6 +96,7 @@ setup_args = dict(
     license='Apache',
     author='The BQplot Development Team',
     url='https://github.com/bloomberg/bqplot',
+    package_data={'bqplot.map_data': ['*.json']},
     include_package_data=True,
     cmdclass=cmdclass,
     install_requires=[


SylvainCorlay:
@ibdafna and @ChakriCherukuri said in the weekly meeting that we should not go ahead. These files are used.

maartenbreddels:
Note that this PR fixes the sdist to include those, not remove it.

martinRenou:
I guess there was a confusion with the other PR that removes source map files.

maartenbreddels:
I'm still looking into why this fix is needed. The wheel seems fine but the sdist does include the data (due to the manifest.in file), but it's not being installed. The fact that on conda forge it does get installed, might indicate a version issue or my misunderstanding.

maartenbreddels:
Ok, I'm clueless now why we need this PR, From master `static` does get installed, but the `map_data` does not, can you confirm this for me @martinRenou ?

```
# from master
$ rm -rf dist build MANIFEST bqplot.egg-info  
$ python setup.py -v sdist 
$ pip install -v  dist/bqplot-0.12.18.tar.gz 
$ ls $CONDA_PREFIX/lib/python3.7/site-packages/bqplot
__init__.py        _version.py        colorschemes.py    figure.py          interacts.py       marks.py           scales.py          toolbar.py
__pycache__        axes.py            default_tooltip.py install.py         market_map.py      pyplot.py          static             traits.py
```

maartenbreddels:
Since conda-forge installs using `pip install .`, which first builds a wheel which it then installs, it does pick up the map_data:

```
$ pip install -v .
$ ls $CONDA_PREFIX/lib/python3.7/site-packages/bqplot                                                                                                                            dev
__init__.py        _version.py        colorschemes.py    figure.py          interacts.py       market_map.py      pyplot.py          static             traits.py
__pycache__        axes.py            default_tooltip.py install.py         map_data           marks.py           scales.py          toolbar.py
```

Conclusion: `pip install <source-package>.tar.gz` is broken/different from `pip install .` and `pip install <name>.whl`. This was always the case I think, but nobody installs from source packages it seems.

PR #1232, opened 2020-10-28
fix: marks were not removed when changed before views created

maartenbreddels:
Closes https://github.com/spacetelescope/jdaviz/issues/270

```python
from bqplot import *
from ipywidgets import *
from bqplot.interacts import *

scale_x1 = DateScale()
scale_y1 = LinearScale()
x = np.arange('2010-01-01', '2020-04-01', dtype='datetime64[D]')
y1 = np.cumsum(np.random.random(x.shape) - 0.5)
line1 = Lines(x=x, y=y1, scales={'x':scale_x1, 'y':scale_y1})
y2 = np.cumsum(np.random.random(x.shape) - 0.5)
line2 = Lines(x=x, y=y2, scales={'x':scale_x1, 'y':scale_y1})

fig1 = Figure(marks=[line1, line2], interaction=BrushSelector(x_scale=scale_x1,
                     y_scale=scale_y1, marks=[line1]), background_style={'fill':'#eee'})

fig1

# next cell, run a few times, and a line mark stays visible

fig1.marks = [line1, line2]
fig1.marks = []
```

Fix coming, would first like to see this fail on CI.

### DIFF
--- a/js/src/Figure.ts
+++ b/js/src/Figure.ts
@@ -292,7 +292,9 @@ class Figure extends widgets.DOMWidgetView {
 
     replace_dummy_nodes(views) {
         _.each(views, function(view: any) {
-            if (view.dummy_node !== null) {
+            // It could be that the dummy node is removed before we got a change to replace it
+            // This happens when the marks list is changed rapidly
+            if (view.dummy_node !== null && view.dummy_node.parentNode) {
                 view.dummy_node.parentNode.replaceChild(view.el, view.dummy_node);
                 view.dummy_node = null;
                 this.displayed.then(function() {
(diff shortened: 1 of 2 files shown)


maartenbreddels:
I've not been able to trigger the same bug in Box:
https://github.com/jupyter-widgets/ipywidgets/blob/51322341d4f6d88449f9dbf6d538c60de6d23543/packages/controls/src/widget_box.ts#L116

rosteen (approved):
I installed and tested this fix with Jdaviz and it appears to resolve the bug. I spent some time doing actions that were consistently able to trigger the bug on the master branch, and was unable to trigger the bug behavior. Looks fixed to me.

PR #1235, opened 2020-11-05
fix: source distribution with prebuild js should skip js_command

### LINKED ISSUE #1234: Source distribution is broken.
To reproduce:
```
$ python setup.py -v sdist  
$ pip install -v  dist/bqplot-0.12.18.tar.gz
....
running bdist_wheel
  running jsdeps
  Installing build dependencies with npm.  This may take a while...
  > npm install
  error: [Errno 2] No such file or directory: '/private/var/folders/4h/rklzvjwj08170p1nlsntfdkw0000gn/T/pip-req-build-8ge5c6k1/js': '/private/var/folders/4h/rklzvjwj08170p1nlsntfdkw0000gn/T/pip-req-build-8ge5c6k1/js'
  Building wheel for bqplot (PEP 517) ... error
  ERROR: Failed building wheel for bqplot
....
```

This seems to be caused by 0467ef1e901b5d50bfb850639bcde06172e634fe where we now rely on jupyter_packaging. However, we first checked if all our targets already existed, and this feature does not seem to be present in jupyter_packing.

Solution:
I propose we revert 0467ef1e901b5d50bfb850639bcde06172e634fe so we can continue doing releases, and see how we can push this in jupyter_packaging.

maartenbreddels:
Fixes #1234 
we might want to upstream this to jupyter_packing.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@
 
 from __future__ import print_function
 
-from setuptools import setup, find_packages
+from setuptools import setup, find_packages, Command
 
 from jupyter_packaging import (
     create_cmdclass,
@@ -26,6 +26,7 @@ from jupyter_packaging import (
 
 import os
 from os.path import join as pjoin
+from pathlib import Path
 from distutils import log
 
 here = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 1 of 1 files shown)


martinRenou:
I'll need to do that in other places where I added jupyter-packaging...

Thanks for finding it!

maartenbreddels:
Note that I also added `build_dir='js/lib', source_dir='js/src'` which can skip builds if possible.

martinRenou:
Or we update jupyter-packaging and release it.

maartenbreddels:
On it.

maartenbreddels:
As discussed with @martinRenou , we'll merge this first, so we can do a release.

Issue #1243, opened 2020-11-29
MarketMap becomes irresponsive if data is updated and a cell was selected


acelere:
If a MarketMap object has a cell selected and the data is updated/changed, the MarketMap becomes irresponsive.
If data is updated with no cells selected, it works fine.

Sorry for the cluttered example, but this is what I had in hand
Example code:
```
#create MarketMap
import pandas as pd
from bqplot.market_map import MarketMap
from ipywidgets import Layout

class ParameterMap:
    def __init__(self, sub_sampled_data, title, map_groups):
        self.sub_sampled_data = sub_sampled_data
        self.map_names = list(self.sub_sampled_data.columns)
        self.map_codes = [i for i in range(len(self.map_names))]
        self.map_groups = []
        for key in map_groups.keys():
            self.map_groups.append(map_groups.get(key))
        
        self.map = MarketMap(names=self.map_names, groups=self.map_groups,     
                               layout=Layout(min_width='50px', min_height='70px'),
                                 enable_hover=False, cols=3,
                            map_margin={'top':50, 'bottom':0, 'left':0, 'right':25})

        self.map.font_style = {'font-size': '10px', 'fill':'white'}
        self.map.title = title
        self.map.title_style = {'fill': 'Red'}
        
    def update_map(self, sub_sampled_data, map_groups):
        self.sub_sampled_data = sub_sampled_data
        self.map_names = list(self.sub_sampled_data.columns)
        self.map_codes = [i for i in range(len(self.map_names))]
        self.map_groups = []
        for key in map_groups.keys():
            self.map_groups.append(map_groups.get(key))
        
        self.map.names = self.map_names
        self.map.groups = self.map_groups

data_label = 'test_dat0'
data_length = 10
test_data = {}
for i in range(data_length):
    test_data[f'{data_label}_{i}'] = [1,2,3,4] 
test_data = pd.DataFrame(test_data)
sub_sampled_data = test_data
map_groups = {}
for p in sub_sampled_data.columns:
    map_groups[p] = 'FTI'
slicemap = ParameterMap(sub_sampled_data, 'Parameter Map', map_groups)
slicemap.map
```
and on another notebook cell:

```
#update the MarketMap data
data_label = 'test_dat1'
data_length = 10
test_data = {}
for i in range(data_length):
    test_data[f'{data_label}_{i}'] = [1,2,3,4] 
test_data = pd.DataFrame(test_data)
sub_sampled_data = test_data
map_groups = {}
[...]

PR #1244, opened 2020-11-29
fix: MarketMap becomes irresponsive if data is updated and a cell was selected

acelere:
*issue number of reported bug #1243 *

The update_data() function does not set the selected array to empty, which caused an error in the apply_selected() function where the selected_cell would end up with a "node null" exception.

Proposed fix sets the selected array length to zero **always** when data is updated.

*Tested with same code example documented in the referenced issue.
Results are below:

marketmap_bug_1

### DIFF
--- a/js/src/MarketMap.ts
+++ b/js/src/MarketMap.ts
@@ -185,7 +185,7 @@ export class MarketMap extends Figure {
 
     update_title(model, value) {
         this.title.text(this.model.get("title"))
-           .styles(this.model.get("title_style"));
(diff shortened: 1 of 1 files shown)


maartenbreddels:
Thanks, I've changed the code a bit, but indeed this seems like a good fix!

maartenbreddels:
Oops, I push a wrong change, which closed the PR. Had to reopen on https://github.com/bqplot/bqplot/pull/1285

PR #1250, opened 2020-12-16
Use skip_if_exists from upstream jupyter-packaging

jtpio:
Follow-up to https://github.com/bqplot/bqplot/pull/1235 and https://github.com/jupyter/jupyter-packaging/pull/54.

Using the latest `jupyter-packaging~=0.7.7` which includes `skip_if_exists`.


SylvainCorlay:
Thanks!

PR #1251, opened 2020-12-17
Fix package_data issue

martinRenou:


maartenbreddels:
I guess you forgot about https://github.com/bqplot/bqplot/pull/1230 :)

martinRenou:
Indeed yes, sorry for that.

I think we all got confused during the meetings.

maartenbreddels:
No problem, and packaging is a mess :)

PR #1252, opened 2020-12-18
Update to jupyter-packaging~=0.7.9

jtpio:
Which includes a fix for `skip_if_exists`.

Follow-up to #1250 using https://github.com/jupyter/jupyter-packaging/pull/58


jtpio:
Marking as a draft for now, as we would need to double check these targets:

https://github.com/bqplot/bqplot/blob/0620195ad4191db7b5bffab2d36e76564d435e24/setup.py#L78-L79

Are part of the sdist.

jtpio:
Update done in #1222, closing.

PR #1255, opened 2021-01-07
Add Github action for checking sdist and wheel

martinRenou:
Maybe at some point we can get rid of Travis


martinRenou:
To see the build running, check this PR: https://github.com/martinRenou/bqplot/pull/1

PR #1261, opened 2021-01-21
Fix Directed Graph Mark

ibdafna:
- [x] Implement new logic for arced line path calculation

- [x] Fix line alignment issue
graph


martinRenou:
We need to do the same for other link types

PR #1263, opened 2021-01-21
Fix for JupyterLab 2

martinRenou:
The `build` script must not build the labextension, otherwise, it makes the JS package dependent on JupyterLab 3

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 dist
 build
 eggs
+MANIFEST
 parts
 bin
 var
@@ -63,3 +64,5 @@ node_modules
 *.swp
 
 *.tsbuildinfo
+
+js/package-lock.json
--- a/js/package.json
+++ b/js/package.json
@@ -19,10 +19,11 @@
     "url": "https://github.com/bloomberg/bqplot.git"
   },
   "scripts": {
-    "build": "jlpm run build:css && jlpm run build:js && jlpm run build:labextension && webpack --mode=production",
+    "build": "jlpm run build:css && jlpm run build:js",
     "build:css": "lessc less/bqplot.less css/bqplot.css",
     "build:js": "tsc",
     "build:labextension": "jupyter labextension build .",
+    "build:extensions": "jlpm run build && jlpm run build:labextension && webpack --mode=production",
     "watch": "npm-run-all -p watch:*",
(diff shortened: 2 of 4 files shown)


jasongrout:
@martinRenou - does this point to a corresponding change in the widget cookiecutter?

jasongrout:
Is the main change here that you can build the js package with jlab 2, i.e., you can do the release with jlab 2?

If you do the release with jlab 3, do the built artifacts (and published npm package) still work with jlab 2? (Thinking about it more - I suppose this is the case?)

If this is mainly about making the release process work with jlab 2, and not about having the final package compatible with jlab 2, then probably it doesn't need to ported to the cookiecutter (I think it's fine for the cookiecutter to depend on jlab 3 for the release). However, if it is about making the published package compatible with jlab 2, then we probably ought to push changes over to the cookie cutter.

martinRenou:
> Is the main change here that you can build the js package with jlab 2, i.e., you can do the release with jlab 2?

No, this is not about making releases. It is about installing the package with JupyterLab 2 (Sorry I haven't been clearer).

Given this reported issue: https://github.com/jupyter-widgets/tutorial/issues/118#issue-787105444 I believed that installing the extension with JupyterLab 2 was broken. But after trying locally with bqplot it looks like I cannot reproduce, so maybe this fix is not needed.

martinRenou:
The matplotlib issue might have been due to the post-install script actually: https://github.com/matplotlib/ipympl/blob/0.6.2/js/package.json#L20

Sorry for the confusion. We can close this PR I guess.

jasongrout:
I also just checked with the cookiecutter, which uses similar logic, and installing into jlab 2 works just fine.

jasongrout:
> The matplotlib issue might have been due to the post-install script actually: https://github.com/matplotlib/ipympl/blob/0.6.2/js/package.json#L20

That looks like a problem, yes.

> We can close this PR I guess.

I agree.

jasongrout:
> The matplotlib issue might have been due to the post-install script actually: https://github.com/matplotlib/ipympl/blob/0.6.2/js/package.json#L20

Are you opening an issue or PR in ipympl?

martinRenou:
Yes I already removed the post install script and released, but I also applied the change this PR is making that I should maybe revert.

Issue #1256, opened 2021-01-08
Get rid of Travis?


martinRenou:
Get rid of Travis and use only Github Actions?

ibdafna:
I think that's a good idea.

martinRenou:
It looks like Travis is not triggered anyway, or at least not reporting back. I opened a PR removing it. Also added the Python and JavaScript tests to the CI as they were not run on the Github actions,

PR #1267, opened 2021-01-26
Run tests on the CI

martinRenou:
Tests were not run on Github actions. And Travis seems to not be triggered for some time now...


martinRenou:
Fix #1256

martinRenou:
We actually need to rethink JS testing in bqplot.

Due to the recent transition to Webpack 5, and due to all the test modules we rely on (karma-chai, karma-mocha) that didn't follow the webpack 5 update, this will actually be difficult. We'll need to find another way to test without those modules.

martinRenou:
I guess Galata will be the answer to this

Issue #1245, opened 2020-12-08
d3-selection-multi does not work in federated bundle when ipyvolume is installed


maartenbreddels:
Installing #1222 and https://github.com/maartenbreddels/ipyvolume/pull/355
we get: 
this.fig_background.append(...).attr(...).attr(...).attr(...).attr(...).attr(...).styles is not a function
Which should be provided by https://github.com/d3/d3-selection-multi/blob/master/index.js

For some reason the d3 instance is shared between ipyvolume and bqplot, and bqplot gets the wrong one?

Note that this issue shows up even when not using ipyvolume at all.

maartenbreddels:
@jasongrout i thought bqplot and ipyvolume cannot affect each other with the federated bundles. It it possible that bqplot gets ipyvolume's d3 version, which is not money patched by 'd3-selection-multi' ? I thought by default packages are not shared.

maartenbreddels:
By default bundled is true, maybe false will work (source https://github.com/jupyterlab/jupyterlab/issues/9496#issuecomment-749381599 )

maartenbreddels:
Fixed by https://github.com/bqplot/bqplot/pull/1222/commits/3181faf0470e621c977822d257d32f05d6ca01e8

jasongrout:
> I thought by default packages are not shared.

By default, all direct dependencies are shared. See https://jupyterlab.readthedocs.io/en/stable/extension/extension_dev.html#deduplication-of-dependencies

agoose77:
This can be closed?

maartenbreddels:
Yes, the magic line was https://github.com/bqplot/bqplot/pull/1222/files#diff-e51a40ac250c9696142466f114f754161e7e5102c0cdb5354548b757deb272f6R89

PR #1271, opened 2021-02-03
Refactor: remove d3 selection multi

maartenbreddels:
`d3-selection-multi` monkey patches `d3-selection`, this does not work well
with federated modules, since another provider may provide a semver
compatible d3 that is unpacked. This causes issues with jupyterlab in combination with other widgets/extensions that use (and thus *can* provide) d3.

Related: #1245 #1222

It seems https://github.com/bqplot/bqplot/pull/1222/files#diff-e51a40ac250c9696142466f114f754161e7e5102c0cdb5354548b757deb272f6R89 does not work at all, I may have been lucky in thinking it did before, or this is a regression in some tooling, in any case, we should get rid of this monkey patching library.

Note: there are many more uses of `.styles` and `.attrs`, opening this soon to get feedback.


js/package.json
```diff
@@ -85,8 +85,7 @@
       "@jupyter-widgets/base": {
         "bundled": false,
         "singleton": true
-      },
-      "d3": false
```
martinRenou, line 89: Do we need to remove this?

JohanMabille:
I'm totally in favor of this approach. Besides, `d3-selection-multi` has been archived and is not maintained anymore.

js/package.json
maartenbreddels: yes!

maartenbreddels: Hahah, I thought you reminder me to remove this, I misunderstood.
Yes, we can remove this because we can share d3 now.

dhirschfeld:
> It seems #1222 (files) does not work at all

Does this mean https://github.com/bqplot/bqplot/releases/tag/0.12.20 is broken on jlab3?

maartenbreddels:
It works when no other federated module provides d3, so by itself bqplot works. It might not work when others are installed.

SylvainCorlay:
Thanks! Sorry for not merging earlier!

PR #1276, opened 2021-02-16
Feature: enable deduplication of d3-selection for shared browser events

maartenbreddels:
d3-selection has global state (the event) that is obtained
using `require("d3-selection").event`. If multiple copies of this
js module exists in the frontend, they will not read eachother's
events.

This is required for 3rd party libraries, such as bqplot-image-gl
to be able to read the last d3 event.

Since Jupyter Lab will deduplicate direct dependencies, we only need
to add the d3-selection as an explicit dependency, and it will be
shared with other extensions.

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -72,6 +72,7 @@
     "@lumino/messaging": "^1",
     "@lumino/widgets": "^1",
     "d3": "^5.7.0",
+    "d3-selection": "^1",
     "is-typedarray": "^1.0.0",
     "popper.js": "^1.0.0",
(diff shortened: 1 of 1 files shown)


martinRenou:
@maartenbreddels LGTM should we merge?

maartenbreddels:
Yeah, and do a release, with the latest bqplot-image-gl, we now need this, or get in the JS console:
```
Error: Shared module d3-selection doesn't exist in shared scope default
```
Since we don't ship it anymore, and bqplot is a provider of it.

PR #1279, opened 2021-02-18
Improve pyplot UX

martinRenou:
Improve pyplot API to be closer to Matplotlib

cc. @SylvainCorlay @ibdafna 

pyplot


martinRenou:
Let's not merge this. As discussed in-person it might not be a good approach. We should maybe rename the pyplot API to remove the confusion with Matplotlib.

PR #1282, opened 2021-02-19
Add TypeScript typing

davidbrochart:


js/src/MarketMap.ts
```diff
@@ -643,7 +644,8 @@ export class MarketMap extends Figure {
     hide_tooltip() {
          this.tooltip_div.style("pointer-events", "none");
          this.tooltip_div.transition()
-            .styles({"opacity": 0, "display": "none"});
+            .style("opacity", 0)
+            .style("opacity", 0);
```
martinRenou, line 648:
```suggestion
            .style("display", "none");
```

davidbrochart: Oops, thanks!

js/src/Axis.ts
```diff
@@ -119,11 +121,13 @@ export class Axis extends WidgetView {
                 const numDateTicks = (this.width < DATESCALE_WIDTH_THRESHOLD) ?
                                         5 :
                                         undefined;
-                this.axis.tickValues(this.axis_scale.scale.ticks(numDateTicks));
+                const scale = this.axis_scale.scale as d3.ScaleTime<Date, number>;
```
martinRenou, line 124: For scales, we won't need these `as ...`. This is handle by this PR https://github.com/bqplot/bqplot/pull/904

js/src/Selector.ts
```diff
@@ -96,7 +97,7 @@ export abstract class BaseXSelector extends BaseSelector {
         if(this.model.get("scale")) {
             const that = this;
             return this.create_child_view(this.model.get("scale")).then(function(view) {
-                that.scale = view;
+                that.scale = view as WidgetView as LinearScale;
```
marimeireles, line 100: You can have double `as`?
Are you qualifying a `WidgetView` as a `LinearScale` here?

marimeireles (commented):
Hey @davidbrochart !
As I said in the meeting this is more like me trying to educate myself in ts and bqplot, hope I was able to make some pertinent questions though.
I tested a bunch of different examples with your code and everything seems to work fine :)

js/src/Axis.ts
```diff
@@ -150,12 +154,15 @@ export class Axis extends WidgetView {
                     this.axis.tickValues(useticks);
                 }
             } else {
-                this.axis.tickValues(this.axis_scale.scale.ticks());
+                const scale = this.axis_scale.scale as (d3.ScaleLinear<number, number> |
```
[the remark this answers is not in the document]
davidbrochart: But the previous `if` branches eliminate "log", "date" and "ordinal". Maybe we should not include `d3.ScaleLogarithmic<number, number>` BTW.

js/src/Axis.ts
```diff
@@ -667,21 +674,21 @@ export class Axis extends WidgetView {
         }
         const scale_range = this.axis_scale.scale.domain();
         const max_index = (this.axis_scale.scale.domain().length - 1);
-        step = (scale_range[max_index] - scale_range[0]) / (num_ticks - 1);
+        step = ((scale_range[max_index] as any) - (scale_range[0] as any)) / (num_ticks - 1);
```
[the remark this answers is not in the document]
davidbrochart: I think the domain of a scale can be of many types, but I might be wrong.

js/src/Map.ts
```diff
@@ -411,15 +411,15 @@ export class Map extends Mark {
...
+    map: d3.Selection<SVGSVGElement, any, any, any>;
+    map_id: string;
+    width: number;
+    height: number;
+    enable_hover: boolean;
+    stroke_g: d3.Selection<SVGGElement, any, any, any>;
+    fill_g: d3.Selection<SVGGElement, any, any, any>;
+    highlight_g: d3.Selection<SVGGElement, any, any, any>;
+    transformed_g: d3.Selection<SVGGElement, any, any, any>;
+    zoom: d3.ZoomBehavior<Element, any>;
```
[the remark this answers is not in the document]
davidbrochart: There is: https://github.com/bqplot/bqplot/blob/f81dbf099bf275c5317044fc8b1efdd291733a14/js/src/Map.ts#L414

js/src/MarketMap.ts
```diff
@@ -71,7 +74,6 @@ export class MarketMap extends Figure {
         this.fig_click.classed("g-click", true);
         this.fig_hover.classed("g-hover", true);
         this.fig_names.classed("g-names", true);
-        this.axis = [];
```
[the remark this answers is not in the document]
davidbrochart: IIRC it is not used.

js/src/Selector.ts
davidbrochart: Yes, casting `view` to `LinearScale` is not accepted because the objects are too different if I understand correctly, so you need to cast `view` to `WidgetView` first, then `WidgetView` to `LinearScale`.

js/src/LinearScale.ts
```diff
@@ -15,11 +15,12 @@
 
 import * as d3Scale from 'd3-scale';
 import { Scale } from './Scale';
+import { LinearScaleModel } from './LinearScaleModel';
 
 export class LinearScale extends Scale {
 
     render() {
-        this.scale = d3Scale.scaleLinear();
+        this.scale = d3Scale.scaleLinear() as d3.ScaleLinear<number, number>;
```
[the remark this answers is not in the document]
davidbrochart: To be honest, I don't know.

js/src/ColorScale.ts
```diff
@@ -52,7 +52,8 @@ export class ColorScale extends Scale {
         this.trigger("color_scale_range_changed");
     }
 
+    scale: d3.ScaleLinear<number, number> |
```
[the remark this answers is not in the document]
davidbrochart: It's good to know that the examples confirm these types, so maybe the docs are wrong.

js/src/Selector.ts
martinRenou: You can specify to `create_child_view` all the right types, something like:

```
return this.create_child_view<LinearScaleModel>(this.model.get("scale")).then(function(view: LinearScale) {
```

davidbrochart: Good to know, thanks!

js/src/Axis.ts
```diff
@@ -57,7 +59,7 @@ export class Axis extends WidgetView {
         this.listenTo(this.model, "change:scale", (model, value) => {
             this.update_scale(model.previous("scale"), value);
             // TODO: rescale_axis does too many things. Decompose
-            this.axis.scale(this.axis_scale.scale); // TODO: this is in redraw_axisline
+            this.axis.scale(this.axis_scale.scale as d3.AxisScale<d3.AxisDomain>); // TODO: this is in redraw_axisline
```
martinRenou, line 62: Instead of doing this, I guess we want to define `axis_scale` as being a `Scale` widget view, then define `scale` as being a `d3.AxisScale<d3.AxisDomain>` in the `Scale` widget implementation. Then this would work without the need of casting.

martinRenou (changes requested):

js/src/Axis.ts
martinRenou: My comment does not make any sense. Sorry for that. Not sure how we could get rid of this cast!

PR #1283, opened 2021-02-19
Karma to Jest

martinRenou:


martinRenou:
Just testing on the CI now. It's not ready yet.

martinRenou:
It fails with `require('raw-loader!../shaders/scales.glsl')`. I guess we need to mock it.

martinRenou:
From my experience on this PR, Jest is no easier than Karma. Now that Karma-webpack has been updated we can use it #1289.

PR #1285, opened 2021-02-22
Fix market map styles

maartenbreddels:
Re-open of #1244


acelere:
Thanks @maartenbreddels

PR #1289, opened 2021-03-02
Fix Karma tests and run them on the CI

martinRenou:


martinRenou:
Merging, as there is no critical decision to make on this.

PR #1291, opened 2021-03-02
Refactor the way we load shaders

martinRenou:


maartenbreddels:
Yes, this is fantastic, this was really painful to see in the code :)

martinRenou:
Yep! It has been problematic when I tried testing with Jest. Not such a blocker otherwise, but I also think it's nicer that way :)

martinRenou:
CI failure is a network issue

PR #1297, opened 2021-03-11
Use nbval on the CI

martinRenou:
This would have helped catch this issue: https://github.com/bqplot/bqplot/pull/1296


martinRenou:
Note that I run nbval on `examples/Marks`, because some of the more complicated examples are fetching data online, which could lead to unwanted network errors on the CI.

PR #1299, opened 2021-03-16
Line typing

marimeireles:


js/src/Mark.ts
```diff
@@ -197,13 +197,15 @@ export abstract class Mark extends widgets.WidgetView {
         this.d3el.style("display", visible ? "inline" : "none");
     }
 
-    get_colors(index) {
+    //how do I know if smth is a color attr or a data attr, this changes the type
+    //cause it could be a numpy array
+    get_colors(index): string {
```
marimeireles, line 202: I'm not super sure about this type. It comes from `color` from here: https://github.com/bqplot/bqplot/blob/5595c241093d6f300eac76795d03217c625922bc/bqplot/marks.py#L368, but I'm confused... It just says that's an array.
Maybe it's even just a `: Color` type, because we're getting it from an array of colors? (https://github.com/bqplot/bqplot/blob/5595c241093d6f300eac76795d03217c625922bc/bqplot/marks.py#L304)

marimeireles (commented):
Hey @martinRenou, finally ready :)
Thank you!

marimeireles:
I think I should have ran the tests on a different version of the file, sorry for that, so ignore my ping haha
Tomorrow morning I'll fix it!

js/src/Lines.ts
```diff
@@ -592,74 +591,76 @@ export class Lines extends Mark {
...
-            if (this.legend_el) {
-                this.legend_el.select(".dot").attr("d", this.dot.size(0));
+            if (this.legendEl) {
+                this.legendEl.select(".dot").attr("d", this.dot.size(0));
             }
         }
     }
 
-    update_marker_size(model, marker_size) {
+    //removed unused var model
```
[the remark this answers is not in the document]
martinRenou: Nope, that will not work :) This method is used by `this.listenTo(this.model, "change:marker_size", this.update_marker_size);` and passes the `model` as first argument. I suppose this is not tested well.

js/src/Mark.ts
martinRenou: There are two parameters: `color` and `colors`.
- `color` is an array of numbers, and we use the color scale to map this number to an actual color string. To give you an example, this is useful when you want to map some data to a color (like mapping a stock price to a bar color in a bar chart).
- `colors` is a list of string (CSS colors). Those are the "default" colors that are used either when the color data is NaN or when there are no color data (if `color` is None).

You can find an example of this in the `examples/Marks/Pyplot/Lines.ipynb` Notebook, section "Coloring Lines according to data".

This is not true for all the marks, but it is at least true for Lines, Bar, Scatter and I think PieChart.

So basically you are right, the return type is a string here.
You can also type `index` as being a `number` here: `get_colors(index: number): string {`

js/src/Mark.ts
```diff
@@ -197,13 +197,15 @@ export abstract class Mark extends widgets.WidgetView {
         this.d3el.style("display", visible ? "inline" : "none");
     }
 
-    get_colors(index) {
+    //how do I know if smth is a color attr or a data attr, this changes the type
```
martinRenou, line 200: Maybe remove this comment

js/src/Lines.ts
```diff
@@ -592,74 +591,76 @@ export class Lines extends Mark {
         }
     }
 
-    update_dots_xy(animate) {
+    update_dots_xy(animate): voic {
```
martinRenou, line 594: void?

js/src/Lines.ts
```diff
@@ -592,74 +591,76 @@ export class Lines extends Mark {
...
         if (this.model.get("marker")) {
-            const x_scale = this.scales.x, y_scale = this.scales.y;
-            const animation_duration = animate === true ? this.parent.model.get("animation_duration") : 0;
+            const xScale = this.scales.x, yScale = this.scales.y;
+            const animationDuration = animate === true ? this.parent.model.get("animationDuration") : 0;
```
martinRenou, line 597:
```suggestion
            const animationDuration = animate === true ? this.parent.model.get("animation_duration") : 0;
```

js/src/Lines.ts
```diff
@@ -413,143 +413,142 @@ export class Lines extends Mark {
...
           })
-          .duration(animation_duration);
+          .duration(animationDuration);
 
-        curves_sel.select(".area")
+        curvesSel.select(".area")
           .transition("update_line_xy")
-          .attr("d", function(d: any, i) {
-            return that.area(d.values);
+          .attr("d", function(d, i) {
```
martinRenou, line 494:
```suggestion
          .attr("d", d => {
```

js/src/Lines.ts
```diff
@@ -163,173 +164,172 @@ export class Lines extends Mark {
...
 
-    relayout() {
+    relayout(): void {
         this.set_ranges();
         this.update_line_xy(false);
     }
 
-    selector_changed(point_selector, rect_selector) {
-        if(point_selector === undefined) {
+    //erased rect_selector unused var
```
martinRenou, line 264: Maybe remove this comment

martinRenou (changes requested):

js/src/Mark.ts
marimeireles: Yeah, I'll remove all of them, it was just me writing along the way so I could make questions in the review :)

marimeireles:
hey @martinRenou, thanks for the help! So, as I said, a couple of weird bugs. One of them is this kind: 
```
    src/Lines.ts(256,42): error TS2769: No overload matches this call.
      Overload 1 of 2, '(data: [number, number][]): string', gave the following error.
        Argument of type '[number[]]' is not assignable to parameter of type '[number, number][]'.
          Type 'number[]' is missing the following properties from type '[number, number]': 0, 1
      Overload 2 of 2, '(data: [number, number][]): void', gave the following error.
        Argument of type '[number[]]' is not assignable to parameter of type '[number, number][]'.
```
Complaining about the type of `legendEl`. I'm using the same type as in the master branch and I'm having this issue... Which means it's not the type that's the problem.

It's happening in other locations too:
```
    src/Lines.ts(348,40): error TS2769: No overload matches this call.
      Overload 1 of 2, '(data: [number, number][]): string', gave the following error.
        Argument of type '[number[]]' is not assignable to parameter of type '[number, number][]'.
      Overload 2 of 2, '(data: [number, number][]): void', gave the following error.
        Argument of type '[number[]]' is not assignable to parameter of type '[number, number][]'.
```
Any clue?

js/src/Lines.ts
```diff
@@ -670,16 +671,16 @@ export class Lines extends Mark {
     }
 
     dot: any;
-    legend_el: d3.Selection<SVGGElement, any, any, any>;
-    legend_line: d3.Line<[number, number]>;
-    legend_path_data: [number, number][];
+    legendEl: d3.Selection<any, any, any, any>;
+    legendLine: d3.Line<[number, number]>;
+    legendPathData: [number[]];
```
martinRenou, line 676: Your compilation issue seems to be related to the type you changed here, `[number, number][]` was more specific so you might want to keep it that way.

js/src/Lines.ts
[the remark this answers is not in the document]
marimeireles: Should we create tests for these?

marimeireles:
So, there's one problem here.
I can't use the previous type of `area: d3.Line<[number, number]>;` for  reasons that I can't understand :/
If I use it instead of the current `any` that I'm using I get the following error:
`src/Lines.ts(433,23): error TS2339: Property 'y0' does not exist on type 'Line<[number, number]>'.`
And a few others.

js/src/Lines.ts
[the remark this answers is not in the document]
martinRenou: Yes, more test coverage is good!

You will see your code fail if you try to change the `marker_size` attribute from Python.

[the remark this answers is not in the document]
martinRenou: You can't really change the signature of a function/method and expect it to work as intended, even if you don't use some of those arguments.

It's not super-well documented in Backbone, but I remember the callback function passed to [listenTo](https://backbonejs.org/#Events-listenTo) is expected to have a signature like this: `function(model, value): void`, `model` being the object you listen to, and `value` I suppose is the value of the attribute you listen to (Don't take my words as granted, the documentation being what it is I think backbone's source code might be the only source of truth).

So if you change the signature like you are doing, the parameter `marker_size` will actually contain `this.model` instead of the marker size, that's why it doesn't work.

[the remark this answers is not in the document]
marimeireles: huh, I just wrote the following test:

```

def test_marker_size(scales):
    lines = bqplot.Lines(x=[[0, 1], [0, 1, 2]], y=[[0, 1], [1, 0, -1]], scales=scales)

    lines.marker_size = 7
```

Using the code without the `model` variable and the test passed.
Did I write the test wrong?

marimeireles:
One more question, what are these empty methods down here?

https://github.com/bqplot/bqplot/blob/5595c241093d6f300eac76795d03217c625922bc/js/src/Lines.ts#L663-L670

js/src/Lines.ts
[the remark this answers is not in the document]
martinRenou: That doesn't work, unfortunately. You are only testing the Python code here, the is no actual JavaScript model and view that are generated. The Python will run just fine, it's the JavaScript code that will fail.

You can try in a Notebook and you should see an error in the browser console.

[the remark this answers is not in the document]
martinRenou: But you can actually add a JavaScript test directly here: https://github.com/bqplot/bqplot/blob/master/js/src/test/lines.ts, and changing the `marker_size` doing something like: `model.set('marker_size', 7)`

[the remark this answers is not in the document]
marimeireles: ah, right :)
thanks Martin!
I actually didn't find the js tests thought we only had the python ones :see_no_evil:

martinRenou:
> One more question, what are these empty methods down here?

Those are implementations of the parent's abstract methods. We need them although they don't do anything.

But the fact they don't do anything is a good clue showing that `Lines` is not implemented correctly, it should implement those methods if we want lines to behave the same as other marks when it comes to selecting/styling marks.

marimeireles:
Is `bar.ts` a good place to look for a similar implementation?
I can do this in this PR.

martinRenou:
> Is bar.ts a good place to look for a similar implementation? I can do this in this PR.

Maybe `Scatter`/`ScatterBase` is the best place to look at, it's the only mark that is fully implemented.

I think it would make sense to have it in a separate PR :)

marimeireles:
I added a few other tests for the `model.set` since I was there already.

README.md
```diff
@@ -110,6 +110,32 @@ import bqplot
 
 That's it! You're ready to go!
 
+### Running tests
```
martinRenou, line 113: Thanks for adding docs, if this is missing, please add it to the `CONTRIBUTING.md` file instead of the README

marimeireles: k!

martinRenou: Actually, I was quite wrong. It's not `CONTRIBUTING.md` where the dev installation instructions are, it's actually in the README but it's earlier in the file.

js/src/Figure.ts
```diff
@@ -110,8 +110,8 @@ class Figure extends widgets.DOMWidgetView {
         // If there is no restriction on the plottable area of the figure,
         // then these two variables are the maximum of the values in the
         // corresponding variables x_pad_dict, y_pad_dict.
-        this.x_padding_arr = {};
-        this.y_padding_arr = {};
+        this.xPadding_arr = {};
+        this.yPadding_arr = {};
```
martinRenou, line 114: `xPaddingArr` and `yPaddingArr`?

js/src/Lines.ts
```diff
@@ -140,14 +142,14 @@ export class Lines extends Mark {
         }
     }
 
-    update_labels() {
+    update_labels(): any{
```
martinRenou, line 145:
```suggestion
    update_labels(): void {
```

js/src/Lines.ts
```diff
@@ -163,173 +165,171 @@ export class Lines extends Mark {
...
           })
           .style("fill-opacity", function(d, i) {
-              return fill === "inside" ? fill_opacities[i] : "";
+              return fill === "inside" ? fillOpacities[i] : "";
           });
         curves.select(".area")
-          .style("fill", function(d, i) { return that.get_fill_color(d, i); })
-          .style("opacity", function(d, i) { return fill_opacities[i]; });
+          .style("fill", (d, i) => { return this.get_fill_color(d, i); })
+          .style("opacity", function(d, i) { return fillOpacities[i]; });
```
martinRenou, line 206: There are still some `function`s around

js/src/Lines.ts
```diff
@@ -592,72 +591,72 @@ export class Lines extends Mark {
...
         const that = this;
-        const fill_color = this.model.get("fill_colors");
+        const fillColor = this.model.get("fill_colors");
         const opacities = this.model.get("opacities");
         this.d3el.selectAll(".curve").each(function(d, i) {
             const curve = d3.select(this);
             curve.selectAll(".dot")
                 .style("opacity", opacities[i])
-                .style("fill", that.get_mark_color(d, i) || fill_color[i]);
+                .style("fillTS2339:", that.get_mark_color(d, i) || fillColor[i]);
```
martinRenou, line 637:
```suggestion
                .style("fill:", that.get_mark_color(d, i) || fillColor[i]);
```

js/src/Lines.ts
```diff
@@ -592,72 +591,72 @@ export class Lines extends Mark {
...
+        const yPadding = xPadding;
+        if(xPadding !== this.xPadding || yPadding !== this.yPadding) {
+            this.xPadding = xPadding;
+            this.yPadding = yPadding;
             this.trigger("mark_padding_updated");
         }
     }
 
     update_marker_style() {
         const that = this;
```
martinRenou, line 630: Maybe get rid of that one

js/src/Lines.ts
```diff
@@ -670,16 +669,16 @@ export class Lines extends Mark {
...
-    legend_line: d3.Line<[number, number]>;
-    legend_path_data: [number, number][];
+    legendEl: d3.Selection<any, any, any, any>;
+    legendLine: d3.Line<[number, number]>;
+    legendPathData: [number, number][];
     selector: BaseSelector;
-    selector_model: SelectorModel;
-    area: d3.Area<[number, number]>;
+    selectorModel: SelectorModel;
+    area: any;
```
martinRenou, line 677: What was the problem with `d3.Area<[number, number]>`?

js/src/Lines.ts
marimeireles: Yes... But I was obliged to use `functions` in some instances. I changed the cases where we were using `that` to `this`.
We sometimes might want to use `function`, right? I don't think it qualifies as bad code per si, it's just related to what you want to do with the scope of your function. So... I'd say it's okay to have them?

martinRenou: The thing is that if we need to change that code and call an instance method in there with `this`, we won't use the right context. Just for this and for the sake of homogeneity I would change them all. Let's get rid of all the old JavaScript :P

js/src/Lines.ts
marimeireles: Hum, so this is a difficult one. As you can see

```
    update_marker_style() {
        const that = this;
        const fillColor = this.model.get("fill_colors");
        const opacities = this.model.get("opacities");
        this.d3el.selectAll(".curve").each(function(d, i) {
            const curve = d3.select(this);
            curve.selectAll(".dot")
                .style("opacity", opacities[i])
                .style("fill:", that.get_mark_color(d, i) || fillColor[i]);
        });
    }
```

We're making use of both objects here, both the `this` from inside the scope of the function the `this` of the outside scope ;/
I looked a bit into `d3` docs and I believe that's the recommended way of doing it, with `each` and using `select.this`, I tried to be more specific and get the element but I couldn't, it's just data... not a DOM object, meaning it doesn't have info like `parent` and stuff.
I don't really know how to address to this issue other than using `that`, please let me know if you have an idea!

js/src/Lines.ts
marimeireles: Hum, I see. Okay then.

marimeireles: I didn't know it was old JS, thought it was used side by side with arrow functions.

js/src/Lines.ts
martinRenou: Indeed. That's an annoying one. Let's keep it that way for now then :)

js/src/Lines.ts
marimeireles: This kind of error that I couldn't figure out why:
```
src/Lines.ts:433:23 - error TS2339: Property 'y0' does not exist on type 'Line<[number, number]>'.

433             this.area.y0(this.parent.plotarea_height);
                          ~~

src/Lines.ts:435:23 - error TS2339: Property 'y0' does not exist on type 'Line<[number, number]>'.

435             this.area.y0(0)
                          ~~

src/Lines.ts:437:23 - error TS2339: Property 'y0' does not exist on type 'Line<[number, number]>'.

437             this.area.y0(function(d: any) { return yScale.scale(d.y0) + yScale.offset; })
                          ~~

src/Lines.ts:473:12 - error TS2339: Property 'y1' does not exist on type 'Line<[number, number]>'.

473           .y1(function(d: any) { return yScale.scale(d.y) + yScale.offset; })
               ~~

src/Lines.ts:476:23 - error TS2339: Property 'y0' does not exist on type 'Line<[number, number]>'.

476             this.area.y0(this.parent.plotarea_height);
                          ~~

src/Lines.ts:478:23 - error TS2339: Property 'y0' does not exist on type 'Line<[number, number]>'.

478             this.area.y0(0)
                          ~~

src/Lines.ts:480:23 - error TS2339: Property 'y0' does not exist on type 'Line<[number, number]>'.

480             this.area.y0(function(d: any) { return yScale.scale(d.y0) + yScale.offset; })
                          ~~


Found 7 errors.


```
Here's a link for the original code: https://github.com/bqplot/bqplot/blob/2560b1e347d4d0f1f4a0a533c035b6957f91fd78/js/src/Lines.ts#L478

js/src/Lines.ts
marimeireles: hehe yeah, but if you have input on how to fix it I'm interested!

martinRenou: I remember we had to deal with something similar before but I don't recall where.

js/src/Lines.ts
martinRenou: In your error message, it complains because it thinks `this.area` is a `Line<[number, number]>`, do you know where that come from? It should think it's a `d3.Area<[number, number]>`

marimeireles: lol ;/
I was copying the Line type
let's talk about this no further :see_no_evil:

marimeireles: hehe thank you!

marimeireles:
Ok... I knew this file would be problem cuz I got confused with the type stuff, I have to go from the place that I'm in now... Will prbb do it later tonight or tomorrow morning!

marimeireles:
@martinRenou I think it's finally ready?

js/src/Lines.ts
```diff
@@ -719,23 +716,23 @@ export class Lines extends Mark {
...
+  update_dots_xy(animate): void {
     if (this.model.get('marker')) {
-      const x_scale = this.scales.x,
-        y_scale = this.scales.y;
-      const animation_duration =
-        animate === true ? this.parent.model.get('animation_duration') : 0;
+      const xScale = this.scales.x,
+        yScale = this.scales.y;
+      const animationDuration =
+        animate === true ? this.parent.model.get('animationDuration') : 0;
```
martinRenou, line 724:
```suggestion
        animate === true ? this.parent.model.get('animation_duration') : 0;
```

js/src/ScatterGL.ts
```diff
@@ -108,37 +106,37 @@ class AttributeParameters {
...
+    mesh_per_attribute: number = 1,
+    normalized: Boolean = false,
+    use_colormap: Boolean = true
   ) {
     super(array, item_size, mesh_per_attribute, normalized);
     this.use_colormap = use_colormap;
   }
 
-  use_colormap: boolean;
+  use_colormap: Boolean;
```
martinRenou, line 124:
```suggestion
  use_colormap: boolean;
```

js/src/ScatterGL.ts
```diff
@@ -108,37 +106,37 @@ class AttributeParameters {
...
   constructor(
     array: TypedArray,
-    item_size = 1,
-    mesh_per_attribute = 1,
-    normalized = false,
-    use_colormap = true
+    item_size: number = 1,
+    mesh_per_attribute: number = 1,
+    normalized: Boolean = false,
+    use_colormap: Boolean = true
```
martinRenou, line 118:
```suggestion
    normalized: boolean = false,
    use_colormap: boolean = true
```

js/src/ScatterGL.ts
```diff
@@ -108,37 +106,37 @@ class AttributeParameters {
   array: TypedArray;
   item_size: number;
   mesh_per_attribute: number;
-  normalized: boolean;
+  normalized: Boolean;
```
martinRenou, line 109:
```suggestion
  normalized: boolean;
```

js/src/ScatterGL.ts
```diff
@@ -95,9 +93,9 @@ const create_colormap = function (scale) {
 class AttributeParameters {
   constructor(
     array: TypedArray,
-    item_size = 1,
-    mesh_per_attribute = 1,
-    normalized = false
+    item_size: number = 1,
+    mesh_per_attribute: number = 1,
+    normalized: Boolean = false
```
martinRenou, line 98:
```suggestion
    normalized: boolean = false
```

js/src/ScatterGL.ts
```diff
@@ -674,7 +668,7 @@ export class ScatterGL extends Mark {
   }
 
   update_attribute(
-    name: string,
+    name: String,
```
martinRenou, line 671:
```suggestion
    name: string,
```

(Applies multiple places in the file)

js/src/ScatterGL.ts
```diff
@@ -1255,7 +1230,7 @@ export class ScatterGL extends Mark {
 
     elements_added
       .append('path')
-      .attr('transform', (d, i) => {
+      .attr('transform', function (d, i) {
```
martinRenou, line 1233:
```suggestion
      .attr('transform', (d, i) => {
```

(applies multiple places in the file)

js/src/test/lines.ts
```diff
@@ -9,81 +9,100 @@ describe('lines >', () => {
...
+    lines.model.set('labels_visibility', 'label');
+    lines.model.set('curves_subset', [2, 4, 5]);
+    lines.model.set('line_style', 'solid');
+  });
+
   it('create 1d ordinal', async function () {
-    const x = ['a', 'b'];
-    const scale_x = await create_model_bqplot(
+    let x = ['a', 'b'];
+    let scale_x = await create_model_bqplot(
```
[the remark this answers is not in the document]
marimeireles: werid! I ran the prettify on everything again :/
I thought it was fixing this stuff automatically...
I'll have a look again thanks, sry for this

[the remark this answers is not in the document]
martinRenou: No worries at all :)

js/src/ScatterGL.ts
marimeireles: I was planning not to touch too much on the other files, but I can do this one!

marimeireles: I mean, actually, it's a bunch of changes. Sorry, I was planning to do this kind of thing in the actual refactoring of the file... I can remove the changes I did here, I was changing a variable name only ;/
Other than that, I think it's ready for a new review.

marimeireles:
Hey @martinRenou, so this is a hard to catch bug. A few things I learned:

1. The problem lies on setting the color of the mark. The mark you're using doesn't matter, the color attr. is not being persisted and it's defaulted to black.
2. The image we get out of galata is this weird one:
lines_update_ipynb_cell_10
but in fact if you run the notebook you'll have the following:
image
Which is much more similar to the one expected:
image
So, I think galata is just showing us what is the difference between them?

This bug didn't happen before, it was after you squashed things... I was wondering if you remember changing something in this regard @martinRenou?
I went through the `diff` a couple of times...

This part is different:

```
@@ -543,6 +543,24 @@ export abstract class Mark extends widgets.WidgetView {
     return _.compact(hit_check).length > 0;
   }
 
+  // For backward-compatibility with bqplot plugins
+  // TODO Remove it when we make a backward-incompatible release
+  set x_padding(value: number) {
+    this.xPadding = value;
+  }
+
+  get x_padding(): number {
+    return this.xPadding;
+  }
+
+  set y_padding(value: number) {
+    this.yPadding = value;
+  }
+
+  get y_padding(): number {
+    return this.yPadding;
+  }
+
```

But doesn't really make a diff. I think. It's literally just changing variable names from snake to camel case.

marimeireles:
Really cool this galata thing. Thank you for find the issue Martin! <3

martinRenou:
> Really cool this galata thing

Yep! We really should add more of those tests and use Galata in more projects :D It's super useful already

PR #1305, opened 2021-03-24
The "Pretty" PR

marimeireles:
Adds configuration for both prettier and eslint. Enforces prettier rules but not eslint rules.


.github/workflows/build.yml
```diff
@@ -62,6 +63,9 @@ jobs:
         yarn install
         yarn run test
 
+    - name: Run prettyfier
+      run: yarn run prettier --check .
```
martinRenou, line 67:
```
run: |
  cd js
  yarn run prettier --check .
```

marimeireles: Ah, I thought it'd pick it up from the command above.
Thanks!

martinRenou: I recall there was a `working-directory` option that you can use otherwise. Something like:
```
run: yarn run prettier --check .
working-directory: js
```

PR #1306, opened 2021-03-25
Add types to HandDraw

martinRenou:


js/src/HandDraw.ts
```diff
@@ -28,163 +27,155 @@ export class HandDraw extends Interaction {
...
 
-    // Update line index
-    this.update_line_index();
-    this.listenTo(this.model, 'change:line_index', this.update_line_index);
-    this.model.on_some_change(['min_x', 'max_x'], this.set_limits, this);
+    this.model.on_some_change(['min_x', 'max_x'], this.setLimits, this);
   }
 
-  set_lines_view() {
+  private async setLinesView(): Promise<void> {
```
marimeireles, line 40: methods names should be snake case?

marimeireles: or is it overriding something else?

marimeireles (commented):
Other than this one question, looking good! :)

js/src/HandDraw.ts
martinRenou: Nope, I think we decided to follow the JupyterLab conventions, so no more snake_case :)

marimeireles: oh!
and pretty is not complaining? so ok :)

martinRenou: Some of them are still snake_case because I am waiting for us to work on super classes like `Mark` before renaming them

PR #1309, opened 2021-03-26
Add types to Hist

martinRenou:


martinRenou:
Setting `colors` to `['red', 'blue']` does not seem to work anymore

PR #1317, opened 2021-04-06
Automatically commit pretty TypeScript code

martinRenou:


martinRenou:
It looks like the action doesn't work as intended (it might only work if the branch is pushed upstream https://github.com/bqplot/bqplot, which we don't do)

PR #1320, opened 2021-04-08
Fix JupyterLab 2 build issue

martinRenou:


.github/workflows/build.yml
```diff
@@ -67,42 +67,18 @@ jobs:
...
-      working-directory: ui-tests
-
-    - name: Wait for JupyterLab
-      uses: ifaxity/wait-on-action@v1
-      with:
-        resource: http-get://localhost:8888/api
-        timeout: 20000
-
-
-  build:
```
martinRenou, line 83: Just reducing the number of builds here

PR #1323, opened 2021-04-09
Graph Fix arrows

martinRenou:
This is is a different approach than https://github.com/bqplot/bqplot/pull/1261 for trying to fix the directed graphs.

The approach here is not pixel perfect, and does not work in the case where the marker is not a circle.

arrow

### DIFF
--- a/js/src/Graph.ts
+++ b/js/src/Graph.ts
@@ -66,7 +66,8 @@ export class Graph extends Mark {
       .append('defs')
       .append('marker')
       .attr('id', 'arrow')
-      .attr('refX', 0)
+      .attr('viewBox', '0 0 10 10')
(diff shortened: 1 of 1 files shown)


js/src/Graph.ts
```diff
@@ -66,7 +66,8 @@ export class Graph extends Mark {
       .append('defs')
       .append('marker')
       .attr('id', 'arrow')
-      .attr('refX', 0)
+      .attr('viewBox', '0 0 10 10')
+      .attr('refX', 15)
```
martinRenou, line 70: This is hardcoded to 15 (radius of the circle). Not sure it's a good idea.

martinRenou:
This is not the right fix. This cannot work.

PR #1340, opened 2021-04-20
fix: under jlab3, d3.event can be null because of duplicate d3 modules

maartenbreddels:
Similar problem as described in #1276

In my case, I had ipyvolume installed, which had it's own copy of d3 and d3-selection. Bqplot also has its own d3-selection, ipyvolume sets it's d3.event, and bqplot read d3.event from its private module, giving null:

This prevented zooming from working with the PanZoom Interact, giving:
```
PanZoom.js:168 Uncaught TypeError: Cannot read property 'preventDefault' of null
```

According to:
https://developer.mozilla.org/en-US/docs/Web/API/Window/event
This is supported by all browsers.

I think a prettier solution would be to move to d3-selection 2.0, which supports the event being passed as argument. However, I'm not sure that this will work if 3rd party libraries (such as ipyvolume) still have the older version of d3-selection.

### DIFF
--- a/js/src/PanZoom.ts
+++ b/js/src/PanZoom.ts
@@ -15,13 +15,7 @@
 
 import * as widgets from '@jupyter-widgets/base';
 import * as d3 from 'd3';
-// var d3 =Object.assign({}, require("d3-selection"));
-const d3GetEvent = function () {
-  // In JupyterLab we can have require('d3-selection').event
-  // be a different module as the one that is used in d3.on
-  // leading to a null event, for now we can use window.event as fallback.
-  return require('d3-selection').event || window.event;
-}.bind(this);
+import { d3GetEvent } from './utils';
 import * as interaction from './Interaction';
 import * as _ from 'underscore';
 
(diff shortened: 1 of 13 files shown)


js/src/PanZoom.ts
```diff
@@ -17,7 +17,10 @@ import * as widgets from '@jupyter-widgets/base';
 import * as d3 from 'd3';
 // var d3 =Object.assign({}, require("d3-selection"));
 const d3GetEvent = function () {
-  return require('d3-selection').event;
+  // In JupyterLab we can have require('d3-selection').event
+  // be a different module as the one that is used in d3.on
+  // leading to a null event, for now we can use window.event as fallback.
+  return require('d3-selection').event || window.event;
```
martinRenou, line 23: Should we do the same in other files (like Graph.ts)? This d3-selection hack is done in multiple places

maartenbreddels: Do you agree with the solution? If so, I think we could put it in utils.ts then

martinRenou: Yep!

PR #1341, opened 2021-04-21
Refactor ticks

martinRenou:
Reboot of https://github.com/bqplot/bqplot/pull/1216


martinRenou:
One case when the old version was better (left: old, right: new): 

diff

PR #1347, opened 2021-05-11
Fix zooming

### LINKED ISSUE #81: Scroll behavior of notebook while zooming in firefox
Since #76, zooming behavior is enabled in Firefox but the scroll behavior of the notebook is not disabled properly.

### LINKED ISSUE #1304: Zoom is not working anymore
Example code
```
import numpy as np

plt.figure(1, title='Line Chart')
np.random.seed(0)
n = 200
x = np.linspace(0.0, 10.0, n)
y = np.cumsum(np.random.randn(n))
plt.plot(x, y)
plt.show()
```

Now clicking the panzoom button, we don't have a working zoom anymore.

This was originally reported downstream at https://github.com/spacetelescope/jdaviz/issues/486 and the upstream issue 
at https://github.com/d3/d3-zoom/issues/231

This affects Chrome 89.0.4389.90 (OSX), and is demonstrated by zoom not working at: https://observablehq.com/@philjones/zoom-svg-chrome-89-0-4389-90-bug

maartenbreddels:
This fixes #1304 and fixes #81 (Whow, that's an oldie!)

### DIFF
--- a/js/src/PanZoom.ts
+++ b/js/src/PanZoom.ts
@@ -43,14 +43,9 @@ export class PanZoom extends interaction.Interaction {
             that.mouseup();
           })
       )
-      .on('mousewheel', () => {
+      .on('wheel', () => {
         that.mousewheel();
-      })
-      .on('DOMMouseScroll.zoom', () => {
-        that.mousewheel();
-      })
-      .on('mousewheel.zoom', null)
-      .on('wheel.zoom', null);
+      });
     this.active = false;
 
     this.update_scales();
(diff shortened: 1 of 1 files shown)


maartenbreddels:
Support seems good https://developer.mozilla.org/en-US/docs/Web/API/Element/wheel_event

maartenbreddels:
I think downgrading nodejs should work. Maybe try a few major versions back.

(from mobile phone)

pllim:
Instead of messing with nodejs, I got something to install by hacking your `yarn.lock` before installing from source:

```diff
diff --git a/js/yarn.lock b/js/yarn.lock
index ace7b2fb..c7f003d3 100644
--- a/js/yarn.lock
+++ b/js/yarn.lock
@@ -3756,7 +3756,6 @@ execa@^5.0.0:
   dependencies:
     cross-spawn "^7.0.3"
     get-stream "^6.0.0"
-    human-signals "^2.1.0"
     is-stream "^2.0.0"
     merge-stream "^2.0.0"
     npm-run-path "^4.0.1"
```

```
Building wheels for collected packages: bqplot
  Building wheel for bqplot (PEP 517) ... done
  Created wheel for bqplot: filename=bqplot-0.12.26-py2.py3-none-any.whl ...
  Stored in directory: ...
Successfully built bqplot
Installing collected packages: bqplot
Successfully installed bqplot-0.12.26
```

Not sure why the version isn't reporting "dev" and also zoom still does not work for me.

Issue #1378, opened 2021-06-15
Non-numeric Ordinal Brushing


dfreeman06:
**Describe the bug**
There seems to be an issue with Brushes for Non-numeric ordinal scales. This example would work if ["A","B","C"] were [1,2,3]. 

**To Reproduce**
```python
from bqplot import *
from ipywidgets import *
from bqplot.interacts import *

scale_x1 = DateScale()
scale_y1 = OrdinalScale(domain=["A","B","C"])
x = np.arange('2020-01-01', '2020-04-01', dtype='datetime64[M]')
scatter1 = Scatter(x=x, y=["A","B","C"], scales={'x':scale_x1, 'y':scale_y1})

x_axis = Axis(scale = scale_x1)
y_axis = Axis(
    scale = scale_y1,
    orientation="vertical",
)

fig1 = Figure(marks=[scatter1], 
              axes=[x_axis, y_axis],
              interaction=BrushSelector(x_scale=scale_x1,
                     y_scale=scale_y1, marks=[scatter1]), background_style={'fill':'#eee'})
fig1
```
Brush the plot as shown in the screenshot below.
```python
fig1.interaction.selected_y
```

**Expected behavior**
Would expect the brush to have `array([0,2], dtype=int32)`. 

**Screenshots**
![image](https://user-images.githubusercontent.com/6493354/122127694-8cf3a900-ce01-11eb-9b8d-3ff354ce231f.png)

dfreeman06:
May be related to https://github.com/bqplot/bqplot/pull/1226?

PR #1379, opened 2021-06-16
Map ordinal selected values against their position in the domain

dfreeman06:
Attempt to address #1378. Mapping the Ordinal selected values from a brush through their index position in the domain.


js/src/OrdinalScaleModel.ts
```diff
@@ -27,7 +27,7 @@ export class OrdinalScaleModel extends ScaleModel {
   }
 
   typedRange(values) {
-    return new Int32Array(values.map(Number));
+    return new Int32Array(values.map((value) => this.domain.indexOf(value)));
```
bollwyvl, line 30: could it just be:
```ts
return new Int32Array(values.map(this.domain.indexOf));
```
this would avoid allocating a new function/scope each time... though maybe the `this` shenanigans come into play...

dfreeman06: I originally tried that but it was throwing errors... probably due to `this` shenanigans.

bollwyvl: k, how about:

```ts
return new Int32Array(values.map(this.domain.indexOf, this.domain));
```

dfreeman06: Yes that behaves and pushed the change.

sanbales:
This [PR](https://github.com/bqplot/bqplot/pull/1226) didn't add a test.  Looking through the tests, I'm not sure how we would add a test for this.  @martinRenou / @maartenbreddels :  can you point us to a PR that did something similar and had a test for it?

The process for producing the error is [here](https://github.com/bqplot/bqplot/issues/1378), but it requires manual brushing.

martinRenou:
Since recently we added galata tests in bqplot. Galata should have everything for "manually" testing the brush and checking that it renders properly. But that requires good knowledge of galata which I don't have...

Can this be bug be triggered programmatically from Python? That should help create a visual test. I can look into it.

dfreeman06:
Thanks @martinRenou, it looks like that should work. However, in investigating programmatically setting the brush `selected_y` values another concern popped up. 

This PR originally introduces using the domain index for the values in the brush's `selected` traits which is different than the existing behavior which uses the domain values itself. It might be more consistent to deal with the indices in the domain, however to minimize changes to other users it looks like the:

```typescript
  typedRange(values) {
    return new Int32Array(values.map(this.domain.indexOf, this.domain));
  }
```
could be replaced with:

```typescript
  typedRange(values) {
    return values;
  }
```

There is also a new galata test notebook `ordinal_scale.ipynb` which tests two different figures. One figure uses the ordinal scale with strings while the other uses the ordinal scale with integers. The test notebook also programatically sets the brush's selected values and captures the output screenshots.

sanbales:
> 
> 
> Since recently we added galata tests in bqplot. Galata should have everything for "manually" testing the brush and checking that it renders properly. But that requires good knowledge of galata which I don't have...
> 
> Can this be bug be triggered programmatically from Python? That should help create a visual test. I can look into it.

@martinRenou : @dfreeman06 added a test using Galata that we believe is working and testing the brushes programmatically, at least locally it is passing, would it be possible to run the workflow to see if everything is on the up and up?

sanbales:
Just a friendly ping to the BQPlot team to see if there is something else we should do for this PR.  Also curious if @maartenbreddels has any issues with this as per his [PR](https://github.com/bqplot/bqplot/pull/1226).

js/src/OrdinalScaleModel.ts
```diff
@@ -27,7 +27,7 @@ export class OrdinalScaleModel extends ScaleModel {
   }
 
   typedRange(values) {
-    return new Int32Array(values.map(Number));
+    return values;
```
sanbales, line 30: As per @bollwyvl 's recommendation, this could be:

```typescript
return new Int32Array(values.map(this.domain.indexOf, this.domain));
```

Which would make the scale work with the indeces of the `domain` (I believe that is what is meant by range [here](https://github.com/bqplot/bqplot/pull/1226#discussion_r509214964), @maartenbreddels please don't hesitate to let us know if we misunderstood what was intended by `range`).

We went with the `values` instead because otherwise it could break other people's code, but we acknowledge it would probably be better to use the line above.

bollwyvl:
y'all let me know if there's anything i can do to push this forward...

sanbales (approved):
Not sure non-contributor reviews matter, but in case it does, here is my approval.

martinRenou:
Thanks for adding tests. Ideally, we would add a galata test that simulates a selection with the mouse and test that the `selected_y` attribute gets changed properly. But I don't know yet how to do that, so let's merge and see later!

Thanks!

bollwyvl:
Everybody's real busy, for sure, but Is there a timeline for a `.30` release? Anything we can do to help?

PR #1389, opened 2021-07-19
Get rid of WebGL related stuffs

martinRenou:
Very early draft. Please don't merge.


bqplot/marks.py
```diff
@@ -730,12 +730,6 @@ def default_opacities(self, value):
     _model_name = Unicode('ScatterModel').tag(sync=True)
```
martinRenou, line 730: Need to remove threejs dependency from package.json

martinRenou:
PR tested with https://github.com/bqplot/bqplot-gl 's master branch (adding visual tests with Galata and current JS tests soon)

maartenbreddels:
What about creating a webgl canvas, and letting the marks create a threejs renderer if needed?

martinRenou:
@maartenbreddels thanks for the review :) 

> What about creating a webgl canvas, and letting the marks create a threejs renderer if needed?

What would be the upside of this?

maartenbreddels:
Most of the code in the Figure class can stay as it is (only the line/lines that creates the rendered will change), and we don't need any Figure subclass.

martinRenou:
Right! Yeah that would be cool. I'll try it this way then.

maartenbreddels:
This means you can write a threejs, raw WebGL, or Regl Mark and mix them.

martinRenou:
> The only worry I have is that now we don't lazily create the context, like before using createWebGLRenderer (which now should be called createWebGLCanvas I think. Otherwise, we'll create these contexts too often, and the web-browser will show this 'broken' image because we created too many WebGL context.

This should be resolved now :)

PR #1394, opened 2021-08-06
Add a default toolbar for figure

trungleduc:
## References

This patch adds a default toolbar to `Figure` widget, this toolbar contains the same buttons as in `Toolbar` widget but it is implemented directly in `Figure` class

## User-facing changes

- A new toolbar is shown on top right of figure when the cursor is over the figure.

bq1

- New boolean parameter `display_toolbar` (default `True`) is added to constructor of `Figure` to show or hide the toolbar.
-  The toolbar widget is removed from `pyplot` figure. Now `display_toolbar` parameter of `plt.show` is used to show or hide the default toolbar.

bq2

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,3 +64,6 @@ node_modules
 *.swp
 
 *.tsbuildinfo
+
+# VS Code settings
+.vscode/*
--- a/.gitignore
+++ b/.gitignore
@@ -64,6 +64,3 @@ node_modules
 *.swp
 
 *.tsbuildinfo
-
-# VS Code settings
-.vscode/*
--- a/bqplot/pyplot.py
+++ b/bqplot/pyplot.py
@@ -118,7 +118,7 @@ def hashable(data, v):
     return True
 
 
-def show(key=None, display_toolbar=False):
+def show(key=None, display_toolbar=True):
     """Shows the current context figure in the output area.
 
     Parameters
@@ -152,13 +152,8 @@ def show(key=None, display_toolbar=False):
         figure = current_figure()
     else:
         figure = _context['figure_registry'][key]
-    if display_toolbar:
-        if not hasattr(figure, 'pyplot'):
-            figure.pyplot = Toolbar(figure=figure)
-            figure.pyplot_vbox = VBox([figure, figure.pyplot])
-        display(figure.pyplot_vbox)
-    else:
-        display(figure)
+    figure.display_toolbar = display_toolbar
+    display(figure)
 
 
 def figure(key=None, fig=None, **kwargs):
--- a/bqplot/pyplot.py
+++ b/bqplot/pyplot.py
@@ -50,7 +50,6 @@ Pyplot
 import sys
 from collections import OrderedDict
 from IPython.display import display
-from ipywidgets import VBox
 from ipywidgets import Image as ipyImage
 from numpy import arange, issubdtype, array, column_stack, shape
 from .figure import Figure
@@ -58,7 +57,6 @@ from .scales import Scale, LinearScale, Mercator
 from .axes import Axis
 from .marks import (Lines, Scatter, ScatterGL, Hist, Bars, OHLC, Pie, Map, Image,
                     Label, HeatMap, GridHeatMap, topo_load, Boxplot, Bins)
-from .toolbar import Toolbar
 from .interacts import (BrushIntervalSelector, FastIntervalSelector,
                         BrushSelector, IndexSelector, MultiSelector,
                         LassoSelector)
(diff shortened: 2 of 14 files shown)


bqplot/pyplot.py
```diff
@@ -118,7 +118,7 @@ def hashable(data, v):
     return True
 
 
-def show(key=None, display_toolbar=True):
+def show(key=None, display_toolbar=False):
```
martinRenou, line 121: Should we really change this? This will break long-time users assumption that it shows a toolbar by default.

js/src/Figure.ts
```diff
@@ -327,6 +329,10 @@ export class Figure extends widgets.DOMWidgetView {
       window.removeEventListener('resize', this.debouncedRelayout);
     });
 
+    if (this.model.get('show_toolbar')) {
```
martinRenou, line 332: We should maybe handle the case when `show_toolbar` is dynamically updated. Something like:

```typescript
this.model.on('change:show_toolbar', () => {
  if (this.model.get('show_toolbar')) {
    this.create_toolbar();
  } else {
    this.remove_toolbar();
  }
});
```

js/src/Figure.ts
```diff
@@ -1249,6 +1255,77 @@ export class Figure extends widgets.DOMWidgetView {
...
+   * Generate an integrated toolbar which is shown on mouse over
+   * for this figure.
+   *
+   */
+  create_toolbar(): void {
+    const toolbar = d3
+      .select(document.createElement('div'))
+      .attr('class', 'toolbar_div');
+
+    const _panzoom = document.createElement('button');
```
martinRenou, line 1268: No need for the `_` prefix for those variables.

js/src/FigureModel.ts
```diff
@@ -76,6 +79,88 @@ export class FigureModel extends widgets.DOMWidgetModel {
...
+      this._panzoomData.panning = true;
+    }
+  }
+
+  /**
+   * Creates a panzoom interaction widget for the this model.
+   *
+   * It will discover the relevant scales of this model.
+   */
+  _create_panzoom_model(): Promise<PanZoomModel> {
```
martinRenou, line 113: You can use the `private` keyword instead of the `_` prefix

js/src/FigureModel.ts
```diff
@@ -85,4 +170,10 @@ export class FigureModel extends widgets.DOMWidgetModel {
     scale_y: { deserialize: widgets.unpack_models },
     layout: { deserialize: widgets.unpack_models },
   };
+
+  private _panzoomData: {
```
martinRenou, line 174: No need for the `_` prefix

.gitignore
```diff
@@ -64,3 +64,6 @@ node_modules
 *.swp
 
 *.tsbuildinfo
+
+# VS Code settings
+.vscode/*
```
martinRenou, line 69: I personally disagree on having this kind of gitignores on repos, it doesn't scale. You can have a global gitignore on your setup.

martinRenou (commented):
Looks great! I only have minor comments

bqplot/pyplot.py
martinRenou: Actually, maybe `display_toolbar` should be kept `True` by default and used for setting the figure's `show_toolbar` property?

SylvainCorlay: Thinking again about this, this sounds like a good idea by @martinRenou.

martinRenou: Should we remove the old and ugly toolbar in this PR as well?

trungleduc: i think we should since user can still add this ugly toolbar independently.

trungleduc:
Added animation for `pyplot.show` behavior.

martinRenou (approved):
LGTM! Would you mind squashing the commits a bit?

trungleduc:
Thanks @martinRenou, yes can you squash it please!

SylvainCorlay:
Note: I think this is a 0.13 thing since it is a significant change of behavior.

martinRenou:
Should we move forward introducing bqscales and removing threejs then?

I was also wondering if we should still maintain a 0.12.x branch?

SylvainCorlay:
Yes, we should start merging thing and create a 0.12.x branch pointing at the last commit before this PR.

SylvainCorlay:
@maartenbreddels just wanted to ping you here. You should be interested in this.

maartenbreddels:
Thanks for the ping.
Really love how this looks, a bit worried that the Figure class gets a bit bloated, and a reminder to us that we should think of composability a bit more for widgets (at least have a few strategies in mind for this).
In any case, really nice work @trungleduc !

Issue #1393, opened 2021-08-04
How to ask save_png not to pop up a file dialog when filename is given?


pllim:
`figure.save_png(filename)` pops up a file dialog on Chrome. I don't need it because I am already giving it a filename with full path. How do I disable the file dialog?

pllim:
Closed by accident.

maartenbreddels:
What do you want to behavior to be?
PS: The filename is just a filename hint to the browser, it's not a full path, you cannot save to a particular path on the client if that's what you want.

pllim:
> save to a particular path

Alas, that is what I want. Calls the API with given filename, gets a file written, no file dialog.

pllim:
If you can somehow give me the Numpy array of the data to save, I can possibly use `scikit-image` to do that without the front-end limitation.

https://scikit-image.org/docs/dev/api/skimage.io.html#skimage.io.imsave

maartenbreddels:
That is possible with https://ipywebrtc.readthedocs.io/en/latest/ImageRecorder.html but you cannot wait for that in a sync way (e.g. you can do `image_data = get_image_data(fig)`, it needs to go through callbacks, possibly using some wrapper with async/await.

pllim:
> it needs to go through callbacks

Not sure what that means for me on the Python side... Is this something Mario can help with?

martinRenou:
Fixed by #1397

PR #1397, opened 2021-08-19
feat: make a PNG export of the figure available in the kernel

mariobuikhuizen:
A solution for #1393

Usage:
```python
def on_png_received(data):
    with open('mychart.png', 'bw') as f:
        f.write(data)

figure.get_png_data(on_png_received)
```

### DIFF
--- a/bqplot/figure.py
+++ b/bqplot/figure.py
@@ -156,6 +156,12 @@ class Figure(DOMWidget):
                                    display_name='Animation duration')
     display_toolbar = Bool(default_value=True).tag(sync=True)
 
+    def __init__(self, **kwargs):
+        super(Figure, self).__init__(**kwargs)
+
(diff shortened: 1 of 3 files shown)


maartenbreddels:
LGTM, would like a 2nd opinion from @SylvainCorlay or @martinRenou on this.

Some thoughts that don't need addressing perse:
 * download vs upload depends on your perspective, maybe we can avoid that word, maybe call the event 'send_msg'?
 * This particular use case is asking for a `await fig.get_png_data_async()`, which might work with @davidbrochart 's https://github.com/davidbrochart/akernel

martinRenou:
Rebased

pllim:
Thanks! Nice! Which release will this feature be in? 😸

martinRenou:
0.12.31 :) It should come soon

PR #1398, opened 2021-08-23
Run ui-tests on docker

martinRenou:


maartenbreddels:
@martinRenou why this change btw?
In case you want to run this locally, maybe https://github.com/nektos/act is an easier solution? I'm afraid the CI will become more complex with this.

martinRenou:
This is to prevent the difference there is in font rendering between the CI and the local run. This way both the CI and local run use the same setup to render and we don't need a threshold anymore for the diff in galata.

We've been using docker in JupyterLab's CI for Galata so I am following their setup.

martinRenou:
> In case you want to run this locally, maybe https://github.com/nektos/act is an easier solution?

👍🏽 We should maybe document doing this to run the tests locally indeed

Issue #1300, opened 2021-03-17
Bar plots with LogScale don't show up properly 


sntgluca:
Hi,

**Describe the bug**
I can't make Bar plots work using a LogScale scale.
I imagine the problem must be connected with the need for the minimum value of a LogScale be non-negative. It's similar but not the same as #1218.
<img width="663" alt="Screenshot 2021-03-17 at 16 54 53" src="https://user-images.githubusercontent.com/10941257/111498002-de789300-8741-11eb-80db-ca66ff9180b0.png">
<img width="710" alt="Screenshot 2021-03-17 at 16 55 08" src="https://user-images.githubusercontent.com/10941257/111498007-df112980-8741-11eb-9e15-c27de3c2ebaa.png">
<img width="731" alt="Screenshot 2021-03-17 at 16 55 34" src="https://user-images.githubusercontent.com/10941257/111498010-dfa9c000-8741-11eb-87c8-62183730550f.png">

**To Reproduce**
Here's a simple _working_ example using a LinearScale:

```python
import bqplot as bq
x = [i*3 for i in list("abc")] 
y = np.abs(np.random.randn(10) * 100)
```
```python
sc_x = bq.OrdinalScale()
sc_y = bq.LinearScale()
scales = {"x": sc_x, "y": sc_y}

line = bq.Lines(x=x, y=y, scales=scales, marker='circle')
bars = bq.Bars(x=x, y=y, scales=scales, opacities=[0.1])

ax_x = bq.Axis(scale=sc_x, grid_lines="none")
ax_y = bq.Axis(scale=sc_y, grid_lines="none", orientation="vertical")
bq.Figure(marks=[line, bars], axes=[ax_x, ax_y], title="Linear scale")
```

This example doesn't work at all:
```python
sc_x = bq.OrdinalScale()
sc_y = bq.LogScale()
scales = {"x": sc_x, "y": sc_y}

line = bq.Lines(x=x, y=y, scales=scales, marker='circle')
bars = bq.Bars(x=x, y=y, scales=scales, opacities=[0.1])

ax_x = bq.Axis(scale=sc_x, grid_lines="none")
ax_y = bq.Axis(scale=sc_y, grid_lines="none", orientation="vertical")
bq.Figure(marks=[line, bars], axes=[ax_x, ax_y], title="Empty log scale")
```

And setting the minimum value for the LogScale leads to weird behaviour:
```python
sc_x = bq.OrdinalScale()
[...]

PR #1375, opened 2021-06-10
Use d3.symlog scale instead of d3.log

martinRenou:
This allows having a plotting range <=0 

Fixes https://github.com/bqplot/bqplot/issues/1300 and fixes #1219

PR #1423, opened 2021-12-07
fix: properly treat Bar.base with stacked Bars

maartenbreddels:
This also allows the use of bars with log scales

# bug

We currently subtract `base` (baseValue in js code) each time from the value, which caused stacked bars to not show correct values. See demonstration of the bug:
bqplot-bars-base-bug

Code to reproduce:
```python
import bqplot
import numpy as np
import bqplot.pyplot as plt
N = 1000
np.random.seed(42)
log = False
sx = bqplot.LinearScale(min=0, max=4)
sy = bqplot.LinearScale(min=-2, max=6)
if log:
    sy = bqplot.LogScale(min=0.1, max=10)

fig = plt.figure(scale_x=sx, scale_y=sy)
x, y, c = np.random.normal(size=(3, N))
# mark = plt.scatter(x, y**2)
# h = plt.bar(x, y**2)
if log:
    mark = plt.bar([1,2,3], [[1, 1, 1], [2, 3, 4]], base=0.1, color=["orange", "green"], scales={'x': sx, 'y': sy})
else:
    mark = plt.bar([1,2,3], [[1, -1, 1], [2, -3, 4]], base=0, color=["orange", "green"], scales={'x': sx, 'y': sy})
plt.show()
import ipywidgets as widgets
if log:
    slider = widgets.FloatLogSlider(min=-1, max=1.5, step=0.01)
else:
    slider = widgets.FloatSlider(min=-1, max=7)
    
widgets.jslink((mark, 'base'), (slider, 'value'))
slider
```

# solution

I've refactored the code a bit, which makes it easier to comprehend I think, and corrected the double subtraction.

Bars now behave as I expect:
bqplot-bars-base

There are a few issues/cornercase:
 * If base is larger then the first bar value, there is no proper way to display it, so we skip it and print a warning to the console.
 * Using a log scale, negative bars (i.e. where base is larger then the cumulate sum of y) makes totally no sense, but one could argue that's the user error.

### DIFF
--- a/js/src/Bars.ts
+++ b/js/src/Bars.ts
@@ -421,7 +421,7 @@ export class Bars extends Mark {
           rang === 'y' ? rangeScale(d.y1) : rangeScale(d.y0)
         )
         .attr(rangeControl, (d: BarGroupValue) =>
-          Math.abs(rangeScale(d.y1 + d.yRef) - rangeScale(d.y1))
+          Math.abs(rangeScale(d.y0) - rangeScale(d.y1))
         );
     } else {
       bandWidth = Math.max(1.0, this.groupedScale.bandwidth());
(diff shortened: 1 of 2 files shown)


martinRenou (commented):
Should the top of the bars actually be equal to `base + y`?

IIUC, that is what plotly does:
base


This would fix your issue with stacked bars, because bars would always have the same height, whatever the base is.

maartenbreddels:
I think the current behaviour makes sense, given that if you plot in log space, and base is 10, and y value is 100, you want a bar from `10` to `100`, not till `110`

ibdafna:
@maartenbreddels Are you still looking to pursue this? Looks not not based on the last comment, but I thought I'd ask 😃

maartenbreddels:
Yeah, I think this is ready for review/merge. I just expressed I disagreed with Martin. Also, as it is now, the behaviour is the same as before, so non-breaking.

maartenbreddels:
@mariobuikhuizen and I looked at this, we think this is an improvement, replaces #1375 and fixes #1300

Good to merge @martinRenou ?

maartenbreddels:
It fixes a 'subset' of #1300, which is LogScale with non-negative values and a non-negative base (which the user has to set).

martinRenou (approved):
Thanks!

PR #1424, opened 2021-12-07
feat: log scale support for ScatterGL Mark

maartenbreddels:
ScatterGL assumed the scales to be linear, this adds support for log scales.

This would need to be ported to bqplot-gl and master as well.

### DIFF
--- a/js/shaders/scales.glsl
+++ b/js/shaders/scales.glsl
@@ -1,3 +1,6 @@
+#define SCALE_TYPE_LINEAR 1
+#define SCALE_TYPE_LOG 2
+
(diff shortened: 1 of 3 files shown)


martinRenou:
Rebased on 0.12.x

maartenbreddels:
The visual test might be a real issue, and might mean I misunderstand the stacked bar chars with mixed positive/negative values.

martinRenou:
This time the errors seem legit. I can investigate.

martinRenou:
This was breaking the datescales, it should work now.

Ordinal scales also don't work, but they didn't work prior to this PR apparently.

PR #1429, opened 2021-12-20
Make docs executable

martinRenou:
cc. @kaiayoung @ChakriCherukuri @ibdafna 

https://user-images.githubusercontent.com/21197331/146786698-690174dd-5ef8-40f5-8b9c-b93971d03559.mp4

We would need Thebe to support Pyolite as a kernel for having in-browser execution, but that's a start :)


martinRenou:

PR #1438, opened 2022-01-18
Updated S&P index performance nb

ChakriCherukuri:
1. Added trend line
2. Updated csv file


ibdafna:
@martinRenou failure seems to be unrelated?

martinRenou:
Yes indeed. I will try to debug this separately.

PR #1440, opened 2022-01-19
Fix CI

martinRenou:


martinRenou:
At first I was fixing the CI with setting the env variable: `NODE_OPTIONS: --openssl-legacy-provider`. But this only fixes the CI, not installing from source on user machine.

The second approach is taken from https://github.com/jupyterlab/jupyterlab/pull/11249

martinRenou:

PR #1447, opened 2022-01-27
Add default color CSS selectors to node text

ibdafna:
Signed-off-by: Itay Dafna <idafna@seas.upenn.edu>

- [x] Add default CSS selectors for dark and light themes
- [ ] ~Add node_text_color traitlet to Graph object~ (will do this in a separate PR)

Visual regressions tests will likely need to be updated for Node objects as default text color for dark theme is now lighter.


ibdafna:
update galata references

ibdafna:
update galata references

martinRenou (commented):
Looking at the visual diffs, this has an impact on the text color **in** nodes, should we keep the text in nodes black?

ibdafna:
I intentionally changed the text color of the nodes so that the text is
still visible if the text exceeds the bounds of the node's radius on the
dark theme. Otherwise it has the same color as the background and makes it
impossible to read.

martinRenou:
Makes sense, thanks!

martinRenou:

martinRenou:

PR #1466, opened 2022-02-10
fix: line label invisible in dark theme

martinRenou:


martinRenou:
That's not working. I will backport more PRs before trying to backport this one again.

PR #1487, opened 2022-03-29
fix: Figure.update_gl shim make for bqplot 0.12 compatibility

maartenbreddels:
This makes bqplot-image-gl work with bqplot 0.12.x and future versions.


martinRenou:
Thanks!

PR #1506, opened 2022-07-16
Add +, x and . as markers

### LINKED ISSUE #1457: Support marker style "." for pyplot
https://github.com/bqplot/bqplot/blob/f0b51fb09f75f4bb562312e261d7d0b452880b3c/bqplot/pyplot.py#L96-L97

Matplotlib has "." marker style. It is great for scatter plot when there are a lot of clustered data points.

mwcraig:
## References

Closes #1457 

## Code changes

See below 

## User-facing changes

This adds three more scatter marker shapes:

+ `plus`, which makes the shape `+`, with the shortcut `p`
+ `crosshair`, which makes the shape `×`, with the shortcut `x`
+ `point`, which makes a small circle, with the shortcut `.`

N/A

### DIFF
--- a/bqplot/marks.py
+++ b/bqplot/marks.py
@@ -578,7 +578,7 @@ class Scatter(_ScatterBase):
         Font-awesome icon for that mark
     name: string (class-level attribute)
         User-friendly name of the mark
-    marker: {'circle', 'cross', 'diamond', 'square', 'triangle-down', 'triangle-up', 'arrow', 'rectangle', 'ellipse', 'plus', 'crosshair'}
+    marker: {'circle', 'cross', 'diamond', 'square', 'triangle-down', 'triangle-up', 'arrow', 'rectangle', 'ellipse', 'plus', 'crosshair', 'point'}
         Marker shape
     colors: list of colors (default: ['steelblue'])
         List of colors of the markers. If the list is shorter than the number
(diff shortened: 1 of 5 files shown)


mwcraig:
@ibdafna @pllim

pllim (approved):
I like the new markers. Thanks!

Note to self: Open follow-up issue after this PR is merged to request "+" to be changed to thin "+" sign and "P" (uppercase P) to be the current "thick +" sign that is mapped to "+" symbol.

ibdafna:
Thanks for the PR, Matt! For what it's worth, the pyplot interface was not
designed with for compatibility with matplotlib. It's an issue we want to
resolve for 1.x. Chakri can probably discuss it in more details as it's his
implementation

ibdafna:
Will dig into the visual regression test during the week. Looks like node is failing building node-pre-gyp for some reason.

ibdafna (changes requested):
@mwcraig Thanks again for this PR! 🤗

Would it be possible to make the addition consistent by adding the new marker annotations to the following places?

- https://github.com/bqplot/bqplot/pull/1506/files#diff-fd3647496cb92f713ffed86551cf5eb1ab870896603af2be0f35e549bed8ebd4R409
- https://github.com/bqplot/bqplot/pull/1506/files#diff-fd3647496cb92f713ffed86551cf5eb1ab870896603af2be0f35e549bed8ebd4R330

Also, an updated example in the Lines.ipynb (where we list the different available markers) notebook would be awesome! 

image

And finally, once the notebook above is updated, we can add some visual regression testing via this PR. Thanks again! 💚

mwcraig:
@ibdafna -- I've addressed the review comments.

martinRenou (approved):
Thanks a lot Matt!

martinRenou:

PR #1513, opened 2022-08-13
exclude tests top-level package, add some JupyterLab trove classifiers

bollwyvl:
## References

- addresses https://github.com/conda-forge/bqplot-feedstock/pull/86
  - the feedstock will fail downstream once these are removed: this is fine

## Code changes

- removes the `tests` top-level module, which conflicts with other distributions
  - as it's "last-in-wins," this is not a reliable place to put them
  - some downstreams like `constructor` actively choke on them
- adds some JupyterLab trove classifiers

## User-facing changes

- prospective users would be able to see bqplot in the PyPI browse UI when looking for JupyterLab extensions

- n/a

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -20,6 +20,9 @@ install_requires =
     bqscales>=0.2.4,<0.3
 classifiers =
     Framework :: Jupyter
+    Framework :: Jupyter :: JupyterLab
+    Framework :: Jupyter :: JupyterLab :: 3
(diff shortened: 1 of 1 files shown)


bollwyvl:
The lint findings seem irrelevant... seems like a single-purpose PR running some formatters would be more helpful than anything else, with the appropriate git history flags to hide the churn.

martinRenou:
Thanks!

martinRenou:

PR #1521, opened 2022-09-02
Fix processPhosphorMessage for ipywidgets 7

martinRenou:
## References

Similar approach to what we do in ipyleaflet for supporting both ipywidgets 7 and 8


martinRenou:

PR #1523, opened 2022-09-02
Fix luminowidget

martinRenou:
The `luminoWidget` accessor does not work undo ipywidgets 7


martinRenou:

PR #1541, opened 2022-11-29
fix: axis formatting depended on initial values or changes

### LINKED ISSUE #1536: Issues with updating Axis.tick_labels
Updating tick labels using ``Axis.tick_labels`` doesn't always work properly - here are cases that currently don't work well:

* If I run all cells in the notebook 'Axis Properties.ipynb' in one go the labels do not update to be roman numerals, but things work fine if I run slowly cell by cell, which seems to suggest some kind of race condition if the labels are set before the figure is visible.
* Similarly, if I set the tick values and formats in the same cell, sometimes this doesn't work, which is probably the same as the above issue.
* If the tick values are updated, the tick labels revert to not having custom labels even if the new tick labels have entries in the tick_labels dictionary - for example if the tick_labels dictionary has entries for tick values of [1, 2, 3] and I change the tick_values from [1, 2, 3] to [1, 2] then the tick labels are not used. It's then not possible to re-assign tick_labels to the same dictionary as before since it won't trigger any updates so one needs to change it to an empty dictionary and change it back.

maartenbreddels:
The codepath to to axis formatting for the initial values was different from when properties such as tick_values are updated afterwards.
This was probably due to some technical debt/circular dependency in the code.
If did the minimal amount of change to make sure the behaviour is the same, while having a consistent codepath.

Fixes #1536

### DIFF
--- a/js/src/Axis.ts
+++ b/js/src/Axis.ts
@@ -59,12 +59,14 @@ export class Axis extends WidgetView {
 
     Promise.all([scale_promise, offset_promise]).then(() => {
       this.create_listeners();
-      this.tick_format = this.generate_tick_formatter();
+      this.create_axis();
       this.set_scales_range();
+      this.update_scales();
+      this.set_tick_values();
+      this.tickformat_changed();
       this.append_axis();
     });
   }
-
   create_listeners() {
     // Creates all event listeners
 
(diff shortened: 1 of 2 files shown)


maartenbreddels:
meeseeksdev backport to 0.12.x

PR #1550, opened 2023-01-27
Update CI

martinRenou:


martinRenou:
Let's go ahead with this, it's already an improvement in the CI, even though it does not pass yet

PR #1551, opened 2023-01-27
Try fixing ReadTheDocs

martinRenou:


martinRenou:
The hook seems to work now https://readthedocs.org/projects/bqplot/builds/19315259/ though it does not report back to Github unfortunately

PR #1563, opened 2023-02-01
Improve docs API further

martinRenou:


bqplot/axes.py
```diff
@@ -12,21 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-r"""
```
martinRenou, line 15: These were for Sphinx that we don't use anymore

PR #1566, opened 2023-02-01
Add link to the gallery

martinRenou:
@ChakriCherukuri Please tell me what you think

Screenshot from 2023-02-01 11-55-04


ChakriCherukuri:
@martinRenou  superb idea!
Instead of just a link can we add all the gallery images using the [grid](https://squidfunk.github.io/mkdocs-material/reference/grids/) layout of mkdocs?

martinRenou:
👍🏽 on making it more integrated to the docs using a grid.

I'm not quite sure though we could keep the same logic for spawning the binder.

PR #1539, opened 2022-11-29
fix: figure is not updating when margins change

mariobuikhuizen:
## Code changes

Extend bypass relayout logic to include margins

## User-facing changes

Changing margins results in an updated figure

To reproduce:
```python
import numpy as np
import bqplot.pyplot as plt

size = 100
scale = 100.0
np.random.seed(0)
x_data = np.arange(size)
y_data = np.cumsum(np.random.randn(size) * scale)

fig = plt.figure()
plt.plot(y_data)
fig
```

```python
fig.fig_margin = {'top': 0, 'bottom': 0, 'left': 30, 'right': 0}
```

PR #1572, opened 2023-02-02
Log console messages in UI-tests + fix visual regression tests

martinRenou:


martinRenou:
@mariobuikhuizen @maartenbreddels visual regression tests passing again! I think we were facing https://github.com/jupyter-widgets/ipywidgets/pull/3549 and updating the `yarn.lock` did the trick

maartenbreddels:
Great to hear!
What makes you think https://github.com/jupyter-widgets/ipywidgets/pull/3549 was the problem? Note that I found out today this was also present in widgets 7 https://github.com/jupyter-widgets/ipywidgets/pull/3687

martinRenou:
> What makes you think https://github.com/jupyter-widgets/ipywidgets/pull/3549 was the problem?

I'm not entirely sure this was the culprit. But we've seen very similar symptoms in ipyleaflet https://github.com/jupyter-widgets/ipyleaflet/pull/968#issuecomment-1219895024: issue in fetching some sub-widget state and not reproducible outside of Galata. So that's what makes me think it's the reason.

mariobuikhuizen:
Thanks for the notification @martinRenou! I've updated #1539, which now has no visual regression.

PR #1588, opened 2023-02-24
chore: update ci, backport of #1550

maartenbreddels:


maartenbreddels:
@martinRenou CI is working for 0.12.x, but not sure why this regression happened. Maybe you have seen this error before?

martinRenou:
It looks like ScatterGL does not work anymore in this branch. I've not seen this happen on master.

maartenbreddels:
For some reason we went from yarn 1 to 3, and that broke a lot, so we're not limiting yarn to <2

martinRenou:
Yeah I see the yarn=1 constraint pop up here and there

maartenbreddels:
@martinRenou can we get the js console output. Looking at the video, we don't get any WebGL output, so it's not a timing issue.

maartenbreddels:
@martinRenou found out how to get js console output, I now see
```
PAGE LOG: THREE.WebGLRenderer 91
PAGE LOG: THREE.WebGLShader: Shader couldn't compile.
PAGE LOG: THREE.WebGLShader: Shader couldn't compile.
PAGE LOG: THREE.WebGLProgram: shader error:  0 gl.VALIDATE_STATUS false gl.getProgramInfoLog   
PAGE LOG: WebGL: INVALID_OPERATION: useProgram: program not valid
PAGE LOG: WebGL: INVALID_OPERATION: drawElementsInstancedANGLE: no valid shader program in use
PAGE LOG: WebGL: CONTEXT_LOST_WEBGL: loseContext: context lost
```

So it seems there is a WebGL issue now, any ideas how to reproduce this locally? Or what can cause this? Maybe it's a playwright version issue.

maartenbreddels:
This was really tricky to get green 😫 

The problem was the 22.04 ubuntu image (ubuntu-latest). I thought I already changed that, but I only changed it for 1 step, not the `visual-regression-tests`. Setting `visual-regression-tests` to ubuntu-20.04 solved it.

Still not sure what the problem is, but maybe in a year from now this problem goes away.

PR #1598, opened 2023-04-11
fix: resize not always working

mariobuikhuizen:
The intersectObserver, used to determine if an element is visible, can contain multiple entries for the same target element. The last entry represents the last observed state, so we should use the last instead of the first for the visibility test.

See: https://github.com/spacetelescope/jdaviz/issues/2129


maartenbreddels:
Nice, thanks!

maartenbreddels:
meeseeksdev backport to master

PR #1601, opened 2023-06-27
fix: remove numpy upper bound pin

maartenbreddels:
We are unlikely to ever be affected by breaking changes in numpy because we do not use a large surface area.

See https://iscinumpy.dev/post/bound-version-constraints/#tldr
 'Anyone can fix a missing cap, but users cannot fix an over
  restrictive cap causing solver errors.'

Fixes https://github.com/spacetelescope/jdaviz/issues/2259


martinRenou:

PR #1603, opened 2023-07-04
fix: background color of figure was not computed properly

maartenbreddels:
On the screen, it is not the computed background color that is seen by a user. Since DOM elements can be transparent, the background color might be one of the parent elements. This is why we need to traverse the DOM tree to find the background color of the first non-transparent parent element.

A more sophisticated solution would be to also use blending when there is a non-fully transparent background color. I think that might be a bit overkill for now.

See also https://github.com/spacetelescope/jdaviz/pull/2264#issuecomment-1612052525


maartenbreddels:
Note that before we would get a different background color in voila and notebook because of style definitions on body. Now we get a white background color for both! :)
I also tested this with jupyter lab and dark themes, which works nicely.

martinRenou:

PR #1619, opened 2023-07-25
fix: set domain to (0, 1) when no data is present

maartenbreddels:
Fixes https://github.com/bqplot/bqplot/issues/1614

If no data is present, a scale has a domain of (0, 1) but if we remove the data afterwards (set it to an empty array) , we set it to (-inf, +inf) which causes rendering issues and make the bqplot figure unable to restore itself.
It is more consistent to behave like no data was passed in, and use (0, 1) for the domain.

### DIFF
--- a/js/src/LinearScaleModel.ts
+++ b/js/src/LinearScaleModel.ts
@@ -78,6 +78,17 @@ export class LinearScaleModel extends ScaleModel {
 
   update_domain() {
     const that = this;
+    // if all domains are empty, we reset to the default domain of (0, 1)
+    if (
+      _.every(this.domains, (d) => {
+        return d.length === 0;
+      })
+    ) {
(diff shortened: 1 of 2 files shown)


maartenbreddels:
I hope i remembered the magic phrase correctly: 

By the power vested in me by all code repositories, I call upon the almighty 'update galata references' phoenix. Arise from the ashes of the old screenshots and blaze a trail of new visuals across our Github sky!

martinRenou:
Looks like the phoenix burst into flames before it could try to fly.

I suspsect the galata bot will only work with the main branch from now on. I guess since it's now using JupyterLab 4 and the new galata.

martinRenou (approved):
Thanks!

martinRenou:
We need to upstream it to bqscales https://github.com/bqplot/bqscales/issues/84

PR #1620, opened 2023-07-27
Forward port fix scale domain no data #1619

martinRenou:
#1619


martinRenou:
update galata references

martinRenou:
Triggering CI

PR #1638, opened 2024-01-30
fix: domain should not be reset when min/max are set

maartenbreddels:
In #1619 we reset the domain to (0, 1) when no data is present. However, this only matters when the min/max are not set.

Issue found in https://github.com/spacetelescope/jdaviz/pull/2661


maartenbreddels:
pytest 8 seems to trigger the failure.

martinRenou (approved):

PR #1645, opened 2024-05-21
Backport PR bqplot#1644: fix: save_png doesn't work when character \xa0 is used (#1644)

maartenbreddels:
The character \xa0 is converted to &nbsp; when .outerHTML() is called, causing an error in the data-URL where the outer HTML is used.

(cherry picked from commit 4d5c2932d8d55e1af96e7129f0550c18c0c3b977)


martinRenou:
Maybe we karma-webpack would need to be pinned, or we can try to update nodejs...

PR #1654, opened 2024-10-11
Create dependabot.yml

pllim:
This setup covers both NPM and Actions. The grouping makes all the changes in one PR instead of spamming each update as a separate PR.


pllim:
@martinRenou , while you are at it, also please consider auto updating workflow versions using Dependabot. Thanks.

martinRenou (approved):
Thanks!

PR #1655, opened 2024-10-11
Use mamba-org/setup-micromamba for visual job

pllim:
See if this helps.


pllim:
Well, at least it ran to completion now, but has some real comparison failures that are beyond my skills to fix.

maartenbreddels:
The diff is visually very small, maybe it's a animation issue/fluke.

@martinRenou any idea what can cause this? A jupyter-lab css variable that changed maybe?

maartenbreddels:
update galata references

maartenbreddels:
update galata references

maartenbreddels:
Annoying, I think the update script looks at who opened the PR, not who send the message.

@martinRenou can you confirm that? And is there a reason for this?

martinRenou:
Indeed, that was the proposed fix when a CVE about this was published so I copied it over in all repos.

I think the solution in here is nicer (looking at the message author as you suggest): https://github.com/jupytercad/JupyterCAD/blob/main/.github/workflows/update_galata_references.yaml

martinRenou (approved):
Thanks!

PR #1670, opened 2025-02-11
fix: ascent is now in scipy.datasets instead of scipy.misc

maartenbreddels:
According to https://docs.scipy.org/doc/scipy-1.12.0/reference/generated/scipy.misc.ascent.html


maartenbreddels:

PR #1673, opened 2025-05-21
feat: add Figure.get_svg_data similar to get_png_data

maartenbreddels:
This is a follow-up of #1397


maartenbreddels:
Note that the scattergl update plots are broken, but we already saw that in https://github.com/bqplot/bqplot/pull/1588#issuecomment-1468683079

PR #1696, opened 2026-04-04
Filled in missing Documentation for Heatmap, Marks(Pie), Marks(market), scales, axes etc. Considered previous style and matched it correctly

ishaan-cherukuri:


main.py
martinRenou, line 1: I suppose this was committed by mistake?

martinRenou:
That's amazing thanks! I'll give this a review

docs/usage/advanced/linking-ipywidgets.md
```diff
@@ -1 +1,167 @@
...
+
+fig = plt.figure()
+x = np.linspace(0, 10, 100)
+scatter = plt.scatter(x, np.sin(x))
+
+# Link a slider to the mark's default_opacities without Python round-trip
+opacity_slider = widgets.FloatSlider(value=1.0, min=0.0, max=1.0, step=0.05,
+                                      description="Opacity")
+# jslink works on simple scalar traits
+widgets.jslink((opacity_slider, "value"), (fig, "fig_margin"))  # example only
```
martinRenou, line 69: I know it's only an example, but it is likely incorrect, `fig_margin` is not a scalar value but a dictionary like `dict(top=60, bottom=100, left=60, right=60)`

martinRenou (changes requested):
Thanks a ton, apart from one comment + the unwanted commited files that's good to go

